PR #20, opened 2024-12-25
Adding error message

Chaitanya84:
Small change to check all config and tools working properly.


ast.c
```diff
@@ -475,6 +475,7 @@ void execute_yapping_call(ArgumentList *args)
 {
     if (!args)
     {
+        yyerror("No argument ");
         yapping("\n");
```
araujo88, line 479: This line should be removed as well.

ast.c
```diff
@@ -508,6 +509,7 @@ void execute_yapping_call(ArgumentList *args)
     else
     {
         // No additional values, so just print the string
+        //yyerror("No argument ");
```
araujo88, line 512: This comment should be removed.

ast.c
```diff
@@ -516,6 +518,7 @@ void execute_yappin_call(ArgumentList *args)
 {
     if (!args)
     {
+        yyerror("No argument ");
         yappin("\n");
```
araujo88, line 522: Same here,

araujo88:
I think the broader issue is: do we want to keep both print to standard output functions flexible enough so that they support no arguments at all (and thus could simply do nothing instead of outputting a newline) or do we want to throw an error and enforce args for these funtions?

araujo88 (changes requested):

Issue #61, opened 2025-01-14
Handling of nested loops


araujo88:
The following code:

```
skibidi main {
 
    flex(rizz i =0; i <= 3; i++){
        yapping("%d",i);

        rizz j = 0;
        goon(j <=3){
            yapping("goon");
            edgy(j == 2){
                yapping("break goon %d",j);
            }
            j++;
        }

        edgy( i ==2){
            yapping("break flex");
            bruh;
        }
    }

    bussin 0;
}
```

Should output:

```
0
goon
goon
goon
break goon 2
goon
1
goon
goon
goon
break goon 2
goon
2
goon
goon
goon
break goon 2
goon
break flex
```

However, its current output is:

```
0
goon
goon
goon
break goon 2
goon
1
goon
goon
goon
break goon 2
goon
2
goon
goon
goon
break goon 2
goon
break flex
3
goon
goon
goon
break goon 2
goon
```

PR #62, opened 2025-01-14
Fix: nested loop break statement

SIGMazer:
## Description
Introducing new way to handle long jumps

## Related Issue

#61 

## Type of Change

- [x] Documentation update
- [ ] Performance improvement

## Checklist

### DIFF
--- a/ast.c
+++ b/ast.c
@@ -2,14 +2,13 @@
 
 #include "ast.h"
 #include <stdbool.h>
-#include <setjmp.h>
 #include <math.h>
 #include <limits.h>
 #include <float.h>
 #include <stdint.h>
 #include <stdio.h>
 
-static jmp_buf break_env;
+static JumpBuffer *jump_buffer = {0};
 
 TypeModifiers current_modifiers = {false, false, false, false, false};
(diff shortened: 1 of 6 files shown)


examples/mixed_types.brainrot
```diff
@@ -1,5 +1,6 @@
 skibidi main {
     chad f = 3.14;
+    bruh;
```
araujo88, line 3: shouldn't this throw an error, since `break` cannot be used outside a loop in C?

SIGMazer: yeah, you right i'll fix it

SIGMazer: it actually  through an error but some how the test didn't get it

araujo88: Yes, it is because of the const test that should throw an error. I'll think of a better way of handling tests that should output stderr.

araujo88: For example, the `baka` test now is disabled because it outputs to stderr.

araujo88 (approved):

araujo88 (approved):

PR #122, opened 2025-02-08
Region-based allocation for AST nodes 

SIGMazer:
## Description

As the AST grows more complex, managing it becomes increasingly difficult, and free_ast becomes larger. This PR introduces a single Region to allocate all AST nodes, allowing them to be freed at once, simplifying memory management and reducing overhead.

## Type of Change

- [ ] Documentation update
- [x] Performance improvement
- [x] Refactor

## Checklist

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -11,7 +11,7 @@ LDFLAGS := -lfl -lm
 # Source files and directories
 SRC_DIR := lib
 DEBUG_FLAGS := -g
-SRCS := $(SRC_DIR)/hm.c $(SRC_DIR)/mem.c $(SRC_DIR)/input.c ast.c
+SRCS := $(SRC_DIR)/hm.c $(SRC_DIR)/mem.c $(SRC_DIR)/input.c $(SRC_DIR)/arena.c  ast.c
 GENERATED_SRCS := lang.tab.c lex.yy.c
 ALL_SRCS := $(SRCS) $(GENERATED_SRCS)
 
--- a/ast.c
+++ b/ast.c
@@ -3435,6 +3435,7 @@ void free_ast(ASTNode *node)
     case NODE_FUNCTION_DEF:
         SAFE_FREE(node->data.function_def.name);
         // Free parameters
+        free_parameters(node->data.function_def.parameters);
         if (node->data.function_def.body)
         {
             free_ast(node->data.function_def.body);
@@ -3786,6 +3787,7 @@ void enter_function_scope(Function *func, ArgumentList *args)
         Variable *var = variable_new(curr_param->name);
         var->var_type = curr_param->type;
         add_variable_to_scope(curr_param->name, var);
+        SAFE_FREE(var);
 
         switch (curr_param->type)
         {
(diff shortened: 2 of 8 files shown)


SIGMazer:
@araujo88 can you take a look

araujo88:
@SIGMazer can you explain in more detail how this region-based allocation works? Or provide a reference?

SIGMazer:
> @SIGMazer can you explain in more detail how this region-based allocation works? Or provide a reference?

We pre-allocate a memory chunk (Arena) and use it to allocate all `ASTNode` instances. This approach allows us to free the entire Arena at once, instead of traversing each `ASTNode` individually and freeing them one by one.

### Reference
- https://en.wikipedia.org/wiki/Region-based_memory_management
- https://www.gingerbill.org/article/2019/02/08/memory-allocation-strategies-002/

araujo88:
> > @SIGMazer can you explain in more detail how this region-based allocation works? Or provide a reference?
> 
> We pre-allocate a memory chunk (Arena) and use it to allocate all `ASTNode` instances. This approach allows us to free the entire Arena at once, instead of traversing each `ASTNode` individually and freeing them one by one.
> ### Reference
> 
>     * https://en.wikipedia.org/wiki/Region-based_memory_management
> 
>     * https://www.gingerbill.org/article/2019/02/08/memory-allocation-strategies-002/

Very interesting! I wasn't familiar with that approach. Should we be concerned with the disadvantages? I understand that, as for the current usages of brainrot right now (which are small programs) this shouldn't be a concern.

> Region-based memory management works best when the number of regions is relatively small and each contains many objects; programs that contain many sparse regions will exhibit [internal fragmentation](https://en.wikipedia.org/wiki/Internal_fragmentation), leading to wasted memory and a time overhead for region management. Again, in the presence of region inference this problem can be more difficult to diagnose.

SIGMazer:
> > > @SIGMazer can you explain in more detail how this region-based allocation works? Or provide a reference?
> > 
> > 
> > We pre-allocate a memory chunk (Arena) and use it to allocate all `ASTNode` instances. This approach allows us to free the entire Arena at once, instead of traversing each `ASTNode` individually and freeing them one by one.
> > ### Reference
> > ```
> > * https://en.wikipedia.org/wiki/Region-based_memory_management
> > 
> > * https://www.gingerbill.org/article/2019/02/08/memory-allocation-strategies-002/
> > ```
> 
> Very interesting! I wasn't familiar with that approach. Should we be concerned with the disadvantages? I understand that, as for the current usages of brainrot right now (which are small programs) this shouldn't be a concern.
> 
> > Region-based memory management works best when the number of regions is relatively small and each contains many objects; programs that contain many sparse regions will exhibit [internal fragmentation](https://en.wikipedia.org/wiki/Internal_fragmentation), leading to wasted memory and a time overhead for region management. Again, in the presence of region inference this problem can be more difficult to diagnose.

I don't think this affects our case because we allocate a single contiguous region for all AST allocations, meaning we won't encounter any sparse regions. In the future, if we need to delete or transform nodes before freeing the entire Arena, we can introduce optimizations such as merging underutilized regions. Alternatively, we could use a bump allocator, which treats the region as a stack to avoid fragmentation. However, in our current case, I believe our implementation will not face any issues.

araujo88 (approved):

araujo88 (approved):

PR #123, opened 2025-02-12
Cooked giga and thicc up

### LINKED ISSUE #99: Add support for other type modifiers
Currently, type modifiers such as `giga` and `thicc` are not implemented.

dominik-air:
## Related Issue

Resolves #99 

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement
- [ ] Refactor

## Checklist

- [ ] My code follows the style guidelines of this project
- [ ] I have performed a self-review of my own code
- [ ] I have documented my changes in the code or documentation
- [ ] I have run the unit tests locally
- [ ] All new and existing tests pass


araujo88:
You implemented `long` and `long long` as integer types, when they are actually a type modifier. You can have a `long double` or `long long double` for example. `long float` and `long long float` are not valid types.

dominik-air:
lil-yachty-drake

My bad. Gonna rework this no cap

araujo88:
Take your time to cook, no rush. And thanks for being willing to grind like a true sigma. Reach out on Discord if you need some guidance on this issue.

SIGMazer (commented):
.

dominik-air:
> i think the errors because you edit the code of array access

Yeah, I didn't understand how Value works. Now it's doing what I was expecting

araujo88:
@dominik-air do you need some help with this?

SIGMazer:
@dominik-air You can reduce the scope of the PR to make it just cover one of the modifiers. i think it will be simpler

araujo88:
> @dominik-air You can reduce the scope of the PR to make it just cover one of the modifiers. i think it will be simpler

Indeed, it is better to keep PRs as granular as possible.

dominik-air:
@SIGMazer Agreed. This will be just a `giga` MR, but I think adding `thicc` later won't be as challenging.

@araujo88 Help is always appreciated. I kinda tried to refactor some parts to reduce code duplication and this took me a while to implement without new bugs coming up. I still need to work on the "promotion" and node operations for those new types.

Dockerfile
```diff
@@ -0,0 +1,16 @@
+FROM ubuntu:latest
```
dominik-air, line 1: I compile and run brainrot inside a Docker container because I wasn't able to make it work on my Mac. Maybe this could be a future addition to the codebase to ease the setup process. Wdyt @araujo88 @SIGMazer 

Of course a separate MR would be better for this contribution. I'll remove this in the final MR

ast.c
```diff
@@ -5,8 +5,6 @@
 #include <stdbool.h>
 #include <math.h>
 #include <limits.h>
-#include <float.h>
-#include <stdint.h>
```
dominik-air, line 9: Unused includes

ast.c
```diff
@@ -1081,6 +1118,52 @@ void *handle_unary_expression(ASTNode *node, void *operand_value, int operand_ty
     }
 }
 
+Value retrieve_array_access_value(ASTNode *node, Value default_return_value)
```
dominik-air, line 1121: I wanted to reduce code duplication, so I extracted the array access code into this function

Dockerfile
SIGMazer: yup also i was thinking of add docker container to test the interpreter in other OS

ast.c
SIGMazer: about think i was planing to refactor it in some sort of look up table so it will be like this 

```c
    get_value[var->type](array_data, idx);
```

i think this will make it easy for us when implementing pointers and multi-N arrays

Dockerfile
SIGMazer: btw what the problem in mac os? maybe we can solve it

ast.c
dominik-air: yup, good idea, but for now I just wanted to clean up a little bit

Dockerfile
dominik-air: tbh I don't even remember what was the issue, since I first came across brainrot some weeks ago

I just went with a Docker container because it was less fuss

SIGMazer: Fair! If you try it and it has an issue please open an issue with the error

alxlemesh:
Some checks were not successful😂

SIGMazer:
@dominik-air need any help here?

Nebula-Studios:
Where blud go? 😭

leo-aa88:
@dominik-air u alive bro

PR #182, opened 2026-08-17
fix: stop `based` (default) from firing over a matching case in `ohio` (#179)

### LINKED ISSUE #179: ohio/based: default case fires unconditionally regardless of match order
## What happens

`based` (default) in an `ohio` (switch) statement fires unconditionally the moment the interpreter's scan reaches it, regardless of whether a later `sigma rule` (case) would have matched. In other words, `based`'s position in source order determines whether it wins — real switch statements shouldn't care about that at all.

## Repro

```c
skibidi main {
    rizz choice = 2;
    ohio (choice) {
        based:
            yapping("DEFAULT FIRED");
            bruh;
        sigma rule 1:
            yapping("case 1");
            bruh;
        sigma rule 2:
            yapping("case 2");
            bruh;
    }
    bussin 0;
}
```

Prints `DEFAULT FIRED` — even though `choice == 2` and `sigma rule 2` is right there. The existing `test_cases/switch_case.brainrot` doesn't catch this because it (like every switch example in this repo) places `based` last, after every numbered case — the one ordering where the bug is invisible.

## Root cause

`execute_switch_statement` in `ast.c` (around line 613) does a single sequential scan over cases:

```c
while (current_case)
{
    if (current_case->value)
    {
        int case_value = evaluate_expression(current_case->value);
        if (case_value == switch_value || matched)
        {
            matched = 1;
            execute_statements(current_case->statements);
        }
    }
    else
    {
        // Default case
        if (matched || !matched)
        {
            execute_statements(current_case->statements);
            break;
        }
    }
    current_case = current_case->next;
}
```

`if (matched || !matched)` is `true` for every possible value of `matched` — it's a tautology, so the default branch always executes and immediately `break`s the very first time the scan reaches a `based` case. It doesn't matter whether `matched` was already `1` (i.e. it's meant to be a legitimate fallthrough into default after a prior case) or still `0` (i.e. nothing has matched yet and this should only fire if *nothing later matches either* — which a single forward-only scan can't know before it's finished).

Found via `-Wtautological-negation-compare` while turning on `-Werror` for the WASM build in #178 — clang flags this, gcc's `-Wall -Wextra -Werror` (what the native build actually uses) does not, which is why it's shipped silently until now.

## Why I'm not fixing it here

Fixing this changes real interpreter behavior for `ohio`/`based`, not build tooling — it needs a real decision about intended semantics (does `based` need to move to a genuine two-pass scan: check every case for a match first, then only run default if none matched? does case order matter at all, matching real C, or is source order intentionally significant here?) plus a `test_cases/*.brainrot` fixture covering "default declared before a later matching case" once fixed. That's out of scope for #178, which is adding a WASM build target — it works around this with a named, commented `-Wno-tautological-negation-compare` on the wasm build only, pointing at this issue.

## Suggested next step

Restructure `execute_switch_statement` to find the matching case (or confirm none matches) before running anything, so `based`'s position in the case list stops affecting control flow. Add a `test_cases/switch_case_default_first.brainrot`-style fixture (default before a matching case) to `tests/expected_results.json` once fixed, so this can't silently regress again.

ethanhawkes-gif:
Fixes #179.

### Problem
`execute_switch_statement` (ast.c) walked the case list once, forward, and ran the `based` (default) case the instant the scan reached it:

```c
// Default case
if (matched || !matched)   // always true
{
    execute_statements(current_case->statements);
    break;
}
```

Because `matched || !matched` is a tautology, `based` fired unconditionally the moment it was reached — so a `based` declared **before** a numbered `sigma rule` that actually matches would win, purely because of its position in source order. Every switch example in the repo puts `based` last, the one ordering where the bug is invisible, which is why it shipped silently (and why only clang's `-Wtautological-negation-compare`, under the wasm `-Werror`, ever flagged it).

Repro (from the issue) — prints `DEFAULT FIRED` even though `choice == 2`:

```c
skibidi main {
    rizz choice = 2;
    ohio (choice) {
        based:
            yapping("DEFAULT FIRED");
            bruh;
        sigma rule 1: yapping("case 1"); bruh;
        sigma rule 2: yapping("case 2"); bruh;
    }
    bussin 0;
}
```

### Fix
Two-phase dispatch:
1. **Find the entry case** — the first `sigma rule` whose value equals the switch value, or `based` only if no numbered case matches. This is what stops `based`'s position from deciding whether it runs.
2. **Execute from the entry onward**, falling through into later cases (numbered or default) exactly like a C `switch`, until a `bruh` (break) `longjmp`s out.

`based` now runs iff no numbered case matches, regardless of where it is declared. Fallthrough and break semantics are unchanged; the `no match + no default` path now correctly does nothing instead of relying on the default always running.

### Tests
- Adds `test_cases/switch_case_default_first.brainrot` + its `tests/expected_results.json` entry: `based` declared before a matching case (must pick the case), plus a second switch where nothing matches (must still pick `based`) — so the fix can't over-correct and can't silently regress.
- Full suite: **76 passed** with the repo's own `make` build (which carries `-fsanitize=address,undefined -Werror`) — no ASan/UBSan diagnostics.
- Also hand-checked fallthrough (case without `bruh`), a mid-list `default` with a match after it, and `default`-without-`bruh` falling through to the next case.

### Note (out of scope here)
The wasm target's `-Wno-tautological-negation-compare` suppression (added in #178, commented as pointing at this exact `if (matched || !matched)`) is now dead — the tautology is gone. I left the Makefile untouched since I can't compile-test the emcc/wasm target in this environment; dropping that flag is a safe follow-up when the wasm build is next touched.

Signed-off-by: Ethan Hawkes <ethanhawkes-gif@users.noreply.github.com>


leo-aa88:
Thanks for working on this issue. I'll take a look on your code.

ast.c
```diff
@@ -615,33 +615,49 @@ bool check_and_mark_identifier(ASTNode *node, const String contextErrorMessage)
...
+        entry = default_case;
+
+    if (!entry)
+        return; // no matching case and no default: nothing to run
+
+    PUSH_JUMP_BUFFER();
+    if (setjmp(CURRENT_JUMP_BUFFER()) == 0)
+    {
+        int running = 0;
+        for (CaseNode *current_case = node->data.switch_stmt.cases; current_case; current_case = current_case->next)
```
leo-aa88: Phase 2 re-walks from the head with a `running` flag, but `entry` is already the start node — you can drop the flag and the pre-`entry` rescan:
```c
for (CaseNode *c = entry; c; c = c->next)
    execute_statements(c->statements);
```

ast.c
```diff
@@ -615,33 +615,49 @@ bool check_and_mark_identifier(ASTNode *node, const String contextErrorMessage)
 void execute_switch_statement(ASTNode *node)
 {
     int switch_value = evaluate_expression(node->data.switch_stmt.expression);
```
leo-aa88, line 2158: `evaluate_expression` returns `int`, so `long`/`long long`/`float` switch or case values are truncated before comparison (e.g. `4294967297` would match `sigma rule 1`). Pre-existing, but worth noting since the function was rewritten here.

test_cases/switch_case_default_first.brainrot
```diff
@@ -0,0 +1,29 @@
+skibidi main {
+    🚽 A matching numbered case must win even when `based` (default) is
+    🚽 declared before it in source order. Regression for issue #179.
+    rizz a = 2;
+    ohio (a) {
+        based:
+            yapping("default A");
+            bruh;
```
leo-aa88, line 8: Both blocks put `bruh` right after the default, so the behavior that changed most — C-style fall-through *out of* `based` (a default with no trailing `bruh`, then a later `sigma rule`) — is never exercised. Worth adding a fall-through case to lock it in.

leo-aa88 (commented):
Automated review — the core fix is correct (traced both test cases and the default-first-with/without-match, empty-switch, and first-match-wins paths). A few notes; inline comments below.

One finding can't be attached inline because the file isn't part of this diff:

**README.md:273** — the *Current limitations* bullet still describes #179 as current behavior ("placing it before a case that would otherwise match currently pre-empts that match"). This PR fixes exactly that, so the bullet is now stale and should be removed/updated in the same PR.

leo-aa88 (changes requested):

leo-aa88:

ast.c
```diff
@@ -2156,40 +2156,45 @@ bool check_and_mark_identifier(ASTNode *node, const String contextErrorMessage)
...
+        else
+        {
+            default_case = scan;
+        }
+    }
+
+    if (!entry)
+        entry = default_case;
+
+    if (!entry)
```
leo-aa88, line 2189: This ordering is the crux and it is correct: `entry` is chosen from a numbered match first, falling back to `default_case` only when no numbered case matched, so `based`'s position in the list is irrelevant. Two things I verified are load-bearing here: (1) the early `return` sits before `PUSH_JUMP_BUFFER`, so the no-match/no-default path never pushes and the push/pop stay balanced; (2) phase 2 iterates from `entry` through every following case regardless of numbered-vs-default, which is what makes a match fall through a mid-list `based` into later cases exactly like C. Rebuilding with the old switch body makes the checked-in fixture print `default A/default B/default C`, so the test genuinely catches the regression.

leo-aa88 (approved):
LGTM.

I reviewed the semantics against C's `switch` and verified the whole thing in an isolated worktree rather than trusting the description.

**The root cause and fix are both right.** The old single forward scan ran the default the instant it reached it (`if (matched || !matched)` is a tautology), so a `based` declared before a matching numbered case pre-empted the match. The two-phase rewrite picks the entry case first (first numbered case equal to the switch value, else the recorded default) and only then executes from that entry, falling through into later cases until a `bruh` breaks. `based`'s position no longer decides anything.

**Control flow is clean.** Phase 1 does no statement execution, so it needs no jump buffer; the early `return` for no-match-and-no-default happens before `PUSH_JUMP_BUFFER`, so push/pop stay balanced on every path. The break (`bruh`) longjmps into the else arm, then `POP_JUMP_BUFFER` runs; normal fall-through to the end pops too. No imbalance, no leak.

**The fixture is a genuine regression test.** I rebuilt the interpreter with the old buggy switch body and the checked-in `switch_case_default_first.brainrot` prints `default A / default B / default C` instead of the expected `A two / default B / default C / C one`. So it pins both the position bug (switch A) and the fall-through-out-of-`based` behavior (switch C, where the old code broke immediately and never reached `C one`). Expected-output entry matches byte for byte.

**Adversarial cases all match C**, ASan/UBSan clean:
- default in the middle with a match after it -> takes the match (the exact #179 shape).
- no match and no default -> runs nothing, no crash.
- multi-case fall-through with no `bruh` -> runs each in turn.
- a match that falls through *through* a mid-list default -> executes the default's statements too, like C.
- duplicate case values -> first wins.

Full `pytest` is green (502 passed) with no regression, and the switch region is clang-format-clean. The README limitation bullet is correctly dropped.

**One non-blocking follow-up, already called out in the description:** the wasm target's `-Wno-tautological-negation-compare` (added in #178 specifically for this `if (matched || !matched)`) is now dead code suppression. Harmless as-is; worth removing next time the wasm build is touched. Not a reason to hold this.

Ship it.

Issue #186, opened 2026-08-18
make valgrind likely never runs real leak detection (ASan + valgrind incompatibility)


leo-aa88:
## Summary

`make valgrind` runs `run_valgrind_tests.sh` against `$(TARGET)` (the `brainrot` binary), which is always built with `-fsanitize=address,undefined` per the main `brainrot` target's `CFLAGS`. Running an ASan-instrumented binary under valgrind is a known-broken combination unless specific accommodations are made (ASan runtime preloaded first, or a valgrind build with ASan-compat support). Without that, the binary crashes on startup before any program logic executes:

```
==...==ASan runtime does not come first in initial library list; you should either
       link runtime to your application or manually preload it with LD_PRELOAD.
==...== Process terminating with default action of signal 11 (SIGSEGV)
```

This reproduces on unmodified `main` for every single fixture in `test_cases/` — `run_valgrind_tests.sh`'s per-file valgrind invocation crashes immediately, valgrind reports "0 allocs, 0 frees, 0 bytes allocated" (nothing ran), and the script's exit-code check (`-eq 100`) never trips because valgrind's own exit code from a SIGSEGV'd child isn't 100. So `make valgrind` reports success while never actually exercising the interpreter under valgrind.

## Why this matters

If CI runners hit the same incompatibility (plausible — this isn't sandbox-specific, it's inherent to running ASan binaries under valgrind), `make valgrind` has likely been passing without ever catching a real leak. This is how the leaks in #185 went unnoticed: building a plain (non-ASan) binary from the current generated grammar files and running the *same* fixtures under valgrind immediately turns up pre-existing, real leaks in `cooked_syntax_error.brainrot` and `cooked_bad_include.brainrot`.

## Suggested direction

Add a separate build target/binary specifically for the valgrind job that omits `-fsanitize=address,undefined` (keep the sanitizer build for `make test` / normal development, where ASan is genuinely useful), and point `run_valgrind_tests.sh` at that binary instead. Something like:

```makefile
brainrot-valgrind: $(ALL_SRCS)
	$(CC) -Wall -Wextra -Wpedantic -Werror -O2 -g -o $@ $(ALL_SRCS) $(LDFLAGS)

valgrind: ensure-stdrot brainrot-valgrind
	@./run_valgrind_tests.sh brainrot-valgrind
```

(with `run_valgrind_tests.sh` taking the binary path as a parameter, or a hardcoded second binary name).

[...]

Issue #185, opened 2026-08-18
Mid-parse fatal errors (YYABORT/exit) leak lexer and parser allocations


leo-aa88:
## Summary

Any fatal error that terminates the process from *inside* `yyparse()` — via `YYABORT` in a grammar action, or via `exit()` called from a helper invoked during a grammar action — leaks memory that would normally be freed once the enclosing rule finishes reducing (or once `root` is fully built and `cleanup()`'s `free_ast()` can walk it). This is a structural gap: bison's generated parser doesn't unwind semantic-value-stack allocations on `YYABORT`, and lexer-allocated token buffers that haven't yet been claimed by a reduction are simply abandoned.

## Reproduction

Built without sanitizers (a `-fsanitize=address` binary can't run under valgrind at all in most environments — see #2) and run under `valgrind --leak-check=full --error-exitcode=100`:

**Existing `YYABORT` path** ([lang.y:378](../blob/main/lang.y#L378), "Initializer count does not match array dimensions"):
```
rizz arr[][3] = {1,2,3,4,5};
```
```
168 (144 direct, 24 indirect) bytes in 1 blocks are definitely lost
    by 0x... safe_malloc (mem.c:174)
    by 0x... variable_new (ast.c:3829)
    by 0x... yyparse (lang.y:363)

200 (40 direct, 160 indirect) bytes in 1 blocks are definitely lost
    by 0x... safe_malloc (mem.c:174)
    by 0x... create_expression_list (ast.c:3531)
    by 0x... yyparse (lang.y:485)
```

**Genuine bison syntax errors** — reproducible today on unmodified `main` via existing fixtures `test_cases/cooked_syntax_error.brainrot` and `test_cases/cooked_bad_include.brainrot`:
```
24 bytes in 1 blocks are possibly lost
    by 0x... safe_malloc (mem.c:174)
    by 0x... copy_bytes (lang.l:287)
    by 0x... yylex (lang.l:478)
    by 0x... yyparse (lang.tab.c:1800)
```

## Why this matters

`run_valgrind_tests.sh` sweeps every file in `test_cases/`, not just ones referenced in `expected_results.json`, and uses `--error-exitcode=100` without `--errors-for-leak-kinds`, so "possibly lost" blocks already count as failures. In other words, `cooked_syntax_error.brainrot` and `cooked_bad_include.brainrot` — both already in the repo — would fail `make valgrind` today if it were actually able to run (see #2 for why it currently can't).

## Suggested direction

[...]

PR #188, opened 2026-08-18
feat: implement chungus (union) keyword

leo-aa88:
## Description

Implements `chungus` (union), the last ❌ in the README keyword table. Mirrors the existing `gang`/struct implementation as closely as possible:

- New `struct_or_union` grammar nonterminal shared by `struct_def` and the struct/union variable-declaration productions, so `gang`/`chungus` reuse the same rules instead of duplicating them.
- `StructDef` gains `is_union`; struct and union tags share one registry, matching C's shared tag namespace (you can't have both a `gang Foo` and a `chungus Foo`).
- New `compute_union_layout`: every field gets offset 0, blob size is the largest member — this is what gives union fields their overlapping storage.
- Union brace-initializers must supply exactly one value (`chungus Data d = {42};`), matching C's rule that only the first member is initialized from a plain brace list. More than one value (`{1, 2.0}`) is a semantic error. This check runs in the semantic-analysis phase (after parsing completes) rather than mid-parse, specifically to avoid leaking parser-internal allocations — see the note on issue #185 below.
- Struct/union member-access error messages now say "Struct" or "Union" as appropriate instead of always saying "Struct".
- Fixed an off-by-one bug in `count_expression_list` found while implementing the initializer-arity check: it returned N+1 for a single-element circular expression list (a `do…while` that always ran its body once). This also silently affected unsized single-element array declarations (`rizz arr[] = {5};` was allocating a 2-element array).
- Docs: README table flipped to ✅, new §7.10 "Unions (`chungus`)" in `docs/the-brainrot-programming-language.md`.
- Tests: `test_cases/chungus.brainrot` (overlapping storage + single-member init), `test_cases/chungus_init_fail.brainrot` (arity error), `test_cases/semantic_error_chungus_unknown_member.brainrot` (unknown-member error), and `test_cases/unsized_array_one.brainrot` (regression for the `count_expression_list` fix on the array-length path), all added to `tests/expected_results.json`.

While building this I also found (and filed separately, not fixed here as they're pre-existing and out of scope) #185 and #186 — a class of leak on mid-parse fatal errors, and `make valgrind`'s ASan-instrumented binary being unable to run under valgrind in most environments. I designed the initializer-arity check to avoid #185 rather than add a new instance of it.

## Related Issue

N/A — implements a table entry from the README/roadmap, not filed as its own tracked issue.

## Type of Change

- [x] Bug fix (non-breaking change which fixes an issue) — the `count_expression_list` off-by-one

## Checklist

- [x] I have run the valgrind memory tests locally (see note: `make valgrind` as configured can't run meaningfully in this environment because the ASan-instrumented binary crashes immediately under valgrind — filed as #186. I built a non-sanitized binary from the same generated grammar files and ran the full `test_cases/*.brainrot` suite through `valgrind --leak-check=full --error-exitcode=100` by hand: zero new leaks from this change; the only two failures found — `cooked_bad_include.brainrot` and `cooked_syntax_error.brainrot` — are pre-existing and unrelated, filed as #185.)
- [x] All new and existing tests pass (87/87 via `make test`)


leo-aa88:
Addressed the line-number fix (9c19214) — good catch, and the one-liner you suggested was exactly right:

- `create_struct_access_node` now sets `node->line_number = yylineno`, same as `create_node` does for every other node type. `semantic_error_chungus_unknown_member` now correctly reports `at line 8` instead of falling back to `at line 1`. Re-ran the full suite (87/87) and a valgrind pass on the affected fixtures — clean.
- Updated the PR description: fixed the stale "85/85" → "87/87", and listed the two fixtures added in the previous round that the description hadn't caught up to.
- Filed #189 for the tag-namespace last-wins gap, per your "diagnose the clash, or soften the docs" suggestion — leaving it as a tracked follow-up rather than expanding this PR, since (as you noted) it's pre-existing behavior for `gang` too, not something this PR narrows.

Thanks for pushing on these — appreciate the review.

PR #191, opened 2026-08-18
feat: support arrays and structs declared inside functions

leo-aa88:
## Description

Arrays and struct/union locals didn't work inside non-`main` functions:
declaring one and touching it inside a function body raised `Variable 'x'
is not defined` / `Undefined struct or union variable` the moment the
function actually ran.

**Root cause:** array/struct declarations allocated their storage as a
side effect of *parsing* (`set_multi_array_variable` / `add_variable_to_scope`
called directly from `lang.y` grammar actions), always against the single
scope object that exists throughout parsing. Function scopes are only
pushed later, at call time — so a "local" array/struct actually landed in
the wrong (global) scope, and `get_variable()` deliberately refuses to look
past a function-scope boundary, making the declaration invisible as soon as
the function ran.

**Fix:** move that allocation to runtime (`interpreter_visit_declaration`),
in whatever scope is current when the declaration statement actually
executes — the same way plain scalar locals already worked. This gives
function-local arrays/structs real per-call storage (recursion works, two
functions can each have a same-named local without colliding), and required
moving the struct-access chain type resolution the semantic analyzer relied
on (previously piggybacking on that same parse-time Variable) onto the
analyzer's own symbol table instead.

Since they fell out of the same runtime-storage model with a manageable
amount of extra work, this PR also adds:
- **struct-by-value function parameters** (deep-copied, not aliased)
- **struct-by-value return values** (caller copies out of a temp blob)

Array-by-value parameters/returns are intentionally left unsupported,
matching C (arrays already decay to pointers via the existing pointer-param
path).

**Found but out of scope** (pre-existing on `main`, unrelated to this
change, not fixed here): combining a struct member access with another
operand in a binary expression, or assigning one directly to a plain
scalar, crashes the interpreter — reproduces even at top level with
`gang Point p; rizz s; s = p.x;` inside `skibidi main`. Worth a follow-up
issue.

## Related Issue

N/A — no existing issue; found via manual investigation.

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement
- [ ] Refactor

## Checklist

**Note on valgrind:** `make valgrind` doesn't run in this environment —
this build is `-fsanitize=address,undefined`, and Valgrind and ASan can't
coexist (confirmed the same failure on a clean checkout of `main`, so it's
an environment issue, not something this PR introduces). In its place I
leaned on the ASan/UBSan-instrumented binary directly: `make test` (110
cases) passes clean, plus targeted stress runs (3000-iteration loops
exercising recursive/repeated function-local array and struct allocation,
struct-by-value params, and struct-by-value returns) with zero
ASan/UBSan/LeakSanitizer errors.

## Test plan

- [x] `make test` — 110/110 passing (98 original + 12 new across three review rounds)
- [x] Manual repro of the original bug (array/struct local in a function) — now works
- [x] Recursion with a local array/struct — correct per-call isolation
- [x] Two functions with same-named locals — no collision
- [x] Struct param mutated in-callee doesn't affect caller (value semantics)
- [x] Struct returned by value, and struct-to-struct copy-init
- [x] 3000-iteration stress loops under ASan/UBSan — no leaks or errors
- [x] `make format-check` — clean except two pre-existing, unrelated violations already present on `main`

### DIFF
--- a/README.md
+++ b/README.md
@@ -278,7 +278,12 @@ Current limitations include:
 
 - Limited support for complex expressions
 - Basic error reporting
-- No support for arrays in user-defined functions
+- Struct/union function parameters and return values must be a plain
+  struct/union variable of the exact declared type (no sub-expressions,
+  e.g. a chained call or a member access) — arguments and return values are
+  deep-copied, not aliased
+- Arrays cannot be passed or returned by value (only via a pointer
+  parameter, which aliases the caller's array like in C)
 - `ohio`/`based` (switch/default): `based` fires as soon as the interpreter's scan reaches it, so its position among the `sigma rule` cases matters — placing it before a case that would otherwise match currently pre-empts that match ([#179](https://github.com/Brainrotlang/brainrot/issues/179))
 
 ## 🗺️ Roadmap
--- a/ast.c
+++ b/ast.c
@@ -2746,7 +2746,10 @@ void *handle_function_call(ASTNode *node)
                 safe_strdup(&current_return_value.value.strvalue);
             break;
         case VAR_STRUCT:
-            /* struct return not yet supported; fall through to NULL */
+            /* A struct return has no meaningful representation in this
+               generic scalar-expression context; discard it (freeing the
+               blob handle_return_statement allocated) rather than leak. */
+            free_pending_struct_return_value();
             break;
         case NONE:
             return NULL;
@@ -4105,10 +4108,24 @@ void execute_function_call(const String name, ArgumentList *args)
         return;
     }
 
-    enter_function_scope(func, args);
+    /* A previous call's struct return value (see handle_return_statement's
+       VAR_STRUCT case) may still be sitting unconsumed in the global
+       current_return_value slot -- e.g. it was returned from a call used
+       as a bare statement, which has nowhere to put it. Free it now,
+       before this call overwrites the slot. */
+    free_pending_struct_return_value();
+
     current_return_value.type = func->return_type;
     current_return_value.pointer_level = func->return_pointer_level;
     current_return_value.has_value = false;
+
+    if (!enter_function_scope(func, args))
+    {
+        /* Argument binding failed (already reported via yyerror) -- no
+           scope was left behind, and there's no valid parameter state to
+           run the body against, so don't. */
+        return;
+    }
 
     PUSH_JUMP_BUFFER();
     if (setjmp(CURRENT_JUMP_BUFFER()) == 0)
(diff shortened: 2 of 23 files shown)


leo-aa88:
Thanks for the thorough review — all four blocking issues confirmed and fixed in 10213b5, plus the docs and the non-blocking suggestions.

**1. Leaked struct return blob.** `current_return_value`'s struct blob/tag now has a single owner: `free_pending_struct_return_value()`, called before `execute_function_call` overwrites the slot, after a struct-returning call used as a bare statement, in `handle_function_call`'s "not supported in this context" fallback, and by the consumer once it's copied the value out. Repro from the review (`make_point(1,2); make_point(3,4);`) is leak-clean under ASan/LSan now.

**2. memcpy sized from the wrong struct.** All three copy paths (copy-init `gang T x = y;`, init-from-call `gang T x = f();`, and by-value parameter binding) now compare `struct_name` before copying and reject a mismatch as an error instead of trusting the destination's size against a possibly-smaller source. Your `Small`/`Big` repros (copy-init, return, and param) all produce a clean error now instead of a heap-buffer-overflow.

**3. `find_symbol` not function-scoped.** `SymbolEntry` now carries the name of the function it was declared in (or empty for global/`main`-level), tracked via `analyzer->current_function_name` through both `collect_declarations` and `semantic_analyze_with_scope_tracking`. `find_symbol` only returns a match that's global or belongs to the function currently being analyzed. Your two-functions-same-local-name-different-type repro now runs correctly instead of misresolving `bar`'s `p` while analyzing `foo`. Added `same_name_struct_locals_no_collision` as a regression fixture.

**4. Failed struct-arg bind still ran the callee.** `enter_function_scope` now returns `bool`, restores the reversed parameter list on every early-return path, and `execute_function_call` doesn't touch the body if binding failed. `take(make())` now stops cleanly at "Struct argument must be a plain struct variable" — no more running in the caller's scope and then tearing down the wrong one.

**Docs (#5).** Updated the three stale "structs/arrays can't be function params or returns" claims in README/language-doc/user-guide, plus the ROADMAP L8 line, to describe the actual (plain-variable, deep-copy) semantics.

**Non-blocking:**
- `create_function_def_node_struct` now threads `pointer_level` through and rejects `gang T *foo()` at parse time with a clear error, rather than silently treating it as by-value and failing confusingly at the return statement.
- Checked `create_function_ex`'s allocation — it's `SAFE_MALLOC`, which does `memset` to zero (see `lib/mem.c:safe_malloc`), so `return_struct_name` was already zero-initialized; no change needed there.
- Added NULL guards on the destination blob before the two struct memcpys this PR introduced in `interpreter_visit_declaration`.
- New fixture (`initialized_locals_in_function`) covering function-local array/struct/union declarations *with* a braced initializer, which wasn't covered before.

All 108 tests pass (98 original + 10 new across both rounds). Re-ran your exact repros plus my earlier stress tests under ASan/LSan — all clean. `make valgrind` still doesn't run in this environment for the same pre-existing ASan/Valgrind incompatibility you also saw (confirmed again on this round, unrelated to the diff).

One thing I did *not* chase: the pre-existing `p.x + p.y` / `s = p.x;` crash from the original PR description is still there (confirmed unrelated to this branch, reproduces on `main`). Left out of scope per your note.

PR #198, opened 2026-08-18
Fix maxxing (sizeof) on struct/union-typed operands

### LINKED ISSUE #195: maxxing (sizeof) on a struct/union expression fails instead of returning its size
## Summary

`maxxing(expr)` (the `sizeof` operator) doesn't handle struct/union-typed
expressions at all, despite structs/unions being a supported type. It
fails with a misleading error and prints `0` instead of the type's real
size, rather than either working correctly or being rejected with an
accurate diagnostic.

## Reproduction

```c
gang Point {
    rizz x;
    rizz y;
};
gang Line {
    gang Point start;
};

skibidi main {
    gang Point p;
    yapping("%d", maxxing(p));          // struct-typed identifier

    gang Line l;
    yapping("%d", maxxing(l.start.x));  // scalar field: works
    yapping("%d", maxxing(l.start));    // struct-typed field
}
```

Output:
```
Error: Undefined variable in sizeof at line 10
0
4
Error: Invalid type in sizeof at line 13
0
```

`p` and `l.start` are both very much defined — the error messages are
wrong.

## Root cause

`get_type_size_for_descriptor()` ([ast.c:50](../blob/main/ast.c#L50)) has
no `VAR_STRUCT` case and falls through to `default: return 0;`.

- `get_type_size()` ([ast.c:2311](../blob/main/ast.c#L2311)), used for a
  bare identifier operand, treats `base == 0` as "the variable doesn't
  exist" and reports `"Undefined variable in sizeof"` even though the
  variable is real — it just legitimately has no primitive size.
- `handle_sizeof()`'s non-identifier branch ([ast.c:2329](../blob/main/ast.c#L2329))
  switches on `get_expression_type(expr)` and has no `VAR_STRUCT` case
  either, falling to `"Invalid type in sizeof"`.

Both should instead look up the operand's `StructDef` (via
`get_struct_def()`/`var->struct_name` for an identifier, or
`resolve_struct_access()` for a struct-typed field access) and return
`StructDef->total_size`, mirroring how `compute_struct_layout()` /
`compute_union_layout()` already size a *nested* struct/union field
inside another struct (see `get_struct_field_size()` in ast.c) — this bug
is really the same "VAR_STRUCT isn't a case in the primitive-size table"
gap, just hit from a different call path that never got the same fix.

Found while implementing nested struct/union support (#190) — not
specific to nesting, reproduces for any struct/union variable on
unmodified `main`.

ChrisJr404:
Fixes #195. `maxxing` on a struct/union value was erroring out and printing `0` instead of the type's size, so this teaches both sizeof paths to fall back to the type's layout size.

Two spots, exactly as the issue laid out. `get_type_size()` (the bare-identifier path) reads a `0` primitive size as "this variable doesn't exist" and yells `Undefined variable in sizeof`, even though a struct variable is real and just has no primitive size. And `handle_sizeof()`'s non-identifier branch has no `VAR_STRUCT` case, so a struct-typed field access like `l.start` drops to `Invalid type in sizeof`. Both now look up the operand's `StructDef` — `get_struct_def(var->struct_name)` for an identifier, `resolve_struct_access()` for a field access — and return `def->total_size`, the same size `get_struct_field_size()` already uses for a nested field. Unions share that registry, so they come along for free. Struct pointers still hit the descriptor path and report pointer size, and sizeof on primitives is untouched.

Repro from the issue now prints `8 / 4 / 8` instead of two errors and two zeros.

Added `test_cases/sizeof_struct.brainrot` covering a struct identifier, a scalar field, a struct-typed field, a nested scalar field, and a union identifier, with its entry in `expected_results.json` (`8\n4\n8\n4\n8\n`). Sizes were read off the running interpreter, not guessed. `make test` is green — 111 passed, including the new case and the existing `sizeof` primitives test. Binary builds with ASan/UBSan on (the project's default `CFLAGS`) and the new fixture runs clean under it.

One note on `make format-check`: `ast.c` passes, but the target flags two pre-existing lines in `stdrot.c` and `lib/mem.h` that I didn't touch — they fail the same way on a clean `main` with clang-format 21, so it looks like a clang-format version gap rather than anything in this change.

### DIFF
--- a/ast.c
+++ b/ast.c
@@ -2313,6 +2313,18 @@ size_t get_type_size(String name)
     Variable *var = get_variable(name);
     if (var != NULL)
     {
+        /* A struct/union value has no primitive size; look up its
+           definition and use the computed layout size instead. Pointers
+           to structs fall through to the descriptor path (pointer size). */
+        if (var->var_type == VAR_STRUCT && var->pointer_level == 0)
+        {
+            StructDef *def = get_struct_def(var->struct_name);
+            if (def != NULL)
+                return var->is_array ? def->total_size * var->array_length
+                                     : def->total_size;
+            yyerror("Undefined variable in sizeof");
+            return 0;
+        }
         size_t base = get_type_size_for_descriptor(
             var->var_type, var->pointer_level, var->modifiers);
         if (base == 0)
@@ -2359,6 +2371,29 @@ size_t handle_sizeof(ASTNode *node)
         case VAR_CHAR:
             return get_type_size_for_descriptor(
                 type, get_expression_pointer_level(expr), expr->modifiers);
+        case VAR_STRUCT:
+        {
+            int plevel = get_expression_pointer_level(expr);
+            if (plevel > 0)
+                return get_type_size_for_descriptor(type, plevel,
+                                                    expr->modifiers);
+            /* A struct/union-typed field access (e.g. `l.start`) — resolve
+               it to the field's definition and return its layout size. */
+            if (expr->type == NODE_STRUCT_ACCESS)
+            {
+                StructDef *def = NULL;
+                void *base = NULL;
+                StructField *fld = NULL;
+                if (resolve_struct_access(expr, &def, &base, &fld, false))
+                {
+                    StructDef *sdef = get_struct_def(fld->struct_name);
+                    if (sdef != NULL)
+                        return sdef->total_size;
+                }
+            }
+            yyerror("Invalid type in sizeof");
+            return 0;
+        }
         default:
             yyerror("Invalid type in sizeof");
             return 0;
--- a/ast.c
+++ b/ast.c
@@ -2322,7 +2322,7 @@ size_t get_type_size(String name)
             if (def != NULL)
                 return var->is_array ? def->total_size * var->array_length
                                      : def->total_size;
-            yyerror("Undefined variable in sizeof");
+            yyerror("Unknown struct or union type");
             return 0;
         }
         size_t base = get_type_size_for_descriptor(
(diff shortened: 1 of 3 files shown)


leo-aa88:
Thanks for your contribution. I'll provide a review.

ast.c
```diff
@@ -2359,6 +2371,29 @@ size_t handle_sizeof(ASTNode *node)
         case VAR_CHAR:
             return get_type_size_for_descriptor(
                 type, get_expression_pointer_level(expr), expr->modifiers);
+        case VAR_STRUCT:
```
leo-aa88, line 2374: `maxxing(*q)` on a dereferenced struct pointer still lands here and returns 0 with `Invalid type in sizeof` (verified on this branch): `get_expression_type` gives `VAR_STRUCT`, `plevel` is 0, but the node is `NODE_UNARY_OPERATION`, not `NODE_STRUCT_ACCESS`. Same operand class the PR targets — resolving the deref operand's variable and returning its `StructDef->total_size` would close it.

ast.c
```diff
@@ -2313,6 +2313,18 @@ size_t get_type_size(String name)
...
+        /* A struct/union value has no primitive size; look up its
+           definition and use the computed layout size instead. Pointers
+           to structs fall through to the descriptor path (pointer size). */
+        if (var->var_type == VAR_STRUCT && var->pointer_level == 0)
+        {
+            StructDef *def = get_struct_def(var->struct_name);
+            if (def != NULL)
+                return var->is_array ? def->total_size * var->array_length
+                                     : def->total_size;
+            yyerror("Undefined variable in sizeof");
```
leo-aa88: `gang Ghost g; maxxing(g);` reports `Undefined variable in sizeof` — the variable is declared, it's the tag that isn't registered. `resolve_struct_access` uses `Unknown struct or union type` for this exact condition; matching that wording points users at the real mistake.

leo-aa88 (commented):
Built the branch and probed the sizeof paths directly: `make test` green (111), fixture LeakSanitizer-clean under the default ASan/UBSan build, `clang-format` 14 clean on `ast.c`, and `get_type_size` has exactly one caller so the early return can't perturb anything else. Two notes inline, neither a regression.

leo-aa88 (changes requested):

ChrisJr404:
Good catches, both fixed in c43563a.

- \`maxxing(*q)\`: the VAR_STRUCT branch now handles the \`NODE_UNARY_OPERATION\` / \`OP_DEREFERENCE\` operand — resolves the deref operand's variable and returns its \`StructDef->total_size\` instead of falling through to \`Invalid type in sizeof\`. \`maxxing(*q)\` on a \`Point *\` now yields 8.
- \`gang Ghost g; maxxing(g);\`: the unregistered-tag path now reports \`Unknown struct or union type\`, matching \`resolve_struct_access\`, so it points at the missing tag rather than the (present) variable.

Added a dereferenced-pointer line to the \`sizeof_struct\` fixture too. Re-ran \`make test\` — still green (111), and \`ast.c\` is clang-format clean.

leo-aa88 (approved):

PR #216, opened 2026-08-18
Implement gyatt (enum) keyword

leo-aa88:
## Description

Implements the `gyatt` (enum) keyword, matching feature parity with the
existing `gang`/`chungus` (struct/union) support:

- Enum type definitions at the top level, with explicit constant values and
  C-style auto-increment (`OK = 0, WARN = 5, ERR` → `ERR` is `6`)
- Unscoped, C-style constants — `RED` is usable directly as an `int`-valued
  expression anywhere after its enum is defined, matching real C semantics
  (a scoped `Color.RED` form was considered and intentionally not used)
- Variable declarations, both local (inside a function) and global
- Nesting: an enum type can be used as a `gang`/`chungus` field
- Function parameters and return types
- `maxxing` (sizeof), `ohio`/`sigma rule` (switch/case), arithmetic and
  comparisons — an enum value behaves like a plain `int` everywhere C
  would allow it

Enum tags live in their own namespace, separate from the struct/union tag
registry (matching C, where `struct`/`union`/`enum` each get a distinct
namespace). Enum *constants*, however, share a single global namespace
across every enum in the program (also matching C) — two different `gyatt`
types can't declare a constant with the same name.

Unlike a struct/union, an enum variable has no runtime blob — it's a plain
`int` under the hood, so it's routed through the existing scalar
declaration/assignment paths rather than needing new storage machinery.

## Related Issue

N/A

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement
- [ ] Refactor

## Checklist

## Test plan

- [x] `make test` — 124/124 passing, including 13 new `enum_*`/`sizeof_enum`
      fixtures covering definitions, explicit values/auto-increment, local
      and nested (struct/union) declarations, function params/return,
      `maxxing`, `ohio`/`sigma rule`, and 4 error cases (duplicate
      constant, redefinition, unknown type, cross-enum constant collision)
- [x] `make format-check` — clean except a pre-existing, unrelated
      violation in `lib/mem.h` present on `main` before this branch
- [ ] `make valgrind` — not meaningfully runnable in the environment this
      was developed in: valgrind 3.18.1 crashes on ASan-instrumented
      binaries before `main()` even starts, reproducibly on every test
      case (old and new alike) on unmodified `main` too. The `-fsanitize=
      address,undefined` build itself (used by `make test`) reported no
      leaks across the full suite, including a real leak this PR found
      and fixed along the way (`enum_name` wasn't freed in `lib/hm.c`'s
      `hm_free`, mirroring the existing `struct_name` free). Worth a
      clean-environment valgrind run before merge if CI can do it.

### DIFF
--- a/ast.c
+++ b/ast.c
@@ -1376,9 +1376,8 @@ void *handle_binary_operation(ASTNode *node)
     int left_type = get_expression_type(node->data.op.left);
     int right_type = get_expression_type(node->data.op.right);
 
-    // An enum operand is an int in C -- normalize here so every check below
-    // (promotion, and the VAR_INT branches in the float/double paths) treats
-    // it as one, instead of falling through to the VAR_SHORT default.
+    // An enum operand is an int in C -- normalize here so every VAR_INT
+    // check below treats it as one too.
     if (left_type == VAR_ENUM)
         left_type = VAR_INT;
     if (right_type == VAR_ENUM)
@@ -2871,6 +2870,12 @@ void *handle_function_call(ASTNode *node)
             free_pending_struct_return_value();
             break;
         case VAR_ENUM:
+            if (current_return_value.pointer_level > 0)
+            {
+                return_value = SAFE_MALLOC(uintptr_t);
+                *(uintptr_t *)return_value = current_return_value.value.pvalue;
+                break;
+            }
             return_value = SAFE_MALLOC(int);
             *(int *)return_value = current_return_value.value.ivalue;
             break;
@@ -3195,11 +3200,13 @@ bool is_expression(ASTNode *node, VarType type)
     {
         VarType ret =
             get_function_return_type(node->data.func_call.function_name);
-        /* An enum return has type int in C -- treat it as such here so
+        /* A by-value enum return has type int in C -- treat it as such so
            general int-context callers (e.g. ast_expr_to_stdrot_value's
-           varargs marshalling) recognize it without needing a VAR_ENUM
-           case of their own. */
-        if (type == VAR_INT && ret == VAR_ENUM)
+           varargs marshalling) recognize it without a VAR_ENUM case of
+           their own. A pointer-to-enum return is not an int, though. */
+        if (type == VAR_INT && ret == VAR_ENUM &&
+            get_function_return_pointer_level(
+                node->data.func_call.function_name) == 0)
             return true;
         return ret == type;
     }
--- a/ast.c
+++ b/ast.c
@@ -1376,6 +1376,14 @@ void *handle_binary_operation(ASTNode *node)
     int left_type = get_expression_type(node->data.op.left);
     int right_type = get_expression_type(node->data.op.right);
 
+    // An enum operand is an int in C -- normalize here so every check below
+    // (promotion, and the VAR_INT branches in the float/double paths) treats
+    // it as one, instead of falling through to the VAR_SHORT default.
+    if (left_type == VAR_ENUM)
+        left_type = VAR_INT;
+    if (right_type == VAR_ENUM)
+        right_type = VAR_INT;
+
     // Promote types if necessary (short -> int -> float -> double).
     int promoted_type = VAR_SHORT;
     if (left_type == VAR_DOUBLE || right_type == VAR_DOUBLE)
@@ -4549,10 +4557,7 @@ ASTNode *create_function_def_node_enum(String name, String enum_name,
                                        ASTNode *body)
 {
     /* An enum return is a plain int at runtime, so unlike the struct
-       variant above, no pointer_level restriction is needed here -- the
-       generic VAR_INT-shaped path (handle_return_statement, handle_
-       function_call, interpreter_visit_declaration) already handles any
-       pointer_level correctly. */
+       variant above, no pointer_level restriction is needed here. */
     ASTNode *node = create_function_def_node_ex(name, VAR_ENUM, pointer_level,
                                                 params, body);
     if (node)
@@ -4848,6 +4853,7 @@ void free_struct_registry(void)
             StructField *nxt = f->next;
             SAFE_FREE(f->name);
             SAFE_FREE(f->struct_name);
+            SAFE_FREE(f->enum_name);
             SAFE_FREE(f);
             f = nxt;
         }
--- a/ast.c
+++ b/ast.c
@@ -19,6 +19,8 @@ HashMap *function_map = NULL;
 static HashMap *static_variable_map = NULL;
 static HashMap *struct_registry = NULL;
 static StructDef *struct_registry_list = NULL;
+static HashMap *enum_registry = NULL;
+static EnumDef *enum_registry_list = NULL;
 bool struct_def_had_error = false;
 ReturnValue current_return_value;
 Arena arena;
@@ -77,6 +79,8 @@ size_t get_type_size_for_descriptor(VarType type, int pointer_level,
         return sizeof(int);
     case VAR_STRING:
         return sizeof(String);
+    case VAR_ENUM:
+        return sizeof(int);
     case NONE:
     default:
         return 0;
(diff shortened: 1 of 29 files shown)


ast.c
```diff
@@ -1252,6 +1287,11 @@ VarType get_expression_type(ASTNode *node)
         {
             return var->var_type;
         }
+        /* Not a variable -- an enum constant has type int in C. */
+        if (find_global_enum_constant(array_name) != NULL)
+        {
+            return VAR_INT;
```
leo-aa88: **[bug]** Enum constants are typed as `VAR_INT` here, but an enum *variable* still returns `VAR_ENUM` from `get_expression_type` (and so do `NODE_STRUCT_ACCESS` fields and `NODE_FUNC_CALL` returns). `handle_binary_operation` only promotes to `VAR_INT` when a side is literally `VAR_INT` (`ast.c:1385`), so two `VAR_ENUM` operands stay `VAR_SHORT`: operands are read via `evaluate_expression_short` (`*(short *)&ivalue`) and the result is a `short`. Confirmed: `gyatt Status { A = 40000, B = 50000 };` then `x + y` prints `24464` while `A + B` (constants) correctly prints `90000`. Small values in the new tests (`GREEN + BLUE`, `c == RED`) hide this because they fit in 16 bits and `safe_malloc` over-aligns/zero-fills the `short` result. Same hole applies to `s.c + t.c` and `pick(0) + pick(1)` when both sides are enum-typed. The PR claims int-like arithmetic and the semantic analyzer now *allows* `VAR_ENUM` arithmetic (`semantic_analyzer.c:548`), so this is a real correctness miss, not an unclaimed extra.

**Suggestion:** Treat `VAR_ENUM` as `VAR_INT` in `handle_binary_operation`'s promotion (and in the `left_type == VAR_INT` branches of the float/double eval paths). Alternatively, make `get_expression_type` / `is_expression` report `VAR_INT` for enum variables too, matching the constant/function-return special cases already added. Add a fixture with values outside `[-32768, 32767]` covering var+var, field+field, and call+call.

interpreter.c
```diff
@@ -464,6 +470,8 @@ void interpreter_visit_declaration(Visitor *self, ASTNode *node)
     var->modifiers = node->modifiers;
     var->var_type = node->var_type;
     var->pointer_level = node->pointer_level;
+    if (node->var_type == VAR_ENUM)
+        var->enum_name = safe_strdup(&node->enum_name);
```
leo-aa88: **[bug]** `var->enum_name = safe_strdup(&node->enum_name)` runs *before* the `salty` (static) early-return. On every subsequent execution of `salty gyatt Color c;`, a new heap string is allocated, the existing static is found, and `SAFE_FREE(var)` frees only the wrapper `Variable`. Confirmed with ASan (`detect_leaks=1`): three calls to a function containing `salty gyatt Color c;` leak 48 bytes in 2 objects from this exact line; `salty rizz c;` does not leak. `get_variable` does consult `static_variable_map`, so this path is live.

**Suggestion:** Move the `safe_strdup` to after the static-already-exists return (so the first visit still copies `enum_name` into the object `hm_put` snapshots, and later visits allocate nothing). Alternatively `SAFE_FREE(var->enum_name)` before `SAFE_FREE(var)` on that early return. Add a `salty gyatt` fixture that calls the function more than once.

lang.y
```diff
@@ -270,6 +275,89 @@ struct_field
...
+            {
+                char msg[MAX_BUFFER_LEN];
+                snprintf(msg, sizeof(msg), "Unknown enum type '%s'",
+                        $2.data);
+                yyerror(msg);
+                struct_def_had_error = true;
+            }
+            $$ = create_parameter_ex($3.name, VAR_ENUM, $3.pointer_level,
+                                     NULL, (TypeModifiers){0});
+            $$->enum_name = ARENA_STRDUP($2);
```
leo-aa88: **[suggestion]** Nested `gyatt` fields set `Parameter.enum_name`, but the `struct_def` copy loop (`lang.y:186-190`) still only transfers `name` / `type` / `struct_name` / `pointer_level`. `StructField.enum_name` is therefore always the zeroed `SAFE_MALLOC` default, despite `ast.h:93` documenting it as "set whenever type == VAR_ENUM". Layout and the current tests work because they only need `type == VAR_ENUM`. Anything that later type-checks a field by tag (or `SAFE_FREE`s `f->enum_name` next to `f->struct_name`) will see an empty/wrong tag.

**Suggestion:** `f->enum_name = safe_strdup(&p->enum_name);` in the Parameter→StructField walk, and free it in `free_struct_registry` next to `struct_name`.

lang.y
```diff
@@ -281,6 +369,12 @@ function_def
             SAFE_FREE($2);
             SAFE_FREE($3.name);
         }
+    | ENUM IDENTIFIER declarator LPAREN params RPAREN LBRACE statements RBRACE
```
leo-aa88: **[suggestion]** Variable declarations and nested fields reject an unknown enum tag (`Unknown enum type 'Ghost'`). Function return types (`gyatt Ghost pick()`) and parameters (`rizz foo(gyatt Ghost g)`) do not — they parse and run as a phantom int-like type. Confirmed: `rizz foo(gyatt Ghost g) { bussin g; }` with `foo(1)` prints `1` and exits 0.

**Suggestion:** Use the same `get_enum_def` + `struct_def_had_error` check already used for declarations/fields (or a semantic-analyzer equivalent) on the `ENUM IDENTIFIER` function_def and param_list productions.

lang.y
```diff
@@ -270,6 +275,89 @@ struct_field
...
+        {
+            EnumConstant *c = SAFE_MALLOC(EnumConstant);
+            c->name = safe_strdup(&$1);
+            c->value = 0;
+            c->has_explicit_value = false;
+            c->next = NULL;
+            $$ = c;
+            SAFE_FREE($1.data);
+        }
+    | IDENTIFIER EQUALS INT_LITERAL
```
leo-aa88: **[suggestion]** Explicit enumerator values only accept `INT_LITERAL`, so `RED = -1` is `syntax error, unexpected MINUS, expecting INT_LITERAL`. C enums commonly use negatives; the docs say "same rules as C" for auto-increment/explicit values.

**Suggestion:** Accept a unary-minus integer (or a tiny constant-expression) on the right-hand side of `IDENTIFIER EQUALS`, and add a fixture.

ast.h
```diff
@@ -89,11 +90,37 @@ typedef struct StructField
                             self-referential `gang Node *next;` field still
                             needs its tag recorded even though chaining `.`
                             through it isn't supported yet */
+    String enum_name;   /* nested enum tag; set whenever type == VAR_ENUM */
     int pointer_level;
     size_t offset; /* byte offset within the struct blob */
     struct StructField *next;
 } StructField;
 
+/* A single named constant inside an enum body (e.g. `RED` in
```
leo-aa88: **[suggestion]** Several new comments restate the change, retell C-vs-scoped design history, or point at other functions by name (`EnumConstant` / `EnumDef` / `create_function_def_node_enum` / the `handle_identifier` fallback / `enter_function_scope`'s "not routed through set_int_variable"). Repo style is a short WHY, not WHAT or architecture narrative.

**Suggestion:** Collapse to one-liners where the identifier or a nearby check is not obvious (e.g. why enum tags are a separate registry; why the static `enum_name` copy cannot go through `set_int_variable`). Drop the rest.

docs/the-brainrot-programming-language.md
```diff
@@ -587,7 +588,124 @@ through a pointer-typed field isn't yet supported.
...
+gyatt Status {
+    OK = 0,
+    WARN = 5,
+    ERR        🚽 auto-increments to 6
+};
+```
+
+#### Enum Declaration
+
+Declare an enum-typed variable inside a function body (or at the top level):
```
leo-aa88: **[nit]** "Declare an enum-typed variable inside a function body (or at the top level)" is not true of the grammar — file-scope declarations are not a production (`function_def_list` is only functions / `gang` / `chungus` / `gyatt` *types*).

**Suggestion:** Drop "(or at the top level)" or reword to match the `gang`/`chungus` limitation already listed under "Current Limitations".

PR #223, opened 2026-08-19
Phase 2: a typed native ABI (signatures + semantic-checked native calls)

### LINKED ISSUE #205: [Roadmap] Phase 2 — A typed native ABI
> **Roadmap tracking:** #214
> **Depends on:** #204
> **Blocks / unblocks:** #206, #207, #208 (Road B)

---

## Summary

The semantic analyzer cannot check a single builtin argument today. Making native exports self-describing fixes that and unlocks generated bindings (Phase 5 Road B).

| | |
| --- | --- |
| **Roadmap** | Phase 2 — A typed native ABI |
| **Milestone** | M2 — Types flow |
| **Priority** | P0 |
| **Status** | not started |
| **Depends on** | Phase 1 |
| **Blocks** | Phase 3, Phase 4, Phase 5 Road B (typed / generated bindings). Phase 9 and 10 get better signatures from this but are not blocked. |
| **Baseline** | `0cfa958` |

### Verified limitations this phase closes

| # | Limitation | Evidence |
| - | ---------- | -------- |
| L3 | The semantic analyzer types every builtin as `NONE` — it knows a name exists and nothing more. | `semantic_analyzer.c:352` |
| L4 | The registry carries no signature: `StdrotEntry` is `{ name, fn }`. | `stdrot/stdrot_api.h:86` |
| L5 | `StdrotValue` has no pointer, struct, or handle representation — only `int/float/double/short/bool/char/String/void`. | `stdrot/stdrot_api.h:49` |

Until this lands, `WindowShouldClose() -> bool` is indistinguishable from a typo, and a generator has nothing to emit descriptors from.

## Goal

Type-check native calls exactly the way Brainrot-defined functions are type-checked — replacing the `return NONE` at `semantic_analyzer.c:352`.

The analyzer should know, for example:

- `WindowShouldClose() -> bool`
- `GetFrameTime() -> float`
- `LoadTexture(const char *) -> Texture2D`

### Extend `StdrotType`

```c
typedef enum {
    STDROT_VOID,
    STDROT_I8,  STDROT_U8,
    STDROT_I16, STDROT_U16,
    STDROT_I32, STDROT_U32,
    STDROT_I64, STDROT_U64,
    STDROT_F32, STDROT_F64,
    STDROT_BOOL,
    STDROT_STRING,    /* Brainrot String  */
    STDROT_CSTRING,   /* NUL-terminated   */
    STDROT_PTR,
    STDROT_STRUCT,
    STDROT_HANDLE,
} StdrotType;
```

Not every width is needed on day one. **`STDROT_PTR` and honest return values are mandatory**; the rest can land incrementally.

### Make the registry descriptive

```c
typedef struct {
    StdrotType  type;
    const char *type_name;      /* for STDROT_STRUCT / STDROT_HANDLE */
    int         pointer_level;
} StdrotParam;

typedef struct {
    const char        *name;
    StdrotParam        return_type;
    const StdrotParam *params;
    int                param_count;
    StdrotFn           fn;
} StdrotEntry;
```

### The string boundary

Brainrot uses `String { char *data; size_t len; }`; C libraries want `const char *`. Define the conversion **once**, in the ABI layer, rather than scattering `malloc(len+1)/memcpy/'\0'` through every wrapper.

**Ownership rule to decide up front:** adapter-owned scratch buffer, freed after the call, unless the signature is annotated as escaping.

### Concrete tasks

- [ ] Add `STDROT_EXPORT_SIG(...)` macro alongside the legacy `STDROT_EXPORT`.
- [ ] Extend `StdrotType` / `StdrotValue` with at least `STDROT_PTR`, `STDROT_CSTRING`, and a handle representation. Remaining widths can follow.
- [ ] Change `StdrotEntry` to carry `return_type` + `params`.
- [ ] Migrate **all existing builtins** (`yapping`, `yappin`, `baka`, `bet`, `chill`, `ragequit`, `slorp`, …) to signatures.
- [ ] Teach the semantic analyzer to type-check native calls (arity + types), emitting errors instead of `NONE`.
- [ ] Centralize Brainrot `String` ↔ C `const char *` conversion in the ABI layer; document the escaping annotation.
- [ ] Tests: arity mismatch, type mismatch, correct typed return used in an expression.

## Example (what the analyzer should accept / reject)

```c
skibidi main {
    cap ok = bet(W);           /* ok: bet(cap) -> cap */
    bet(2);                    /* error: type mismatch */
    bet(W, L);                 /* error: arity */
    bussin 0;
}
```

## Files likely to change

- `stdrot/stdrot_api.h`, `stdrot.c`
- every `stdrot/*.c` that uses `STDROT_EXPORT`
- `semantic_analyzer.c` / `semantic_analyzer.h`
- `ast.c` marshalling of `StdrotValue` → `current_return_value`
- `test_cases/` native-call type-error fixtures + `tests/expected_results.json`

## Definition of done

- [ ] `STDROT_EXPORT_SIG(...)` exists alongside legacy `STDROT_EXPORT`
- [ ] Existing builtins migrated to signatures
- [ ] Semantic errors for arity and type mismatches on native calls, with tests
- [ ] String conversion lives in one place in the ABI layer

## Guiding principles that apply

- **Don't port, bind.** This phase is the ABI, not a library.
- **Generate, don't hand-write.** Signatures are what a generator (Phase 5) will emit.
- **No libffi.** Generated C means the C compiler does the ABI work. We are not writing a dynamic call machine.

## References

- [docs/ROADMAP.md — Phase 2](https://github.com/Brainrotlang/brainrot/blob/main/docs/ROADMAP.md#phase-2--a-typed-native-abi)
- Appendix B Q6 — ownership of native resources (handles vs GC)

leo-aa88:
## Description

Implements Phase 2 of the FFI roadmap: native exports (`stdrot/*.c`) are now
self-describing, and the semantic analyzer + runtime marshalling both honor
that description end-to-end — not just a type check that the runtime is free
to ignore.

### The ABI

- `StdrotEntry` carries a full signature: `return_type`, `params[param_count]`
  (each entry checked whenever the caller actually supplies it), `min_args`
  (how many of those are mandatory vs. optional-but-typed — `bet`'s trailing
  message string is the one case that needs this), `is_variadic` (allows more
  args beyond `param_count`, left unchecked — format-string tails), and
  `return_like_arg` (-1 normally; ≥ 0 marks an identity-polymorphic return,
  e.g. `slorp<T>(T) -> T`, set explicitly via a dedicated
  `STDROT_EXPORT_SIG_IDENTITY()` macro rather than inferred from shape).
- `StdrotType` gains `STDROT_ANY` (unchecked), `STDROT_PTR`, `STDROT_CSTRING`,
  `STDROT_HANDLE`. The legacy `STDROT_EXPORT(name, fn)` still works as an
  explicitly-unchecked fallback for not-yet-migrated/third-party exports.
- All seven existing builtin exports (`bet`, `ragequit`, `chill`, `yapping`,
  `yappin`, `baka`, `slorp`) migrated to typed signatures. `bet` now honestly
  returns a real `cap` instead of a placeholder int, and its optional message
  argument is type-checked as a string whenever supplied.
- **The descriptor drives marshalling, not just the semantic analyzer.** The
  semantic analyzer accepts int/short/float/double interchangeably for a
  numeric-typed parameter (matching the language's existing numeric-coercion
  rules) and a Brainrot string for a `STDROT_CSTRING` one — so
  `execute_native_call()` now actually converts each checked argument to
  match its declared type before the native function runs, instead of
  handing it whatever union member the call site happened to produce. A
  native wrapper that trusts its own declared signature now gets what it
  asked for. This is enforced on the way *out* too: `execute_native_call()`
  now checks the `StdrotValue` a native actually returns against its
  declared `return_type` (coercing numerically where valid, `exit(1)`ing
  with a clear diagnostic if a native's C implementation returns something
  genuinely incompatible with what it declared) — the return descriptor was
  previously advisory only, trusted by the analyzer but never checked at
  the runtime boundary — including the legacy/untyped `STDROT_ANY` path:
  a native declared `STDROT_ANY` that actually returns a `STDROT_PTR`/
  `STDROT_HANDLE` is now rejected at the runtime boundary too, since the
  static side has no way to catch it (the descriptor says `ANY`, not
  `PTR`, so every pointer-level guard downstream stays silent). `STDROT_PTR`
  is fully supported end to end (marshalled as arguments and return
  values, see `tests/stdrot/testptr.c`); `STDROT_HANDLE` and
  `STDROT_CSTRING`-as-a-*return*-type are honestly *not* supported yet —
  a signature that declares either is rejected with a clear error rather
  than silently treated as unchecked, since nothing in the pipeline
  marshals either end to end (`STDROT_CSTRING` as an *argument* type is
  fully supported — only the return direction has no marshalling code).
  An opaque pointer is now also a real, distinct static type (`VAR_PTR` in
  the analyzer's `VarType`) rather than reusing `NONE` ("unknown, skip
  checking") — `NONE` meant every pointer-typed native call silently
  bypassed binary-op/declaration/argument checks that a known type would
  have gone through. `VAR_PTR`'s wildcard type-compatibility is
  deliberately erased at *every* pointer depth, not just one level like
  C's `void*` — documented as an intentional simplification (`StdrotParam`
  has no field to express a partially-erased pointer), not an accidental
  paraphrase of `void*`'s actual (single-level) conversion rule.
- Native-call checking (arity + argument types, via the same error machinery
  as user-defined function checks) reaches every expression position:
  declaration initializers (including nested calls, calls buried in an
  operand, braced array initializers, struct plain-expression initializers),
  assignments (both sides — RHS and a non-identifier LHS), array indices,
  `maxxing` (sizeof) operands, `bussin` (return) expressions, do-while
  conditions, switch discriminants/case values, and ordinary statement/
  condition position.
- **Argument enforcement is authoritative too, not just returns.** A
  fixed-type parameter's static check fails open whenever the *argument's*
  inferred type is `NONE` — most commonly a legacy `STDROT_ANY`-returning
  call used as an argument, since the analyzer genuinely can't know its
  type ahead of time. `enforce_arg_type()` (the argument-side counterpart
  to `enforce_return_type()`) re-verifies the declared contract at the
  runtime ABI boundary after `coerce_arg_to_param()`'s attempted
  conversion, so a native wrapper can trust its own `StdrotParam` blindly
  regardless of what the call site's static type inference could or
  couldn't determine. A pointer-typed declaration gets the same treatment:
  pointer-level is independently knowable even when the base type isn't
  (`enforce_return_type()` already guarantees a legacy `ANY` export can
  never return `STDROT_PTR`), so `rizz *p = a_legacy_any_call();` is
  rejected on the pointer-level mismatch alone instead of silently
  skipping all checking because the base type also happens to be unknown.
- **One shared rule for where a char narrows to int — now matching actual C
  semantics, not just internal self-agreement.** An earlier round made the
  runtime marshaller (`ast_expr_to_stdrot_value()`) and the static checker
  (`infer_expression_abi_type()`) consult one shared predicate
  (`stdrot_char_narrows_to_int()`, `stdrot.h`) instead of each maintaining
  an independent copy of the exception list — but that predicate itself
  narrowed array access and struct-member access to `int` unconditionally,
  which C does not do: `buf[0]` and `foo.c` are ordinary char lvalues, the
  same as a bare char variable, and are only promoted to `int` for an
  *unprototyped* call or the variadic tail of a variadic one — never
  merely for appearing as an argument to a normally-prototyped native
  call. Both sides agreed with each other, and both were wrong; a prior
  round's tests had enshrined the agreement as correct behavior.
  `takes_char(buf[0])` and `takes_char(foo.c)` now type-check and marshal
  as `STDROT_CHAR` successfully instead of being rejected. Unary
  operations got the same scrutiny: the predicate previously treated
  every `NODE_UNARY_OPERATION` identically regardless of which operator
  ran, which is too coarse — dereferencing a `char*` and pre/post
  increment/decrement all yield a value of the *operand's* type in C (no
  promotion of their own), while unary arithmetic negation genuinely does
  promote to `int`. The predicate is now keyed on the actual operator,
  not just the node shape. Fixing this exposed a second, previously-
  masked bug: `is_expression()` (`ast.c`) had no case for
  `NODE_UNARY_OPERATION` at all, silently falling to a default that could
  never match — meaning a dereference's marshalled type was always forced
  through the `NODE_UNARY_OPERATION` int fallback regardless of what the
  (now-correct) narrowing predicate said, until that gap was closed too.
  The same shared predicate also models `slorp<T>(T) -> T`'s identity
  return correctly for a char *array* (`yap buf[32]`), which marshals as
  `STDROT_STRING`, not `STDROT_CHAR` — `rant s = slorp(buf);` now type-
  checks and works; `STDROT_CSTRING` parameters accept a char-array
  argument for the same reason. `infer_expression_type()` also gained
  `NODE_ARRAY_ACCESS`/a static-only `NODE_STRUCT_ACCESS` path — both
  previously reported `NONE` unconditionally (the latter's existing path
  depends on a runtime-only `get_variable()` lookup, unavailable during the
  single static analysis pass), silently disabling argument checking for
  any native call passing one regardless of its actual element/field type.
- **String argument/return lifetime is now defined, enforced, and actually
  owned end to end — with ownership carried by the value, not a global.**
  Closes a real, reproduced-with-ASan use-after-free, the unbounded leak
  the first fix for it introduced, *and* a reentrancy bug the second fix
  introduced. An identity-style native (`T -> T`,
  `STDROT_EXPORT_SIG_IDENTITY`) returning one of its own string arguments
  unchanged (`identity(slorp(buf))`) used to hand back a pointer
  `execute_native_call()` was about to free as part of its own argument
  cleanup; materializing an independent copy fixed the UAF, but that copy
  initially had no owner and no free path (a leak once per call), and the
  fix for *that* used a single mutable global flag to track ownership —
  which cannot survive nesting. Evaluating one native call's own
  arguments can invoke other native calls (the ordinary case for
  `takes_cstring(identity(slorp(buf)))`, not a contrived one): the
  global, set true by the inner `identity` call, was never reset by the
  outer `takes_cstring` call once its own result turned out to be a
  non-string (`STDROT_INT`, from `strlen()`) — its materialize check is
  gated on `result.type == STDROT_STRING`, so a non-string result never
  touched the flag at all. Whatever consumed `takes_cstring`'s result
  next would then read a stale `true` and reinterpret that `STDROT_INT`'s
  union bits as a heap pointer, handing an essentially-arbitrary address
  to `free()`. Fixed by making ownership part of the result itself:
  `execute_native_call()` now returns a `NativeResult { StdrotValue
  value; bool owns_string; }` (`stdrot.h`), and every consumer —
  the peek/consume memo cache (`native_call_peek()`/`native_call_consume()`,
  which now caches the whole `NativeResult`), `marshal_native_return_value()`
  → `handle_function_call()`'s `VAR_STRING` case, and the deprecated
  write-back path in `execute_func_call()` (a single catch-all free at
  the end of that function, covering every way it can discard a result)
  — reads its own local `NativeResult`, never a value some other,
  possibly-nested call last wrote. Reproduced the reentrant bug first
  (confirmed via `git stash`-based A/B testing that the pre-fix code
  crashes in `safe_free()`, reading a bogus pointer built from an
  integer's bit pattern, exactly as predicted) before shipping the fix,
  and verified 0 leaks/0 errors afterward under both valgrind and ASan.
  `STDROT_CSTRING` arguments are documented as strictly non-escaping (own
  comment, `stdrot_api.h`); an escaping-string annotation (a native
  retaining a `const char *` past its own return) remains explicitly out
  of scope for this phase — documented as a real, tracked gap in
  `docs/ROADMAP.md`, not a silently-dropped requirement.
- **`return_like_arg` may only reference a `STDROT_ANY` parameter.**
  Identity polymorphism (`T -> T`) only means something coherent if the
  referenced parameter carries no fixed representation to coerce into —
  static inference reports the *source* expression's type (before
  coercion), while runtime enforcement sees the argument *after*
  parameter coercion, so a fixed-type parameter (`STDROT_CSTRING`,
  `STDROT_DOUBLE`, `STDROT_PTR`, ...) meant static and runtime could
  disagree about the call's actual result type — reopening the exact
  unsupported-CSTRING-return path a fixed typed signature would reject
  outright, silently changing a numeric argument's coerced type, or
  losing pointer-level tracking entirely. `validate_native_registry()`
  now rejects any identity descriptor whose referenced parameter isn't
  `STDROT_ANY` at load time.
- **A failed `STDROT_CSTRING` argument conversion no longer produces a
  "valid" `CSTRING` with a null payload.** `stdrot_string_to_cstring()`
  can genuinely return `NULL` — not just true OOM, but any string past
  `MAX_ALLOC_SIZE`, reachable by an ordinary program — and
  `coerce_arg_to_param()` used to tag the value `STDROT_CSTRING` (with a
  NULL `.val.cstr`) regardless, which `enforce_arg_type()` then approved:
  an allocation failure silently became a null-pointer dereference inside
  whatever native trusted its own declared signature (e.g. `strlen()` on
  the argument). The value now keeps its original `STDROT_STRING` tag on
  conversion failure, so `enforce_arg_type()` reports a genuine, honest
  type mismatch and exits cleanly instead.
- **Registry descriptors are validated once, at load time, for structural
  coherence** — instead of a malformed one surfacing piecemeal depending
  on which call path happens to touch it. `min_args`/`param_count`
  coherence, `params` required whenever `param_count > 0`,
  `return_like_arg` naming a genuinely mandatory `STDROT_ANY` argument,
  `fn` non-null, non-negative `pointer_level` on the return type and
  every parameter, no two entries exporting the same Brainrot-visible
  name (previously `get_native_function()` just returned the first
  linker-section match, so a colliding generated binding would silently
  "win" by registration order), `pointer_level` nonzero only alongside
  `STDROT_PTR` (a nonzero level on any other base type is internally
  contradictory: static checking would approve it as an ordinary pointer
  to that base type, but the runtime marshaller tags ANY expression with
  pointer_level > 0 as `STDROT_PTR` regardless of declared base type, so
  every call the "valid" descriptor claimed to describe would fail
  `enforce_arg_type()`/`enforce_return_type()`), no `STDROT_NONE`
  ("void")-typed parameter (can't coherently consume an argument — zero
  arguments are already expressed via `param_count == 0`), and every
  `StdrotType` value actually in range (a hand-written `(StdrotType)999`
  shouldn't graduate from validation and become somebody else's problem
  later). **This validates structural coherence, explicitly not "every
  capability this descriptor could describe is implemented."**
  `STDROT_HANDLE` (any position) and `STDROT_CSTRING`-as-a-*return*-type
  remain real, well-formed `StdrotType` values — reserved groundwork for
  capabilities this ABI hasn't finished — and a descriptor using either
  still loads successfully; `semantic_check_native_call()` rejects any
  *call* to it before that call could reach a marshaller with no code
  path to honor it. Collapsing "structurally coherent" and "fully
  implemented" into one load-time check would mean a single `.so`
  exporting one reserved-but-unimplemented capability refuses to load at
  all, taking every unrelated, fully-supported native down with it for a
  call that might never happen — a deliberate two-tier design, documented
  as such directly in `validate_native_registry()`'s own comment, not
  papered over with an inaccurate "trusted after this point" claim. A
  broken descriptor now aborts with a clear diagnostic the moment the
  registry loads — covered by eight dedicated malformed `.so`s under
  `tests/badnatives/`, each built and asserted to abort with the right
  message (`tests/test_brainrot.py`'s `test_bad_registry_rejected_
  at_load`), not just reasoned about.
- **C's default argument promotions now apply to a native's variadic
  tail, centrally, at the ABI boundary — separately from the fixed-
  parameter char handling above.** A fixed, prototyped parameter is
  "converted as if by assignment" to its declared type (no promotion,
  which is exactly why `takes_char(buf[0])` above correctly stays char);
  an argument beyond a variadic native's `param_count` (e.g. `yapping`'s
  format-string tail) has no prototype to convert *to*, and C's own
  variadic calling convention promotes every such argument
  (`char`/`_Bool`/`short` → `int`, `float` → `double`) regardless. Fixing
  the fixed-parameter char behavior without also implementing this half
  turned into a real regression: `yapping("%d", c)` started marshalling
  `c` as `STDROT_CHAR`, which `process_yapping_format()`'s `"%d"` family
  (`stdrot/yapping.c`) never accepted (only INT/SHORT/BOOL) — nothing got
  formatted. `apply_variadic_promotion()` (`stdrot.c`) now runs once,
  centrally, on every argument at index >= `param_count` for a variadic
  native, so any variadic consumer can trust the canonical promoted
  representation instead of a formatter-specific patch understanding
  every legal tagged Brainrot type individually.
- **`infer_expression_abi_type()` now normalizes an enum-typed expression
  to `VAR_INT`, matching what the runtime marshaller actually produces.**
  `StdrotType` has no `ENUM` member; `ast_expr_to_stdrot_value()` always
  marshals an enum-typed expression as `STDROT_INT` (the same
  representation a bare int gets). The function whose entire job is
  reporting "the type a native-ABI marshalling call will actually
  produce" previously left `VAR_ENUM` untouched, only normalizing
  `VAR_CHAR`-related shapes — so `identity<T>(T)`'s static `T` for an
  enum *variable* argument was reported as `VAR_ENUM` while the runtime
  argument/result was always `STDROT_INT`, the same class of static/
  runtime representation gap this whole phase exists to close. Fixed;
  regression-tested with an enum *variable* specifically (an enum
  *constant* already resolves to plain `VAR_INT` via a separate path and
  would have hidden the gap entirely).
- **Static struct-field type inference is now fully recursive**, matching
  the runtime resolver's support for a chain of any depth. The previous
  static fallback for `infer_expression_type()`'s `NODE_STRUCT_ACCESS`
  case only resolved a *direct* identifier object (`f.c`); a chained
  access like `o.inner.c` has an object that's itself a
  `NODE_STRUCT_ACCESS`, so it fell straight through to `NONE` — silently
  disabling argument type-checking for that shape
  (`semantic_check_native_call()`'s `actual == NONE` fallthrough), the
  same class of hole this whole phase exists to close. A new
  `infer_struct_def_static()` helper resolves the object of any chain
  depth purely off the analyzer's own symbol table and the struct-
  definition registry (never runtime storage, unavailable during this
  single static pass) — mirroring, but independent of, the mutating
  recursive resolution `semantic_analyze_with_scope_tracking()`'s own
  `NODE_STRUCT_ACCESS` case already performs elsewhere in the same file
  (independent because `semantic_check_native_call()` runs on a call's
  arguments *before* that case's own argument-visiting loop reaches
  them, so relying on its cached `node->var_type` isn't sound here).
- **A non-char array can no longer be smuggled into a scalar parameter as
  if it were a raw value, on either side of a native call's `param_count`
  boundary.** `Variable`'s own value union (`ast.h`) aliases a scalar's
  storage with an array's backing pointer, so `infer_expression_type()`
  reports the same `VarType` for `rizz arr[2]` as for a scalar `rizz x` —
  and `ast_expr_to_stdrot_value()` would read that union unconditionally,
  reinterpreting the array's backing pointer as the scalar value the
  descriptor declared. Only the `STDROT_ANY` branch had a guard against
  this; an ordinary FIXED scalar parameter did not, so `ragequit(arr)`
  (declared `STDROT_INT`) type-checked and handed the native's C
  implementation an address disguised as an integer. The new shared
  `is_unmarshallable_array_arg()` helper (`semantic_analyzer.c`) now
  guards both the `STDROT_ANY` branch (replacing its previous inline
  copy of the same check) and every fixed scalar parameter. The same
  hazard existed on the *other* side of `param_count` too — a variadic
  native's unchecked tail skipped array-shape validation entirely
  (`yapping("%d", arr)` would format a raw address), even though
  "unchecked type" was never meant to mean "unchecked representation
  validity"; that tail is now validated with the same helper.
- **C's default argument promotions apply only to a genuine C-style
  variadic tail, not to every `is_variadic` native.** `is_variadic` means
  two unrelated things: a real typed signature with a promotable C
  variadic tail (`yapping`'s format arguments), and a legacy/untyped
  `STDROT_EXPORT()` export whose actual signature is simply unknown
  (arity left unchecked for backward compatibility, not because it has C
  varargs semantics). The variadic-promotion fix above briefly attached
  promotion to the `is_variadic` flag itself, which silently changed
  every legacy export's observable argument types (`STDROT_CHAR` →
  `STDROT_INT`, `STDROT_BOOL`/`STDROT_SHORT` → `STDROT_INT`, `STDROT_
  FLOAT` → `STDROT_DOUBLE`) without recompiling or touching that
  binding's own source — an ABI break for any wrapper that switches on
  `args[i].type`. Fixed with an explicit descriptor field,
  `StdrotEntry.promote_variadic_tail`, set only by a new
  `STDROT_EXPORT_SIG_VARIADIC()` macro (migrated `yapping`/`yappin`/
  `baka` to it) — never inferred from `is_variadic`/`param_count`/
  `return_type` shape, the same discipline `return_like_arg` and
  `STDROT_EXPORT_SIG_IDENTITY()` already established for identity
  polymorphism. `validate_native_registry()` additionally rejects
  `promote_variadic_tail == true` with `is_variadic == false` (there's no
  tail to promote). A new legacy test native, `legacy_type_probe`,
  actually inspects `args[0].type` and reports it back — unlike every
  prior legacy test export, which ignored its arguments entirely and so
  could never have caught this regression.
- **The `stdrot.c` ↔ `libstdrot.so` boundary now has an explicit ABI
  version, instead of a struct/enum layout that changed underneath an
  unversioned entrypoint.** Before this PR, the loader dlsym'd
  `stdrot_get_api()` and trusted whatever `StdrotAPI{ StdrotEntry
  *functions; int count; }` came back. This PR changed that layout in
  every dimension that matters: `StdrotType` gained `STDROT_ANY` at
  index 0 (renumbering every existing value), `StdrotEntry` grew from
  `{name, fn}` into the full typed-signature descriptor, and the
  registry section now holds *pointers* to entries rather than entries
  themselves. `dlopen()`-by-bare-filename and `make install` mean an old,
  un-rebuilt `libstdrot.so` is exactly as reachable at runtime as a fresh
  one — and the old loader would have called the old library's
  `stdrot_get_api()` under the SAME name and silently reinterpreted its
  actual memory (e.g. an old entry's own `name` pointer, `"yapping\0"`)
  as this version's shape, no detection at all. Fixed with an explicit
  `STDROT_ABI_VERSION` (`stdrot_api.h`) and a renamed entrypoint,
  `stdrot_get_api_v2()`: an old `.so` simply doesn't export that symbol,
  so `dlsym()` fails the lookup cleanly and `stdrot_load()` exits with
  "rebuild libstdrot.so" instead of dereferencing unfamiliar memory.
  Verified against `tests/old_abi_sim/fake_pre_v2_registry.so`, a
  permanent fixture built standalone (no dependency on the current
  `stdrot_api.h`) to genuinely only export the old symbol under the old
  layout — and that verification caught a second, real, pre-existing bug
  along the way: the missing-symbol error path called `dlclose(lib_
  handle)` without resetting `lib_handle` to `NULL`, so the `atexit(std
  rot_unload)` handler registered at program start (`lang.y`) closed the
  same, already-closed handle again on the way out — a double-`dlclose`
  use-after-free inside glibc's own loader bookkeeping, invisible until
  valgrind actually exercised this exact path for the first time.
- **A variadic native's unchecked tail is now checked for representation
  validity, not just the one array-aliasing bug found last round.** The
  prior round's fix only asked "is this argument an unmarshallable
  array"; the real invariant `STDROT_ANY`'s own accepted-type check
  already enforces is broader: does `ast_expr_to_stdrot_value()` have
  *any* honest `StdrotValue` representation for this expression at all.
  A struct identifier has no `VAR_STRUCT` case in that function's
  `NODE_IDENTIFIER` switch, so it silently produces `STDROT_NONE` —
  `yapping("%d", f)` (a struct) passed the array check and handed the
  formatter "no value" where source code plainly supplied one. Fixed by
  generalizing `is_unmarshallable_array_arg()` into `is_unmarshallable_
  expr()`, reusing the same closed accepted-type set `STDROT_ANY` already
  defines (a pointer-level expression is *not* flagged here, unlike
  `STDROT_ANY`'s own stricter rule, since it has a real representation —
  `STDROT_PTR` — that `STDROT_ANY` specifically declines to accept for
  unrelated reasons). One shape is undetectable statically no matter
  what: an argument that's itself a call to a legacy `STDROT_ANY`-
  declared export whose real, only-known-at-runtime return happens to be
  `STDROT_NONE` (a legacy void-returning export) — `execute_native_call()`
  now rejects a variadic-tail argument that marshals to `STDROT_NONE` at
  the runtime ABI boundary, the same way `enforce_arg_type()` already
  rejects a fixed argument that turns out not to satisfy its declared
  type after all. The diagnostic wording is deliberately representation-
  oriented ("has no supported native ABI representation"), not "variadic
  argument" — the prior wording conflated a genuine C-variadic tail with
  a legacy export's unrelated reason for leaving arity unchecked, exactly
  the distinction `promote_variadic_tail` above exists to keep separate.

### String-return ownership, static native-call typing, and one more ABI hole

- **Every `STDROT_STRING` return is now materialized unconditionally**,
  closing a second, different use-after-free the previous round's alias-
  specific fix didn't cover. That fix only recognized an alias into
  `owned_string_bufs[]` (a nested native call's own `STDROT_STRING`
  result); `coerce_arg_to_param()`'s `STDROT_CSTRING` conversion allocates
  an *independently* tracked buffer per argument
  (`owned_cstring_bufs[]`), and a typed native pointing its `STDROT_STRING`
  result straight at that buffer sailed through unmaterialized, got freed
  by that array's own cleanup loop, and the caller copied from already-
  freed memory — reproduced under ASan with the exact free-then-read
  sequence the pattern predicts. Rather than extend the alias search to
  also check the other scratch array (and be one undiscovered buffer
  type away from the next version of the same bug), `execute_native_call()`
  now deep-copies *every* `STDROT_STRING` result before releasing any
  argument scratch, unconditionally — which buffer (if any) it happened
  to alias no longer matters.
- **`sizeof`/`maxxing` no longer executes a native call to discover its
  type.** Phase 2 gave every native a real, registered signature, so a
  TYPED native's return type is knowable from `StdrotEntry.return_type`/
  `return_like_arg` alone — but `get_expression_type()`'s `NODE_FUNC_CALL`
  case still unconditionally invoked the native (via the type-probe
  cache) to answer that question, meaning `maxxing(slorp(x))` actually
  read from stdin, inside an operator whose defining property in C is
  that its operand is never evaluated. A new `get_native_call_static_type()`
  (`ast.c`) answers this from the descriptor alone, with no execution;
  `get_expression_type()`'s own `NODE_FUNC_CALL` case tries it first
  (falling back to actually invoking only for a genuinely unknowable
  legacy/untyped export, safe there specifically because that caller was
  always going to evaluate the call for its value anyway); `handle_sizeof()`
  uses it directly and does **not** fall back — a legacy call as sizeof's
  direct operand is now a semantic error ("sizeof's operand is never
  evaluated, so this native cannot be invoked to discover it"), not a
  silent execution. Verified by confirming stdin is read exactly once
  (the real, statement-position call), not twice.
- **The native-call type-probe cache no longer has a fixed capacity that
  can silently change what a program means.** The cache existed because
  a native call's type was only knowable by running it; once 16
  simultaneously-pending type-probes (e.g. a long chain of native-call
  operands in one expression — `get_expression_type()`'s `NODE_OPERATION`
  case recursively probes every leaf before any is evaluated) filled every
  slot, the 17th probe silently degraded to `STDROT_NONE`/unknown, which
  could change a chain's *promoted* arithmetic type depending purely on
  how many other calls happened to be pending at the same time — a
  fractional `STDROT_DOUBLE` result could get promotion-truncated to an
  int for no reason related to the expression's actual content. Fixing
  the point above removes most of the remaining pressure (a typed native
  no longer touches this cache at all just to answer a type query), but
  the cache itself now also grows dynamically instead of capping out, so
  no chain length can distort the result regardless. (The dynamic array
  needed its own `atexit`-registered cleanup, the same lazy-registration
  pattern `stdrot.c`'s own pending-argument cleanup already uses — the
  old fixed-size cache lived in `.bss` and needed none; verified via
  valgrind that converting it to a heap allocation didn't introduce a
  leak of its own.)
- **`validate_native_registry()` now validates the `StdrotAPI` table
  itself before ever indexing into it**, not just each entry once inside
  it. A negative `count` previously made the validation loop execute zero
  times, silently accepting a nonsensical table as if it legitimately
  exported nothing; a positive `count` paired with a `NULL` `functions`
  pointer would have crashed the validator itself on its first iteration,
  before it ever got to check anything. Also bounds `count` to a
  generous-but-finite upper limit, so a corrupt or hostile
  `stdrot_get_api_v2()` implementation can't send this loop walking
  through arbitrary out-of-bounds memory one `StdrotEntry` pointer at a
  time.
- **Cleaned up three fossils in the public ABI documentation/dead code**:
  `stdrot_api.h`'s registry-entry comment still claimed the pre-linker-
  section contract (`extern StdrotEntry stdrot_exports[]; extern int
  stdrot_export_count;`) instead of describing `stdrot_get_api_v2()`;
  a `takes_char` test-native comment still described the *previous*
  (now-fixed) char-narrowing bug as correct behavior, contradicting the
  regression fixture sitting right next to it; and `stdrot.h`/`stdrot.c`
  carried a `slorp_string()` bridge stub whose declared signature never
  matched `stdrot/slorp.c`'s actual implementation — confirmed genuinely
  unreachable from anywhere in the codebase and deleted outright, rather
  than left as a documented-but-still-present latent bug in a PR whose
  entire point is making this ABI trustworthy.

### One canonical execution-free expression type, and closing the VAR_VOID/NONE conflation

Round 18 found that the previous round's "static type" fixes were narrow,
shape-specific patches rather than a real solution, and both of the classes
of bug this PR keeps closing had crept back in through them:

- **`NONE` meant both "genuinely unknown" and "known to be void."**
  `stdrot_type_to_vartype(STDROT_NONE)` returned plain `NONE`, the exact
  same sentinel a genuinely-unknowable legacy `STDROT_ANY` export also
  reports — so a TYPED void native (`yapping`, not a legacy export) read
  as "can't tell, don't block it" everywhere a `NONE` result short-
  circuits a check. `rizz x = yapping("hi");` type-checked. A new
  `VAR_VOID` (`ast.h`) splits this apart the same way `VAR_PTR` already
  did for pointers in an earlier round: `NONE` now means only "I don't
  know," `VAR_VOID` means "I know precisely — nothing."
  `check_type_compatibility_ex()` rejects `VAR_VOID` against everything,
  including itself, so this is now caught statically, not just by the
  runtime "native call result (void) cannot be used" checks that still
  handle the genuinely-unknown legacy case.
- **`sizeof` still executed a native call nested inside a larger
  expression.** The previous round's fix only covered a bare native call
  as `sizeof`'s *direct* operand; `maxxing(legacy_int() + 1)` still fell
  through to `get_expression_type()`'s ordinary recursive descent, which
  executes a legacy/untyped call to learn its type. Replaced with
  `infer_runtime_expression_type_noeval()` (`ast.c`), a genuinely
  recursive, execution-free type query with no path anywhere in its call
  graph to `execute_native_call()` — an unknowable call at any depth
  propagates `NONE` up through the whole containing expression instead of
  being resolved by running it. `handle_sizeof()` now gates on this over
  the *whole* operand before ever calling `get_expression_type()`.
  Verified with a new legacy test native that prints an unmistakable
  marker as a side effect (`legacy_int_prints()`,
  `tests/stdrot/testabi.c`): `maxxing(legacy_int_prints() + 1)` is
  rejected with no marker printed.
- **Identity-argument resolution reported source type, not ABI type.**
  `get_native_call_static_type()`'s `return_like_arg` case called plain
  `get_expression_type()` on the referenced argument — reopening, in a
  new guise, the exact source-vs-ABI-type bug this PR has closed multiple
  times before: a char array marshals as `STDROT_STRING`, not its element
  type; a bare char literal marshals as `STDROT_CHAR`, not
  `get_expression_type()`'s own (differently-motivated) `VAR_INT` rule
  for `NODE_CHAR`. A new `infer_runtime_expression_abi_type_noeval()`
  (`ast.c`), built on the function above, mirrors
  `infer_expression_abi_type()` (`semantic_analyzer.c`) exactly — sharing
  its `stdrot_char_narrows_to_int()` rule so the static and runtime sides
  can't independently drift — and is what identity resolution now
  consumes instead. Confirmed via a temporary A/B test (reverting to
  `get_expression_type()`, rebuilding) that `maxxing(identity('a'))`
  silently reported `4` (int) instead of `1` (char), and
  `maxxing(identity(buf))` for a char array silently reported `1` instead
  of correctly rejecting.
- **The native-call cache's destructor freed the table, not the owned
  values still inside it.** `free_native_call_cache()` assumed no entry
  ever needed per-entry cleanup; a valid, unconsumed entry holding an
  owned `STDROT_STRING` (every string result is now materialized
  unconditionally, per the previous round) would leak it. Now frees any
  such string before freeing the array.
- **`handle_sizeof()`'s switch had no `VAR_PTR` case**, despite
  `get_type_size_for_descriptor()` immediately underneath already
  correctly handling `pointer_level > 0` — a pointer-typed native result
  (`STDROT_PTR`, end to end) fell to "Invalid type in sizeof" instead of
  reporting `sizeof(uintptr_t)`.
- **`is_expression()`'s `NODE_FUNC_CALL` case still executed a TYPED
  native unconditionally** via `native_call_peek()`, inconsistent with
  `get_expression_type()`'s own static-first fix from the previous round.
  Now delegates to `get_expression_type()`. Also cleaned up the native-
  call-cache comment block, which still claimed a native call has no
  static return type at all — no longer true for anything typed.

### VAR_VOID's invariant now holds everywhere a value is required, not just at declaration

Round 19 found that round 18's fixes were each individually correct but had
branch-local jurisdiction: a few specific contexts enforced the new
invariants, but the invariants themselves weren't enforced structurally.

- **Arithmetic could still launder a `VAR_VOID` result into a real
  value, and `if`/`while` conditions didn't require a value at all.**
  `yapping("hi") + 1` type-checked: `validate_binary_operation()`'s
  arithmetic-type check only ever flagged `VAR_STRING`/`VAR_BOOL` as
  clearly incompatible, so `VAR_VOID` fell through to `return true`, got
  widened to `VAR_INT` by the ordinary arithmetic-promotion rule, and the
  declaration type-checked as `int = int` — `yapping()` actually
  executed before a much weaker *runtime* check caught the misuse.
  `if`/`while` conditions had the identical gap: each recursively visits
  its condition expression but never asked whether it produced a value
  at all. A single shared check, `require_value_expression()`
  (`semantic_analyzer.c`), closes this everywhere at once — both
  operands of every binary operation, and the condition of `if`/`while`/
  `do`-`while`/a `switch` discriminant — instead of teaching each
  context individually that void is bad. Confirmed via a temporary A/B
  test that the old code genuinely *executes* `yapping()` (observable
  side effect) before its weaker runtime check fires; the fix rejects
  all of these statically, before anything runs.
- **The zero-argument call path bypassed the string-return ownership
  contract the general path enforces.** `execute_native_call()`'s
  `total_args == 0` branch never materialized a `STDROT_STRING` result —
  reasoning ("no argument buffers to alias") that predates the previous
  round's fix making that materialization unconditional for every other
  call shape. Factored the materialization into a shared
  `finalize_native_string_result()`, used by both the zero-argument and
  general paths. Confirmed via a temporary A/B test that the pre-fix
  code crashes under ASan with a genuine heap-use-after-free: two new
  test natives share a heap buffer, one frees and reallocates it as a
  side effect, and `get_expression_type()`'s existing dual-operand peek
  (documented in an earlier round) peeks the string-returning call
  first and the mutator second, before either result is consumed —
  reading the first afterward read already-freed memory.
- **`sizeof` of a `VAR_STRING` expression was inconsistent depending on
  the AST shape it came from**, even though `get_type_size_for_
  descriptor()` has always defined a size for it. A plain string
  identifier's `sizeof` already used that definition; `handle_sizeof()`'s
  generic expression-type switch simply had no `VAR_STRING` case, so a
  native call resolving to `VAR_STRING` (e.g. an identity call on a char
  array) fell to "Invalid type in sizeof" for no reason connected to
  strings actually lacking a size. Added the missing case; the previous
  round's char-array identity fixture, which had enshrined that
  rejection as correct behavior, now asserts the correct
  `sizeof(String)` instead.

### Closing the NONE/void grammar gap and applying require_value_expression() by category, not by fixture

Round 20 found that round 18/19's invariants held everywhere the machinery
itself ran, but not always at the edges feeding into it.

- **`skibidi` (void) still mapped to `NONE` in the grammar, not
  `VAR_VOID`.** `type: SKIBIDI { $$ = NONE; }` meant a user-defined void
  function's statically-known return type collapsed back into
  "genuinely unknown" everywhere `NONE` means "can't tell, fail open" —
  e.g. a variadic native argument's representability check waved a void
  function's call through using the exception meant for a legacy
  `STDROT_ANY` export's real dynamic typing. `skibidi` now maps to
  `VAR_VOID` — in both the grammar rule and, separately, the lexer's
  `current_var_type` side effect a declaration-with-initializer's AST
  node actually gets its type from (`skibidi` was the one type keyword
  that never set it). Making void reachable at declaration/parameter/
  field positions, not just function returns, meant it also needed
  rejecting there: `semantic_visit_declaration()` now rejects a
  `VAR_VOID`-typed declaration unconditionally, and
  `create_default_node()` has its own case instead of a generic
  "unsupported type" fallback. Confirmed via a temporary A/B test that a
  void function's call previously passed static argument-
  representability checking and was only caught by the runtime ABI's
  own `STDROT_NONE` check.
- **`require_value_expression()` wasn't wired into every context that
  actually consumes a value.** An array index, a switch case value, and
  every unary operator other than address-of/dereference (which already
  had their own checks) all recursively visited their operand but never
  required it to produce one. Added to all three.
- **`baka(...)`'s argument was a semantic-analysis blind spot.**
  `NODE_ERROR_STATEMENT` had no case in the scope-tracking switch at
  all, falling to the generic default — so a native call nested inside
  `baka`'s argument never reached `semantic_check_native_call()`,
  tunneling a statically-detectable ABI violation
  (`baka(yapping(42))` — `yapping`'s fixed format parameter is a
  string) past semantic analysis entirely. Added the missing case.
  Confirmed via a temporary A/B test that the nested violation was
  previously caught only by runtime ABI enforcement.
- **`finalize_native_string_result()` (added last round) assumed its
  own copy always succeeded.** `safe_strdup()` can genuinely return
  `NULL` data on allocation failure, and the pre-existing implementation
  proceeded straight to `memcpy()` regardless — a NULL-pointer write on
  OOM. Fixed `safe_strdup()` to check its own allocation before copying;
  `finalize_native_string_result()` now degrades the result to
  `STDROT_NONE` on failure (which already maps to `VAR_VOID`) instead of
  claiming ownership of a copy that doesn't exist, so every value-
  consuming context this PR now hardens already rejects it correctly.

### Separating void from void*, and completing pointer_level's static resolution

Round 21 found that mapping `skibidi` to `VAR_VOID` (round 20) traded one
sentinel collision for another: `VAR_VOID` was now used both for "no
value" and as the base type of `void *`, and several consumers only asked
about the base type, not the `(base type, pointer_level)` pair this
codebase's own type-compatibility model already uses.

- **`require_value_expression()` rejected `VAR_VOID` unconditionally.**
  `(VAR_VOID, 0)` is void — no value, ever; `(VAR_VOID, 1)` is `void *` —
  a real pointer value. `skibidi *p = test_ptr_source();` used as a value
  afterward was rejected as if `p` held nothing. Fixed to also check
  pointer level. A second bug showed up alongside: `create_default_node()`
  decided `VAR_VOID` was illegal before it even had pointer_level to look
  at, so `skibidi *p;` (no initializer) crashed at parse time instead of
  reaching the ordinary (if unsupported for any pointer type today)
  semantic rejection `rizz *p;` alone already gets.
- **A pointer-typed struct field lost its pointer level during native-call
  checking.** `infer_expression_pointer_level()` had no
  `NODE_STRUCT_ACCESS` case, falling back to a field's `node->
  pointer_level` — unpopulated at the point native-call checking runs,
  since that visit happens afterward. A `rizz *p;` field was falsely
  rejected against a `STDROT_PTR` parameter and falsely *accepted*
  against a scalar one, the mismatch surfacing only at runtime ABI
  enforcement. Fixed with the same static struct-definition resolution
  discipline the base-type inference already has — and, once static
  checking stopped masking it, `evaluate_expression_pointer()` needed the
  identical fix on the runtime side (it was reading a NULL pointer for
  a struct-field argument instead of the field's real value).
- **Named by-value void parameters and struct fields were still legal.**
  Round 20 rejected void variable declarations but not these two other
  declaration-like positions — a `skibidi hole;` struct field silently
  got `sizeof(int)` of layout via a defensive fallback meant for a
  different failure mode. Rejected both, scoped to pointer_level == 0
  only — `void *` parameters and fields are real, differently-sized
  types and are untouched.
- **String materialization failure retagged a valid return as void.**
  `finalize_native_string_result()` (added last round) degraded a
  failed copy to `STDROT_NONE` — but the native genuinely returned a
  valid `STRING`; only the adapter's own copy failed. Now reports and
  exits, the same severity class as every other ABI enforcement failure
  in `stdrot.c`, instead of silently turning a validated return value
  into a different type.

### The last projections from (base type, pointer_level) down to base type alone

Round 22 found the remaining places that still collapse a type down to its
base `VarType` before making a decision, now that `skibidi *` is a real,
reachable source-language type.

- **User-defined pointer returns were boxed by base type, not pointer
  depth.** `handle_function_call()` switched on the return's base
  `VarType` first, with pointer-boxing special-cased only inside the
  `VAR_INT`/`VAR_ENUM` branches. A function returning `skibidi *`
  (any base type, `pointer_level > 0`) fell into `VAR_VOID`'s own
  "structurally unreachable" case and returned `NULL` despite genuinely
  having a value; a `chad *`/`yap *`/`cap *`/`smol *` return fell into
  that base type's ordinary scalar case, reinterpreting a raw address as
  a scalar value — confirmed this actually crashes (an ASan SEGV reading
  a float through a mis-typed box). Fixed by checking `pointer_level`
  *before* the base-type switch, matching the rule the native-call
  marshaller already enforces. A second regression from the same
  underlying cause: `handle_return_statement()`'s return-value switch
  still only had a case for `NONE`, a comment left over from before
  `skibidi` meant `VAR_VOID` — an explicit `bussin 0;` inside a real
  void function crashed outright.
- **A pointer array element lost its pointer level during native-call
  checking** — the identical bug struct fields had last round, this time
  in `infer_expression_pointer_level()`'s `NODE_ARRAY_ACCESS` case, which
  never consulted the analyzer's own symbol table. Fixed with the same
  `find_symbol()`-first discipline the identifier case already has.
- **`void *` is a real value, but dereferencing one is still invalid** —
  there's no pointee type to read or write through. Added the check in
  both places a dereference is validated (the general unary case, and
  assignment's separately-duplicated dereference-LHS check), scoped
  specifically to a dereference that would *produce* `(VAR_VOID, 0)` —
  `void **` still dereferences fine, down to `void *`.
- **A `bussin` (return) expression was traversed but never compared
  against the enclosing function's declared return type.**
  `rizz terrible() { bussin yapping("oops"); }` passed semantic analysis
  outright, since `yapping()`'s void return was never checked against
  `terrible`'s declared `int`. Added the same compatibility check
  declaration initializers already get, skipped for struct returns
  (already has its own dedicated check) and genuinely void-declared
  functions (a `bussin`'s value there is deliberately ignored, matching
  existing behavior).

### Pointer-array storage corruption, opaque-pointer erasure, and the missing return producers

Round 23 found a genuine memory-corruption bug hiding behind an under-
tested fixture, plus two more places the runtime hadn't caught up with
the typed model.

- **A pointer-typed array's element address used the wrong stride.**
  `evaluate_multi_array_access()` computed each element's address using
  the array's base `VarType` width (`sizeof(int)`, `sizeof(char)`, ...)
  even though storage had already been allocated at `sizeof(uintptr_t)`
  per element. Element 0 starts at byte 0 under either stride, which is
  exactly why the previous round's own fixture (testing only element 0)
  passed while the bug was still live — element 1 onward lands partway
  inside a neighboring element's own bytes. Confirmed this is real
  memory corruption, not just a wrong value: a temporary A/B test (`git
  stash` to the pre-fix commit) crashed under ASan with a misaligned
  store followed by a SEGV. Fixed with the same pointer_level-dominates
  rule used everywhere else pointer representation is decided.
- **`VAR_PTR` (this codebase's own opaque native-pointer type) sailed
  past the void-pointer dereference/arithmetic rejections added last
  round.** `*test_ptr_source()` let whichever scalar evaluator consumed
  it silently decide the erased pointee's representation after the
  fact — exactly backwards for a typed ABI, where the descriptor is
  supposed to be authoritative. Pointer arithmetic on the same shape
  silently defaulted an unknown element size to 1 byte. Both now
  rejected for `VAR_PTR` alongside `VAR_VOID`, scoped to pointer_level
  == 1 — a multi-level opaque pointer (`void **`) still works, landing
  back at a real opaque pointer value.
- **The return-checker (added last round) could approve a `rant`/`yap`
  return the runtime still couldn't produce.** `handle_return_
  statement()`'s producing switch had no `VAR_CHAR`/`VAR_STRING` case —
  a statically-approved `rant f() { bussin "hello"; }` crashed with
  "Unsupported return type" the moment it actually ran. Fixing this
  exposed a narrower, genuinely new leak: a user-defined function is
  invoked twice for a call embedded in a declaration initializer (a
  pre-existing, documented gap — it has no memo cache the way a native
  call does), and the first call's owned string result was never freed
  — something that could never surface before, since no other return
  type owns a separate heap allocation. Closed by extending the
  existing "clear whatever's still pending before this slot gets
  overwritten" cleanup to also cover an unconsumed owned string.

### Return-channel reentrancy, pointer-array braced initialization, and ownership comments

Round 24 found that `current_return_value` — a single global slot — was
acting simultaneously as a nested callee's result and the caller's own
in-progress return frame, plus a second instance of round 23's
pointer-array stride bug hiding in the braced-initializer path.

- **A nested call inside `bussin` could steal the outer function's
  return type.** `handle_return_statement()` set the global return
  slot's type/pointer_level to the enclosing function's own declared
  return type, then evaluated the return expression, then wrote only
  the value afterward. If evaluating the expression invoked *another*
  call — native or Brainrot-defined — that nested call's own completion
  overwrote the same global slot's type for its own duration, and
  nothing restored the outer function's before the final value got
  written: `gigachad widen() { bussin takes_cstring("abc"); }` correctly
  computed the converted double payload, but the caller then read the
  type left behind by the *nested* call and consumed the wrong union
  member entirely — the exact tag/payload disagreement `NativeResult`
  already eliminated at the native-ABI boundary, one layer up. Confirmed
  via a temporary A/B test (`git stash` to the pre-fix commit) that both
  a nested native call and a nested user-defined call printed
  `0.000000` instead of the correct `3.000000`. Fixed by evaluating the
  return expression into a local first, and only installing type,
  pointer_level, and value together once evaluation is complete.
- **Pointer-array braced initialization still used the old scalar
  representation.** Round 23 fixed pointer_level dominating element
  representation for *reading* an array element back out and for
  *allocating* a pointer array's storage — but the one remaining place
  that *writes* elements (a braced initializer, `rizz *ptrs[2] = { &x,
  &y };`) still dispatched on the array's base type, writing through
  the wrong stride and evaluating `&x` with the wrong evaluator
  entirely. Confirmed this crashed under ASan with a SEGV before
  printing anything; fixed with the same pointer_level-dominates check
  used everywhere else.
- Renamed `free_pending_struct_return_value()` to `free_pending_return_
  value()` — it stopped being struct-only the moment round 23 gave
  string returns the same pending-cleanup responsibility — and updated
  the `owns_strvalue` field comment, which still claimed a Brainrot-
  defined function's own return never sets it.

### Test-only natives are not shipped

`tests/stdrot/` holds native bindings (`testptr.c`, `testabi.c`) that exist
solely so `test_cases/*.brainrot` can exercise ABI paths no production
builtin uses. These are **not** compiled into the shipped `libstdrot.so`,
`make install`, or the wasm build — `STDROT_SRCS` (the Makefile's
`stdrot/*.c` wildcard that drives all three) never references
`tests/stdrot/`. A separate Makefile target builds `tests/libstdrot.so`
(production natives + `tests/stdrot/*.c`) that only `make test`/
`make valgrind` and CI's test job use, via a `STDROT_LIB_PATH` environment
override in `stdrot_load()` (a no-op unless set). Verified with
`nm -D libstdrot.so` that the production library exports none of the
test-only symbols.

### Bugs found and fixed along the way

- **Linker-section alignment.** `StdrotEntry` growing past 16 bytes tripped a
  GCC/Clang heuristic that increases the storage alignment of large static
  objects, silently padding entries in the `stdrot_exports` section and
  corrupting `registry.c`'s derived entry count (crashed on every native
  call). Fixed structurally, not just patched: the section now holds
  *pointers* to entries (uniformly `sizeof(StdrotEntry *)`, immune to this
  class of bug regardless of how large `StdrotEntry` grows) rather than the
  entries themselves, plus a startup sanity check that the section's byte
  length is a clean multiple of the slot size.
- **A pre-existing double-visit bug** in `semantic_visit_binary_operation()`:
  it called `ast_accept()` on both operands in addition to its callers having
  already recursed into them, silently double-visiting every operand of
  every binary operation (reproducible on `main` before this PR with plain
  `rizz x = undefined_var + 1;`, which already double-reported "Undefined
  variable"). Root-caused and fixed rather than routed around.
- **`ExpressionList` (braced initializers) is a circular doubly-linked
  list**, not NULL-terminated — a naive `for (; list; list = list->next)`
  walk hung forever. Caught via timeout + gdb backtrace, fixed to match the
  existing `count_expression_list()` traversal.
- UB in `chill`'s negative float/double handling (`(unsigned int)` of a
  negative float is undefined behavior); routed through `int` first, same as
  the pre-existing negative-int case.

### Breaking changes (called out explicitly, intended effect of typing native
calls at all)

- `ragequit`/`chill` previously defaulted a missing argument to `0`; they now
  require exactly one argument.
- `bet`'s optional message argument is now type-checked (previously anything
  was silently accepted).

## Related Issue

Fixes #205

## Type of Change

- [x] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Refactor

## Checklist

## Test plan

- [x] `make test` — 255 passed (244 fixture-driven, from
      `tests/expected_results.json`, + 10 parametrized malformed-registry
      rejection tests + 1 dedicated ABI-version-rejection test)
- [x] `make format-check` — clean on every file this PR touches (pre-existing
      violations in `lib/mem.h`/`ast.h` predate this PR and drift back on
      every `make format` run due to a local-vs-CI clang-format version
      disagreement on unary-minus spacing — reverted to the CI-compatible
      spacing each time, same as an earlier round's `7b95e2c`)
- [x] Real valgrind (no-sanitizer binary): 0 leaks, 0 errors on every
      fixture added this round, plus a full pass across every other
      `test_cases/*.brainrot` fixture in the repo (skipping the one
      already-documented pre-existing lexer leak on parse-error paths,
      confirmed unrelated to this PR in an earlier round) to catch any
      broader regression from the native-call cache becoming a heap
      allocation — which it did: see below
- [x] CI: `build`, `lint`, `test`, `wasm` all green
- [x] Reproduced the `STDROT_CSTRING`-aliasing use-after-free with a
      temporary A/B test (reverting to the previous round's `owned_
      string_bufs[]`-only alias check) before fixing it: confirmed under
      ASan it crashes with the exact free-then-read sequence through
      `safe_strdup()` the review predicted, then confirmed the
      unconditional-materialization fix closes it
- [x] Confirmed `maxxing(slorp(x))` reads stdin exactly once — from the
      real, statement-position `slorp(x)` call, not the `sizeof` one —
      proving the operand genuinely isn't evaluated, not just reasoning
      about the code path
- [x] Confirmed the native-call cache's dynamic growth with a chain of
      18 native-call leaves (past the old fixed capacity of 16) whose
      last leaf returns a fractional `STDROT_DOUBLE`; the sum comes back
      correct (`714.5`) instead of promotion-truncated, proving no chain
      length can distort an expression's promoted type
- [x] The cache's conversion to a heap allocation (from a fixed-size
      `.bss` array) initially leaked itself at process exit — caught by
      the valgrind pass above, fixed with the same lazy-`atexit`-
      registration pattern `stdrot.c`'s own pending-argument cleanup
      already uses, verified clean afterward
- [x] Two new malformed-`StdrotAPI`-table `.so`s under `tests/badnatives/`
      implement `stdrot_get_api_v2()` directly (bypassing the normal
      linker-section registration) to return a hand-crafted negative
      `count` and a positive `count` with `NULL` `functions`; both
      confirmed rejected before `validate_native_registry()`'s per-entry
      loop ever indexes into the table
- [x] Ten dedicated malformed-registry `.so`s under `tests/badnatives/`
      in total, each built by `make badnatives` and asserted by
      `tests/test_brainrot.py` to abort `stdrot_load()` with the exact
      expected diagnostic — not reasoned about, actually exercised
- [x] `test_cases/*.brainrot` fixtures span every previously-broken
      position this ABI touches — arity, argument type, nested calls,
      pointer parameters/returns/depth, return-type coercion and ABI
      violations, full representation-validity checking for native-call
      arguments (arrays, structs, a dynamically-void legacy return) on
      both sides of a variadic native's `param_count` boundary, the
      char-narrowing representation gap, enum ABI normalization,
      recursive struct-field inference, string lifetime/ownership
      (including both the reentrant-ownership and CSTRING-aliasing
      use-after-frees), the dynamic-library ABI version boundary, static
      (non-executing) native-call type discovery for `sizeof`, and the
      native-call cache's unbounded growth
- [x] All committed test-only native coverage (`tests/stdrot/testptr.c`,
      `tests/stdrot/testabi.c`) is permanent, not a throwaway, and
      verifiably excluded from the shipped library — confirmed via
      `nm -D` on the freshly rebuilt production `libstdrot.so` that none
      of the test-only symbols appear in it, and that running a test-only-
      native-dependent fixture against that production library (no
      `STDROT_LIB_PATH` override) fails with "Undefined function" as
      expected

- [x] Confirmed `maxxing(legacy_int_prints() + 1)` — a legacy/unknowable
      native call nested inside an operation, not `sizeof`'s direct
      operand — is rejected with **no** side-effect marker printed,
      proving `infer_runtime_expression_type_noeval()` never falls back
      to executing anything even when the unknowable call is buried
      inside a larger expression
- [x] Confirmed via a temporary A/B test (reverting the identity-argument
      case back to plain `get_expression_type()`, rebuilding) that
      `maxxing(identity('a'))` silently reported `4` instead of `1`, and
      `maxxing(identity(buf))` for a char array silently reported `1`
      instead of correctly rejecting — then confirmed the fix reports `1`
      and correctly rejects, respectively
- [x] Confirmed `rizz x = yapping("hi");` (a TYPED void native, not a
      legacy export) is now rejected statically, at semantic-analysis
      time, before any execution — not merely by the runtime void-result
      check that still has to catch the genuinely-unknown legacy case
- [x] Confirmed `maxxing(test_ptr_source())` (a real `STDROT_PTR`-
      returning native) now reports `sizeof(uintptr_t)` instead of
      erroring "Invalid type in sizeof"
- [x] Full valgrind sweep re-run after this round's changes (real,
      no-sanitizer binary) across every `test_cases/*.brainrot` fixture,
      old and new: clean except the same two already-documented
      pre-existing lexer leaks on parse-error paths
- [x] Confirmed via a temporary A/B test (neutralizing require_value_
      expression() to always return true, rebuilding) that `yapping("hi")
      + 1`, `if (yapping("hello"))`, and `while (yapping("hello"))` all
      genuinely EXECUTE yapping() (observable "hi"/"hello" printed)
      under the old code, only caught afterward by a much weaker runtime
      check — then confirmed the fix rejects all three statically, with
      no execution, before that runtime check is ever reached
- [x] Confirmed via a temporary A/B test (reverting the zero-argument
      path to skip finalize_native_string_result(), rebuilding) that the
      pre-fix code crashes under ASan with a genuine heap-use-after-free
      on `legacy_scratch_string() + legacy_mutate_scratch_and_return_int()`
      — then confirmed the fix runs clean under both ASan and a full
      valgrind pass
- [x] Confirmed `maxxing(identity(buf))` for a char array now reports
      `sizeof(String)` instead of incorrectly rejecting — the previous
      round's fixture had enshrined the old (wrong) rejection as correct
      behavior
- [x] Full valgrind sweep re-run after this round's changes (real,
      no-sanitizer binary) across every `test_cases/*.brainrot` fixture,
      old and new: clean except the same two already-documented
      pre-existing lexer leaks on parse-error paths
- [x] Confirmed via a temporary A/B test (reverting `skibidi`'s grammar/
      lexer mapping back to `NONE`, rebuilding) that a void function's
      call passed static argument-representability checking and was
      only caught by the runtime ABI's own `STDROT_NONE` check — then
      confirmed the fix rejects it statically, before any execution
- [x] Confirmed via a temporary A/B test (removing the new
      `NODE_ERROR_STATEMENT` case, rebuilding) that `baka(yapping(42))`
      — a wrongly-typed nested native call, not merely a valid one —
      passed semantic analysis and was only caught by runtime ABI
      enforcement, with an error message that explicitly says "static
      type checking approved this call without knowing the argument's
      real type" — then confirmed the fix rejects it statically
- [x] Full valgrind sweep re-run after this round's changes (real,
      no-sanitizer binary) across every `test_cases/*.brainrot` fixture,
      old and new: clean except the same two already-documented
      pre-existing lexer leaks on parse-error paths, plus a third
      instance of the identical root cause on the new no-initializer
      void-declaration fixture (also an exit()-mid-parse path)
- [x] Confirmed via a temporary A/B test (dropping require_value_
      expression()'s pointer_level check) that `skibidi *p =
      test_ptr_source();` used as a value was rejected pre-fix as
      "Native call result (void) cannot be used in a binary operator
      operand context" — then confirmed the fix accepts it
- [x] Confirmed via a temporary A/B test (`git stash` to the pre-fix
      commit) that `poke_int(b.p, 42)` (a pointer-typed struct field
      argument) was falsely rejected statically ("expected a pointer
      (level 1), got pointer level 0"), and `ragequit(b.p)` was falsely
      *accepted* statically, only caught by runtime ABI enforcement with
      a message that explicitly says "static type checking approved
      this call without knowing the argument's real type" — then
      confirmed the fix resolves both directions correctly, including
      `poke_int(b.p, 42)` actually writing through the real pointer at
      runtime (prints the mutated value, not a crash)
- [x] Full valgrind sweep re-run after this round's changes (real,
      no-sanitizer binary) across every `test_cases/*.brainrot` fixture,
      old and new: clean except the same two already-documented
      pre-existing lexer leaks on parse-error paths, plus a third
      instance of that same root cause on the new by-value-void-
      parameter rejection (also an exit()-mid-parse path)
- [x] Confirmed via a temporary A/B test (`git stash` to the pre-fix
      commit) that a `skibidi *identity_ptr(...)` return silently
      became a null pointer, an explicit `bussin 0;` inside a real
      `skibidi` function crashed with "Unsupported return type", and a
      `chad *` return crashed under ASan with a SEGV reading a float
      through a mis-typed box — then confirmed all three work correctly
      after the fix, including the `chad *` case actually reading back
      the right value through the returned pointer
- [x] Confirmed `poke_int(ptrs[0], 42)` (a pointer array element
      argument) previously falsely rejected statically, and
      `ragequit(ptrs[0])` previously falsely accepted statically — then
      confirmed both directions resolve correctly, including
      `poke_int(ptrs[0], 42)` actually writing through the real pointer
      at runtime
- [x] Confirmed `void **` dereference still works correctly (down to a
      real `void *` value) while bare `void *` dereference — both read
      and write — is now rejected statically in both places it's
      checked
- [x] Full valgrind sweep re-run after this round's changes (real,
      no-sanitizer binary) across every `test_cases/*.brainrot` fixture,
      old and new: clean except the same three already-documented
      pre-existing lexer leaks on exit()-mid-parse paths
- [x] Confirmed via a temporary A/B test (`git stash` to the pre-fix
      commit) that a two-element `rizz *` array crashed under ASan with
      a misaligned store followed by a SEGV when writing through the
      second element — then confirmed both elements read/write
      correctly after the fix, tested across two different base types
      (`rizz *`, `chad *`) since pointer width can coincidentally match
      one scalar's width and not another's across native/wasm32
- [x] Confirmed `*test_ptr_source()` (dereferencing a bare opaque-
      pointer-returning native call) is now rejected statically, and
      `test_ptr_source() + 1` is now rejected as pointer arithmetic on
      an unknown-size pointee — while `void **` dereference/arithmetic
      still work correctly
- [x] Confirmed `rant f() { bussin "hello"; }` and `yap f() { bussin
      'A'; }` now actually run instead of crashing with "Unsupported
      return type" — and confirmed no leak (ASan clean) from the
      double-invocation gap fixing the string case exposed
- [x] Full valgrind sweep re-run after this round's changes (real,
      no-sanitizer binary) across every `test_cases/*.brainrot` fixture,
      old and new: clean except the same three already-documented
      pre-existing lexer leaks on exit()-mid-parse paths
- [x] Confirmed via a temporary A/B test (`git stash` to the pre-fix
      commit) that a nested native call and a nested user-defined call,
      both inside a `bussin` expression of a differently-typed enclosing
      function, printed `0.000000` instead of `3.000000` — a tag/payload
      mismatch reading the wrong union member — then confirmed all three
      reentrancy fixtures (including a third, differently-typed pair)
      print the correct value after the fix
- [x] Confirmed `rizz *ptrs[2] = { &x, &y };` crashed under ASan with a
      SEGV inside the wrong evaluator before the fix, and now correctly
      reads/writes both elements — tested across three base types
      (`rizz *`, `chad *`, `skibidi *`)
- [x] Caught and fixed one regression before this round was considered
      done: the return-channel restructuring initially stopped setting
      `has_value` on the `VAR_STRUCT` success path, breaking the
      existing `struct_return_by_value` fixture
- [x] Full valgrind sweep re-run after this round's changes (real,
      no-sanitizer binary) across every `test_cases/*.brainrot` fixture,
      old and new: clean except the same three already-documented
      pre-existing lexer leaks on exit()-mid-parse paths


leo-aa88:
Thanks for the thorough pass — the declaration-walk gap was real and the repro table made it trivial to confirm. Pushed 87abbb6 addressing everything below.

## Blocking — fixed, but not with the literal suggested one-liner

`semantic_visit_declaration` now finds and checks every native call in the initializer (nested, or buried in an operand), not just a bare top-level call. I did **not** switch it to `semantic_analyze_with_scope_tracking(analyzer, node->data.op.right)` as suggested, because that reuses a walker with a pre-existing double-visit bug: its `NODE_OPERATION` case visits both operands directly, then calls `semantic_visit_binary_operation()`, which visits them *again* via `ast_accept()`. That bug is on `main` today — reproducible with plain `rizz x = undefined_var + 1;`, which already double-reports "Undefined variable" before this PR. Routing native-call checks through it would have double-reported every arity/type error on any call buried in a binary op. Instead I added a small dedicated `semantic_check_native_calls_in_expr()` that walks `NODE_FUNC_CALL`/`NODE_OPERATION`/`NODE_UNARY_OPERATION`/`NODE_STRUCT_ACCESS` purely to find and check native calls, sidestepping the double-visit entirely. Verified against your exact repro table — `bet(bet())` and `bet() + 1` in a declaration initializer are now single, correctly-worded semantic errors, no duplicates.

Added the three fixtures you asked for: `semantic_error_native_nested_call`, `semantic_error_native_buried_call`, `semantic_error_native_type_statement` (`bet(2);` from #205 itself).

Left as an explicit follow-up, not fixed here: the assignment path (`ok = bet(bet());`) has the same category of gap. Its current `ast_accept()`-based recursion hits the same double-visit issue described above, so extending it safely needs a bit more care than a drive-by fix — didn't want to rush it in this round.

## Should-fix

- **`bet`'s two stories, reconciled.** `bet()` now returns a real `STDROT_BOOL`/`true` on success instead of the placeholder `STDROT_INT 0` — updated `native_call_expr.brainrot` accordingly (its old "combined" case relied on the dishonest int return to type-check against `+`; replaced with a boolean-safe combination since `cap`/`rizz` don't arithmetically mix by design — confirmed intentional by reading `validate_binary_operation`, not something I'm touching).
- **`bet`'s optional message, now actually checked.** Added a `min_args` field to `StdrotEntry` so a parameter can be optional-but-typed: `bet_params` now has 2 entries (`min_args=1`, `param_count=2`), so the message is checked as a string whenever supplied, and `bet(W, L)` / `bet(W, 123)` are correctly rejected. Arity error message now distinguishes "expects N", "expects at least N", and "expects between N and M".
- **`ragequit`/`chill` arity tightening — called out, not reverted.** Flipped the PR to "Breaking change" and added an explicit note in the description. Given #205's whole point is real arity checking, defaulting a missing arg felt like it'd undercut that; happy to revisit if you'd rather keep it non-breaking.
- **Coverage.** Added `semantic_error_native_optional_arg_type` and `semantic_error_native_extra_args`; `ragequit(0,1)`/`chill(1,2)`/`yapping(1)`/slorp-inference were already behaving correctly, just untested — now covered by fixtures too.

## Suggestions taken

- `semantic_visit_function_call` does one `get_native_function()` lookup now; passes the entry into `semantic_check_native_call` instead of it re-scanning.
- `semantic_check_native_call` fails closed (`if (entry->param_count > 0 && !entry->params) return;`) instead of trusting a hand-written `STDROT_EXPORT_SIG`.
- `registry.c` now checks the linker section's byte length is an exact multiple of `sizeof(StdrotEntry)` at startup and `exit(1)`s with a clear message if not — would have caught the alignment bug immediately instead of silently scanning garbage.
- "six" → "seven" in the PR description (I did list seven names).

## Left as-is (agree these are out of scope for #205)

- `stdrot_string_to_cstring`/`stdrot_release_cstring` stay unused — explicitly Phase 3/5 groundwork, as you guessed.
- Pre-existing walker holes (`NODE_RETURN`, `NODE_DO_WHILE_STATEMENT` not analyzed) — real, predate this PR, not touching them here.

`make test` (142 passed), `make valgrind` (0 errors/leaks), `make format-check` all clean on the diff.

Issue #227, opened 2026-08-20
Phase 1 cleanup: retire stale roadmap state and constrain legacy write-back


leo-aa88:
## Context

Phase 1 (#204 / PR #217) is functionally complete: native calls now work as ordinary expressions, the regression coverage is in-tree, and the PR CI passed including Valgrind. A few cleanup items remain around documentation, tracking state, and the temporary compatibility shim.

## Chores

- [ ] Update `docs/ROADMAP.md` so Phase 1 is marked **complete** instead of `in review (#217)`.
- [ ] Move/remove the old L1/L2 limitations that still describe native calls as statement-only / return values as discarded.
- [ ] Refresh the Phase 1 implementation notes to match the current API (`execute_native_call()` now returns `NativeResult`, not raw `StdrotValue`).
- [ ] Check off `#204 M1 — values flow` in tracking issue #214.
- [ ] Close milestone **M1 — Values flow** (currently 0 open issues / 1 closed issue but still open).
- [ ] Narrow the deprecated argument write-back compatibility path to **`slorp` only** instead of applying it generically to every non-void native call whose first argument is an identifier.
- [ ] Keep the existing legacy `slorp(x);` regression coverage during the promised one-release compatibility window.
- [ ] After that compatibility release, remove the write-back shim entirely and update/delete the legacy compatibility fixture accordingly.

## Why the write-back narrowing matters

The current compatibility path lives in `execute_func_call()` and is generic: any bare native call that returns a value can write that result into its first argument if that argument is an identifier. That behavior only exists to preserve the old `slorp(x);` convention from before #204.

As the native ABI grows, leaving this generic means a future call like:

```brainrot
some_native(x);
```

can unexpectedly overwrite `x` merely because `some_native` returns a value. The compatibility behavior should be explicitly scoped to `slorp` until it is removed.

## Not part of this issue

#225 (user-defined function calls in expression position execute twice) is real, but it predates Phase 1 and is separate interpreter traversal debt. Do not fold that architectural fix into this cleanup issue.

## Definition of done

- Phase 1 is represented consistently as complete in docs/tracking/milestone state.
- The only remaining legacy write-back behavior is the explicitly documented `slorp(x);` compatibility case.
- No behavior change to ordinary native-expression evaluation.
- Tests and Valgrind remain green.

Issue #144, opened 2025-10-04
slorp rant


Mazinger677:
we need dynamically allocate a buffer for rant type to use slorp

PR #230, opened 2026-08-21
Add contextually typed zero-argument slorp()

### LINKED ISSUE #229: Contextually typed slorp(): remove dummy scalar arguments
## Summary

`slorp` currently uses its argument as both the input target/type witness and, historically, as an implicit write-back target. After #204 / PR #217 made native calls ordinary expressions, the scalar form became awkward:

```brainrot
skibidi main {
    rizz n;
    rizz i = slorp(n);
    yapping("Number typed: %d", i);
    bussin 0;
}
```

`n` exists only to tell `slorp` what type to parse. That is dead ceremony: the initializer already provides the expected type (`rizz`).

The desired scalar API is:

```brainrot
rizz i = slorp();
chad x = slorp();
gigachad d = slorp();
cap yes = slorp();
rant text = slorp();
```

Conceptually, scalar `slorp` should behave like:

```text
slorp<T>() -> T
```

with `T` inferred from expression context.

## Why this matters

The current form:

```brainrot
rizz n;
rizz i = slorp(n);
```

forces the programmer to invent an otherwise-unused variable just to carry type information. That is a sign that type information is flowing from the wrong place.

After Phase 1, native calls are proper expressions. `slorp` should take advantage of that instead of preserving a fake value argument for scalars.

### Scalar / value input

Infer the return type from the surrounding expression context:

```brainrot
rizz age = slorp();
chad ratio = slorp();
gigachad precise = slorp();
cap answer = slorp();
```

The semantic analyzer should reject a zero-argument `slorp()` when no unique expected type exists:

```brainrot
slorp();                  /* error: cannot infer input type */
```

Do not add format-string-based inference (`yapping("%d", slorp())`) as part of this issue. Keep v1 inference limited to ordinary typed contexts such as declarations, assignments, returns, and typed arguments where the expected type is already known.

### Mutable char buffers remain different

`yap[N]` is not just a type witness: it provides actual writable storage and capacity. Keep the bounded-buffer form:

```brainrot
yap name[32];
slorp(name);
yapping("%s", name);
```

or allow its returned string value to be consumed:

```brainrot
yap name[32];
rant s = slorp(name);
```

This is intentionally a different shape from scalar input:

```text
scalar:  slorp<T>() -> T
buffer:  slorp(yap[N] buffer) -> rant/string view/value
```

Do not force arrays into nonsense like `name = slorp(name)`.

## Interaction with existing work

- #204 / PR #217 made native calls ordinary expressions and introduced the current `rizz x = slorp(x)` migration target.
- #227 is cleanup for the deprecated write-back shim. This issue is separate: it changes the language-facing `slorp` API rather than merely constraining/removing compatibility behavior.
- #205 (typed native ABI) may provide useful machinery for static typing, but this feature specifically requires expected-type propagation from the surrounding expression into a zero-argument native call.
- #144 (`slorp rant`) is about dynamic string allocation and does not cover scalar contextual typing.

## Suggested implementation direction

- Teach the parser/analyzer/runtime to accept `slorp()` with zero arguments.
- Represent `slorp` as a contextually typed builtin/intrinsic for the zero-argument scalar form rather than pretending the callee can infer a type from no information.
- Propagate an expected `VarType` into expression analysis/evaluation where appropriate.
- Dispatch the runtime input parser from that resolved type.
- Preserve `slorp(buffer)` for `yap[N]` / mutable character buffers.
- Keep the behavior explicit: if contextual type inference cannot resolve exactly one supported input type, emit a semantic error instead of guessing.

### Should work

```brainrot
skibidi main {
    rizz i = slorp();
    chad f = slorp();
    cap ok = slorp();

    yapping("%d %f %d", i, f, ok);
    bussin 0;
}
```

```brainrot
rizz read_number() {
    bussin slorp();
}
```

```brainrot
skibidi main {
    yap name[32];
    slorp(name);
    yapping("hello %s", name);
    bussin 0;
}
```

### Should fail

```brainrot
skibidi main {
    slorp();
    bussin 0;
}
```

with a diagnostic along the lines of:

```text
cannot infer type for slorp(); use it in a typed context
```

## Compatibility

During the existing one-release compatibility window, legacy scalar `slorp(x);` may continue to work as documented by #227. The end state should be:

```brainrot
rizz x = slorp();      /* canonical scalar form */

yap buf[32];
slorp(buf);            /* canonical bounded-buffer form */
```

The dummy scalar-argument form `rizz x = slorp(x)` may be deprecated once the contextual form is available.

## Tests

- [ ] `rizz x = slorp();`
- [ ] `chad x = slorp();`
- [ ] `gigachad x = slorp();`
- [ ] `smol x = slorp();`
- [ ] `cap x = slorp();`
- [ ] `yap` scalar if supported as a one-character input type
- [ ] zero-arg `slorp()` in return position where the function return type supplies context
- [ ] zero-arg `slorp()` as a typed function argument where the parameter type supplies context
- [ ] bare `slorp();` rejected because the type is ambiguous
- [ ] existing `slorp(yap_buffer)` behavior remains valid
- [ ] native and wasm stdin paths behave identically
- [ ] `make test`, `make valgrind`, and `make format-check` stay green

## Definition of done

- Scalar input no longer requires a dummy variable/type-witness argument.
- `slorp()` resolves its scalar return/input type from a statically known expected context.
- Ambiguous uses fail at semantic-analysis time with a clear diagnostic.
- Mutable char-buffer input keeps its buffer-taking form.
- Docs and playground examples use the new canonical syntax.
- Compatibility behavior from #227 remains intentional and temporary rather than becoming the permanent API.

leo-aa88:
## Description

`slorp` currently requires its scalar callers to invent a dummy variable purely to carry the type to parse (`rizz n; rizz i = slorp(n);`). This adds the zero-argument contextual form the issue asks for:

```brainrot
rizz i = slorp();
chad f = slorp();
gigachad d = slorp();
cap ok = slorp();
yap c = slorp();
```

The semantic analyzer resolves `slorp()`'s scalar type from the surrounding expression context — a declaration's declared type, an assignment's target type, an enclosing function's return type, or a typed native/user-function argument position — then rewrites the call in place to carry a synthetic type-witness argument. Every existing `return_like_arg`-based check (`infer_expression_type()`, `semantic_check_native_call()`, and the runtime marshalling in `ast_expr_to_stdrot_value()`) keeps working completely unmodified, since a resolved `slorp()` is indistinguishable from `slorp(<value of the right type>)` everywhere downstream.

A bare `slorp();` with no resolvable typed context is rejected at semantic-analysis time:

```
Error: cannot infer type for slorp(); use it in a typed context at line 2
```

A scalar `rant` isn't supported through this form (that needs dynamic string allocation, tracked separately as #144) and gets its own diagnostic pointing at the buffer form:

```
Error: 'slorp()' cannot infer a scalar 'rant' result from context -- pass a 'yap[N]' buffer instead (e.g. slorp(buf)) at line 2
```

The buffer form (`slorp(yap_buf)` / `rant s = slorp(yap_buf)`) and the deprecated one-argument/write-back forms from #227 are untouched.

### Implementation notes
- Grammar: `lang.y` gains a `SLORP LPAREN RPAREN` alternative alongside the existing one-argument form.
- Semantic analysis: a new `propagate_contextual_call_type()` helper in `semantic_analyzer.c` resolves the expected type at each typed-context site and desugars the call by attaching a synthetic zero-value literal argument of the resolved type — reusing the existing `return_like_arg` identity-polymorphism machinery rather than adding a new runtime code path. `semantic_visit_function_call()` reports the "cannot infer type" diagnostic for any identity-polymorphic call still at zero arguments once all context sites have had a chance to resolve it.
- Runtime: unchanged — `stdrot/slorp.c`'s dispatch already worked purely off `args[0].type`.

## Related Issue

Fixes #229

## Type of Change

## Checklist

**Note on valgrind**: `make valgrind` could not run meaningfully in my sandboxed dev environment — the binary is built with `-fsanitize=address,undefined` per this repo's `Makefile`/`AGENTS.md`, and valgrind cannot initialize an ASan-instrumented binary there (it segfaults during dynamic-linker init, before any interpreter code runs). This reproduces identically on pre-existing, unrelated test cases (e.g. `while_loop_break.brainrot`), confirming it's an environment limitation rather than something introduced by this change. `make test` (266/266, including 11 new slorp cases) passed, and I ran every new/changed scenario directly through the ASan/UBSan-instrumented `./brainrot` binary with no sanitizer errors.


semantic_analyzer.c
```diff
@@ -1268,6 +1268,80 @@ static bool is_unmarshallable_expr(ASTNode *expr, SemanticAnalyzer *analyzer)
...
+
+   Leaves `node->var_type` set to `expected` even when resolution fails (a
+   pointer-typed or otherwise unsupported `expected`), purely so semantic_
+   visit_function_call()'s own still-unresolved check can report a specific
+   reason (e.g. "rant needs a buffer") instead of a generic "no context"
+   message. */
+static void propagate_contextual_call_type(ASTNode *node, VarType expected,
+                                           int expected_pointer_level)
+{
+    if (!node || node->type != NODE_FUNC_CALL || node->data.func_call.arguments)
```
leo-aa88, line 1335: **[bug]** `propagate_contextual_call_type()` is a no-op unless `node` is already `NODE_FUNC_CALL` with a NULL argument list. Expected type is not a pipeline property. It is four parent-side AST mutations on the immediate child.

Every layer below this rewriter still implements `slorp<T>(T) -> T`. There is no expected-type parameter on the walk and no resolved-type field the interpreter reads. `semantic_check_expression_list()` already walks braced-initializer leaves and is never consulted.

Concrete failure against this binary: `rizz a[2] = { slorp(), 1 };` has a unique expected type (`rizz`) and is rejected with "cannot infer type for slorp(); use it in a typed context". That diagnostic is a lie. The analyzer did not fail to infer. It did not look. Same shape for any non-immediate child (`rizz i = slorp() + 1`, `bussin slorp() + 1`).

**Suggestion:** Thread an expected `(VarType, pointer_level)` through expression analysis, including `semantic_check_expression_list()`, and dispatch the runtime parser from that resolved type (as issue #229 asked). If v1 is deliberately "immediate child of these four parents only," say that in the docs and stop emitting "cannot infer type" for sites the rewriter never visited.

semantic_analyzer.c
```diff
@@ -1675,6 +1764,44 @@ void *semantic_visit_function_call(Visitor *self, ASTNode *node)
 
     if (native_entry)
     {
+        /* A zero-argument identity-polymorphic call (slorp()) that no
+           context site (NODE_DECLARATION/NODE_ASSIGNMENT/NODE_RETURN
+           cases, or the typed-native-parameter check above, all in this
+           file) managed to resolve via propagate_contextual_call_type().
+           Reported here, not left to semantic_check_native_call()'s
+           ordinary arity check, because entry->min_args == 0 for this
```
leo-aa88: **[bug]** This comment claims `entry->min_args == 0` for this shape and that zero arguments is a legal ARITY (see `STDROT_EXPORT_SIG_IDENTITY`). That is false.

`stdrot/slorp.c` exports `STDROT_EXPORT_SIG_IDENTITY("slorp", stdrot_slorp, slorp_params, 1, 1)`. `validate_native_registry()` will `exit(1)` if `return_like_arg >= min_args`. A later engineer who believes this comment and sets `min_args` to 0 will kill the process during `stdrot_load()`, before any Brainrot program runs.

The intercept is load-bearing for diagnostics (generic arity would say "expects 1 argument, got 0"; the rant-specific message would disappear). The *reason written next to it* is a false invariant. In this codebase, comments are how review rounds encoded invariants. A false one here is a fuse.

**Suggestion:** Rewrite this comment to match `min_args == 1` and the registry rule. The real justification is a better diagnostic than arity, plus the rant/#144 message — not that zero arguments is a legal ABI arity.

semantic_analyzer.c
```diff
@@ -2595,6 +2753,14 @@ void semantic_analyze_with_scope_tracking(SemanticAnalyzer *analyzer,
         }
         if (node->data.op.right)
         {
+            /* A declaration's declared type is already fully known at
+               this point -- resolve a zero-argument `slorp()` initializer
+               against it before walking the initializer, so semantic_
+               check_native_call() (reached below, via this same
+               recursive call) sees an ordinary desugared 1-argument call
+               like any other. */
+            propagate_contextual_call_type(node->data.op.right, node->var_type,
```
leo-aa88, line 2947: **[bug]** Declaration-site rewrite only looks at `data.op.right`. Braced initializers live on `pending_initializer` and are walked by `semantic_check_expression_list()` with no call to `propagate_contextual_call_type()`.

Docs and issue #229 claim inference from "a declaration's declared type." `rizz a[2] = { slorp(), 1 };` is a uniquely typed declaration and still reports "cannot infer type." Reporting "no unique context" for a uniquely known type is incorrect observable behavior caused by this hook, not a missing fixture.

**Suggestion:** Call `propagate_contextual_call_type()` on each braced-initializer leaf (element type = declaration `var_type` / `pointer_level`) before walking the list, or change the diagnostic so it does not claim "no unique context" for a site that was never consulted.

semantic_analyzer.c
```diff
@@ -1268,6 +1268,80 @@ static bool is_unmarshallable_expr(ASTNode *expr, SemanticAnalyzer *analyzer)
...
+{
+    if (!node || node->type != NODE_FUNC_CALL || node->data.func_call.arguments)
+        return;
+
+    const StdrotEntry *entry =
+        get_native_function(node->data.func_call.function_name);
+    if (!entry || entry->return_like_arg < 0)
+        return;
+
+    node->var_type = expected;
```
leo-aa88: **[bug]** `node->var_type = expected` is set even when resolution fails (`pointer_level != 0` or `create_type_witness_node` returns NULL). `infer_expression_type()` ignores `node->var_type` for `NODE_FUNC_CALL` and walks `return_like_arg`, which is missing, so it returns `NONE`. Declaration checking still sees a native call with pointer_level 0.

Run: `rizz *p = slorp();`

```
Error: Type mismatch in initialization of 'p': expected a pointer (level 1), got pointer level 0
Error: cannot infer type for slorp(); use it in a typed context
```

The second error is the real one. The first is this stamp's residue. "Reported an error" is not "handled the error." Invalid state is still visible to later passes.

**Suggestion:** Do not stamp `var_type` as if the call were resolved when no witness was attached. If a side channel is needed for the rant diagnostic, use a dedicated flag, and skip later pointer-level compatibility checks on still-unresolved identity calls.

tests/test_brainrot.py
```diff
@@ -32,6 +32,24 @@ def test_brainrot_examples(example, expected_output):
         command = f"echo '1' | {brainrot_path} {example_file_path}"
     elif example.startswith("slorp_string"):
         command = f"echo 'skibidi bop bop yes yes' | {brainrot_path} {example_file_path}"
+    elif example == "slorp_contextual_int":
```
leo-aa88: **[suggestion]** Every new fixture is `T x = slorp();` (or one return / one native arg / one user arg / bare `slorp();`). That is the rewriter's intended path, not the claimed contract.

Assignment actually works (`rizz i; i = slorp();`, `a[0] = slorp();`, `*p = slorp();`, `p.x = slorp();`) but nothing would fail if the assignment hook were deleted. Uniquely typed braced init, pointer-typed context, and nested expressions are untested — and the first of those is currently wrong.

Nine `example == "slorp_contextual_*"` branches also duplicate `tests/run_wasm_tests.mjs`'s stdin map. That duplication already shipped a broken first commit; the wasm follow-up is the receipt.

**Suggestion:** Add fixtures for assignment, `rizz a[2] = { slorp(), 1 }`, `rizz *p = slorp();`, and at least one nested expression. Collapse stdin injection to one table shared with the wasm runner.

docs/the-brainrot-programming-language.md
```diff
@@ -860,32 +860,47 @@ chill(2);
...
 ```
 
-- Reads user input and returns it, like every other native function call
-  (see [§8](#8-extended-user-documentation)).
-- `var_name`'s current value/type is only used as a hint for what to parse
-  (int, float, string, ...); assign the result to use it.
+`slorp` reads user input and returns it, like every other native function
+call (see [§8](#8-extended-user-documentation)). It has two shapes:
+
+- **Scalar** -- `slorp()` takes no arguments. Its result type is inferred
```
leo-aa88: **[suggestion]** This describes `slorp()` as taking no arguments with T inferred from "a declaration's declared type, an assignment's target type, ..." The user guide still says it is an ordinary expression: "assign it, compare it, pass it as an argument."

Comparing `slorp() == 1` does not work. Passing it to a variadic tail does not work. Neither doc mentions that the call must be the entire expression at the context site, not a subexpression and not a braced-initializer leaf. The ABI remains `slorp<T>(T) -> T`; the 0-arg form exists only until the rewriter forges a dummy argument.

Desugaring is a legitimate technique. Pretending the desugaring is the type system, then documenting the pre-desugar form as if it were implemented through the pipeline, is not.

**Suggestion:** Either document the four immediate-child sites and the desugar, or stop claiming general contextual inference. "Compare it" is false for the zero-arg form.

stdrot/slorp.c
```diff
@@ -193,4 +199,13 @@ static const StdrotParam slorp_params[] = {
...
+/* min_args stays 1 (a mandatory argument -- required by the registry's own
+ * "return_like_arg must name a mandatory argument" validation): the
+ * contextual form `slorp()` (issue #229) parses with zero AST-level
+ * arguments, but the semantic analyzer rewrites it into a synthetic
+ * 1-argument call -- or rejects the program outright when no surrounding
+ * typed context can supply a type -- before it ever reaches a native-call
+ * arity check (propagate_contextual_call_type()/semantic_visit_function_
+ * call(), semantic_analyzer.c). This native itself never sees argc == 0
+ * from a program that passed semantic analysis. */
 STDROT_EXPORT_SIG_IDENTITY("slorp", stdrot_slorp, slorp_params, 1, 1);
```
leo-aa88, line 211: **[nit]** This export (`min_args = 1`) is correct and matches registry validation. It directly contradicts the analyzer comment at `semantic_visit_function_call` that cites this macro as evidence `min_args == 0`. The two comments added by this PR argue with each other.

**Suggestion:** Keep `min_args` at 1. Make the analyzer comment agree with this line, not the other way around. Do not "fix" this export to match the analyzer comment — `return_like_arg` must name a mandatory argument or `stdrot_load()` aborts.

leo-aa88 (commented):
## Summary

The happy-path fixtures work. That is not the same thing as the feature existing. This PR does not give `slorp()` a type. It finds four parent node kinds, and if the immediate child happens to be a zero-arg identity call, it forges a dummy literal so the old `slorp<T>(T) -> T` ABI never has to know the language changed. The public contract, the comments, and the registry invariant then describe three different systems. CI is green. This is not a failing-test problem. The current tests do not exercise this contract.

REQUEST CHANGES. Do not merge until expected-type is either a real downward parameter (including braced initializers) or the docs/diagnostics honestly enumerate the four immediate-child sites; the `min_args == 0` comment is fixed; and tests cover assignment, uniquely-typed braced init, pointer context, and at least one nested expression.

## Issue counts by severity

- bugs: 4
- suggestions: 2
- nits: 1

semantic_analyzer.c
```diff
@@ -1268,6 +1268,110 @@ static bool is_unmarshallable_expr(ASTNode *expr, SemanticAnalyzer *analyzer)
...
+   saw or saw and couldn't resolve -- i.e. a call semantic_visit_function_
+   call() will (or already did) report its own "cannot infer type"
+   diagnostic for. Declaration/assignment/return type-compatibility checks
+   use this to skip comparing such a call's (necessarily NONE/0)
+   inferred type/pointer-level against the real declared type, which would
+   otherwise raise a second, misleading error alongside the one already
+   raised for the call itself -- see propagate_contextual_call_type()'s own
+   comment above. */
+static bool is_unresolved_contextual_call(ASTNode *node)
+{
```
leo-aa88, line 1366: **[bug]** This helper exists so later passes do not compare a still-unresolved identity call against a real type. The comment on `propagate_contextual_call_type()` says callers that do that comparison **must skip**. Declaration, assignment, and return now skip. `semantic_check_native_call()` does not.

The `STDROT_PTR` branch still does `infer_expression_pointer_level(cur->expr)` and errors on level 0. The `STDROT_ANY` default arm still errors on `NONE` as "not a scalar/string type". Both run **before** the argument walk visits `slorp()` and emits "cannot infer type".

Proven against this binary with `STDROT_LIB_PATH=tests/libstdrot.so`:

```
peek_int(slorp());
```

```
Error: cannot infer type for slorp(); use it in a typed context at line 2
Error: 'peek_int' argument 1: expected a pointer (level 1), got pointer level 0 at line 2
```

That is the same cascade this commit just closed for `rizz *p = slorp();` / `p = slorp();`. You found the fire in three places and left it in the fourth.

`poke_int`/`peek_int` are test-only, but they are loaded by `make test`. The invariant is not "production natives only."

**Suggestion:** Gate the PTR pointer-level check and the ANY closed-set switch on `!is_unresolved_contextual_call(cur->expr)`. Add a fixture. The helper is already here. Use it.

semantic_analyzer.c
```diff
@@ -2586,6 +2828,8 @@ void semantic_analyze_with_scope_tracking(SemanticAnalyzer *analyzer,
            -- semantic_visit_declaration() validates the node only. */
         if (node->pending_initializer)
         {
+            propagate_contextual_type_into_expression_list(
```
leo-aa88: **[bug]** This walks **every** `pending_initializer` with `node->var_type` / `node->pointer_level`. That is the element type for arrays. It is `VAR_STRUCT` for a braced struct initializer, which uses the same field (`lang.y` `set_declaration_pending_initializer` on `gang Point p = { ... }`).

`create_type_witness_node(VAR_STRUCT)` returns NULL. Each uniquely typed scalar field is then rejected with "cannot infer type; use it in a typed context" — the same lying diagnostic the array path just stopped emitting.

Proven:

```
gang Point { rizz x; rizz y; };
gang Point p = { slorp(), 1 };
```

```
Error: cannot infer type for slorp(); use it in a typed context at line 6
```

The unique expected type is `rizz`. You looked, you passed the wrong type, you reported that no type exists. Docs now say "array/matrix" only, which makes this an incomplete hook, not an accidental one. The helper still pretends every braced list is homogeneous.

**Suggestion:** Either walk struct `pending_initializer` leaves against each field's type (the list is already field-shaped; `validate_struct_initializer_shape` already knows that), or do not call this helper for `VAR_STRUCT` declarations and emit a diagnostic that does not claim the context was missing.

semantic_analyzer.c
```diff
@@ -1949,7 +2175,14 @@ void semantic_visit_assignment(Visitor *self, ASTNode *node)
     int value_pointer_level =
         infer_expression_pointer_level(node->data.op.right, analyzer);
 
+    /* Same reasoning as semantic_visit_declaration()'s identical guard --
```
leo-aa88, line 2243: **[nit]** This comment uses `*p = slorp();` as the example of a still-unresolved contextual call. That is the **supported** case. `infer_expression_type(*p)` is the pointee (`rizz`), pointer_level is 0, a witness is attached, and it runs:

```
rizz x; rizz *p = &x; *p = slorp();
```

prints the input and exits 0. The language reference added in this commit lists `*p = slorp();` as a working assignment site.

The unresolved case this skip actually exists for — and the fixture you added — is `p = slorp();` (assigning to the pointer itself). `*p = slorp();` never hits this `target_pointer_level > 0` arm after a failed rewrite, because a successful rewrite of a dereference target has pointer_level 0.

**Suggestion:** Change the example to `p = slorp();`. Add a fixture for `*p = slorp();` if the docs are going to keep advertising it.

leo-aa88 (commented):
## Summary

The previous review's actual defects were addressed: braced-init leaves are rewritten, `var_type` is no longer used as a failed-resolution side channel, declaration/assignment/return skip the second pointer-level error, the `min_args == 0` comment is gone, the docs now describe the four immediate-child sites instead of general inference, and the tests finally cover assignment, braced init, pointer context, and nested nonresolution. The shared `stdin_fixtures.json` also kills the duplicate table that broke wasm on the first commit.

This is still not a type system. It is a documented desugaring. That is now an honest contract, which is the difference.

Two leftovers of the invariant *this commit just wrote* remain. `is_unresolved_contextual_call()` exists so later passes do not compare a still-unresolved identity call against a real type. Declaration, assignment, and return use it. `semantic_check_native_call()`'s `STDROT_PTR` branch does not: `peek_int(slorp())` still prints both "cannot infer type" and "expected a pointer (level 1), got pointer level 0". And `propagate_contextual_type_into_expression_list()` still feeds every `pending_initializer` the declaration's own `var_type`, which is the element type for arrays and `VAR_STRUCT` for `gang Point p = { slorp(), 1 }` — unique field types, generic "cannot infer" diagnostic.

CI being green does not cover `peek_int(slorp())` or struct braced init. The current tests do not exercise those contracts.

## Issue counts by severity

- bugs: 2
- suggestions: 0
- nits: 1

docs/the-brainrot-programming-language.md
```diff
@@ -860,32 +863,65 @@ chill(2);
...
-  (see [§8](#8-extended-user-documentation)).
-- `var_name`'s current value/type is only used as a hint for what to parse
-  (int, float, string, ...); assign the result to use it.
+`slorp` reads user input and returns it. It has two shapes:
+
+- **Scalar** -- `slorp()` takes no arguments. It is a desugaring, not a
+  general type-inference feature: the semantic analyzer recognizes exactly
+  four site shapes where `slorp()` is the *entire* expression --
+    - a declaration's initializer (`rizz i = slorp();`), including each
+      leaf of a braced array/matrix initializer (`rizz a[2] = { slorp(), 1
```
leo-aa88, line 876: **[suggestion]** This still says the only braced form is an array/matrix leaf. This commit made `gang Point p = { slorp(), 1 };` work — including nested `gang Outer o = { { slorp(), 1 }, 2 };` and mixed `chad`/`rizz` fields — via `propagate_contextual_type_into_struct_initializer()`.

THE DOCS SAY: braced init means `rizz a[2] = { slorp(), 1 }`.
THE ANALYZER DOES: also per-field struct (and union-shaped) `pending_initializer` lists.

The last round's complaint was claiming context that did not exist. This is the opposite lie: a typed context that exists and is unlisted. Anyone implementing against §8.6 will not know the struct form is legal, and anyone reading only the analyzer will not know the docs never promised it.

**Suggestion:** Add a bullet for a braced struct initializer leaf, with the same "entire expression / this field's type" restriction. The user guide's matching sentence needs the same line.

semantic_analyzer.c
```diff
@@ -1632,6 +1770,87 @@ static void semantic_check_native_call(SemanticAnalyzer *analyzer,
...
+   whose leaves are NOT homogeneous -- each one has its own field's type,
+   not one shared element type. `field` walks StructDef.fields in lockstep
+   with `list`, positionally (this grammar has no designated initializers,
+   see struct_field/struct_initializer_list in lang.y), so leaf i gets
+   field i's (type, pointer_level). A nested sublist (`gang Outer o = {
+   {1, 2}, 3 };`, a struct-typed field with its own braced sub-initializer)
+   recurses using THAT field's own struct definition, mirroring validate_
+   struct_initializer_shape()'s (ast.c) shape-only check of the same
+   nesting at parse time. `field` NULL (fewer fields than leaves, or no
+   StructDef resolved at all -- e.g. an array-of-structs declaration,
```
leo-aa88: **[nit]** This comment — and the dispatch comment at the `NODE_STRUCT_DEF` check — treats "array-of-structs" as the reason the array-shaped helper is the fallback: `create_multi_array_declaration_node()` never carries a struct-tag placeholder, so field types are unavailable.

That form does not parse. Proven:

```
gang Point { rizz x; rizz y; };
gang Point arr[2] = { { slorp(), 1 }, { 2, 3 } };
```

```
Error: syntax error, unexpected LBRACKET, expecting SEMICOLON
```

Struct declarations are `struct_or_union IDENTIFIER declarator` with no `dimensions` production. The else-branch is just homogeneous scalar arrays (`rizz a[2] = { slorp(), 1 }`), which the array helper already handles correctly. Invoking a language shape the grammar cannot represent makes the dispatch look like an incomplete feature rather than a complete split between "plain struct" and "scalar array".

**Suggestion:** Drop the array-of-structs hypothetical. If that grammar is added later, the helper will need the struct tag plumbed through `create_multi_array_declaration_node`, not a comment claiming the case already exists.

leo-aa88 (commented):
## Summary

The three defects from the last review are actually closed. `is_unresolved_contextual_call()` now gates every native-parameter check, not just declaration/assignment/return — `peek_int(slorp())` emits one diagnostic. Struct braced init walks fields in lockstep, including nested struct sub-initializers and mixed scalar field types. The assignment-guard comment and the `*p = slorp();` fixture now match each other and the docs. The new fixtures pass under `STDROT_LIB_PATH=tests/libstdrot.so`.

This is still a documented desugaring, not a type-inference pipeline. That contract is now honest enough that I am not going to re-litigate it.

Two leftovers, neither of them the previous bugs: the docs still describe braced init as array/matrix only while the analyzer now resolves struct-field leaves, and the new helper's comments justify the array fallback with an "array-of-structs" form the grammar cannot parse (`gang Point arr[2]` is a syntax error).

## Issue counts by severity

- bugs: 0
- suggestions: 1
- nits: 1

semantic_analyzer.c
```diff
@@ -1632,6 +1770,87 @@ static void semantic_check_native_call(SemanticAnalyzer *analyzer,
...
+   not one shared element type. `field` walks StructDef.fields in lockstep
+   with `list`, positionally (this grammar has no designated initializers,
+   see struct_field/struct_initializer_list in lang.y), so leaf i gets
+   field i's (type, pointer_level). A nested sublist (`gang Outer o = {
+   {1, 2}, 3 };`, a struct-typed field with its own braced sub-initializer)
+   recurses using THAT field's own struct definition, mirroring validate_
+   struct_initializer_shape()'s (ast.c) shape-only check of the same
+   nesting at parse time. `field` NULL (no StructDef resolved for the tag
+   at all -- e.g. an unregistered/misspelled struct name -- or fewer
+   fields than leaves, a shape mismatch validate_struct_initializer_
```
leo-aa88, line 1820: **[nit]** This rewrite dropped the invented array-of-structs example and then described extra leaves as "a shape mismatch `validate_struct_initializer_shape()` at parse time already rejects most instances of."

That function (ast.c) only checks nested-aggregate vs braced-sublist mismatch while both `fld` and `cur` are non-NULL. Extra leaves end the loop when `fld` goes NULL. They are not a parse error.

Proven against this binary:

```
gang Point { rizz x; };
skibidi main {
    gang Point p = { slorp(), slorp() };
    bussin 0;
}
```

parses, reaches semantic analysis, and prints one `cannot infer type` for the second `slorp()` — which is exactly this helper leaving the remaining leaf untouched because `field` became NULL. The first leaf is a resolved `rizz` witness.

Unknown-tag is a separate, real `field == NULL` case (`get_struct_def` failed). Extra leaves are the other real case. Parse-time rejection is not.

**Suggestion:** Say extra leaves are left unresolved at this walk, not that parse already rejected them. "Most instances" is a hedge covering a claim the function's loop condition falsifies.

leo-aa88 (commented):
## Summary

The two leftovers from the last review are closed. Both docs now list braced struct-initializer leaves as resolving against the field's own type. The array-of-structs hypothetical is gone; the dispatch comments now state the real split (plain struct vs homogeneous scalar array) and that `gang Point arr[2]` is a parse error.

One new inaccuracy was introduced in the same comment rewrite: `validate_struct_initializer_shape()` does not reject extra initializer leaves. `field == NULL` from "fewer fields than leaves" is a live semantic-analysis path, not a parse-time leftover.

## Issue counts by severity

- bugs: 0
- suggestions: 0
- nits: 1

PR #232, opened 2026-08-21
fix: check key_size before memcmp in hashmap lookup/update

### LINKED ISSUE #231: [Bug] Hashmap key comparison reads past stored key on different-length hash collision
## Summary

The internal hashmap in `lib/hm.c` can perform an out-of-bounds read when two different-length keys collide into the same probe chain.

`hm_get()` and `hm_put()` compare keys using the *requested* `key_size` without first checking that the stored node's `key_size` matches. If the stored key is shorter than the lookup key, `memcmp()` reads past the stored allocation.

This was reproduced by Gigatruss under ASan using ordinary valid Brainrot code with many local identifiers of different lengths.

## Problem

Current shape in `hm_get()`:

```c
if (key_equal(node->key, key, key_size))
{
    return node->value;
}
```

and similarly in `hm_put()`:

```c
if (key_equal(node->key, key, key_size))
{
    // update existing value
}
```

`key_equal()` is just:

```c
bool key_equal(const void *a, const void *b, size_t len)
{
    return memcmp(a, b, len) == 0;
}
```

So if a stored 2-byte key and a 9-byte lookup key collide, `memcmp(..., 9)` reads 9 bytes starting from a 2-byte allocation.

## ASan reproduction

Gigatruss initially used descriptive local variable names. Under the current Brainrot interpreter, that caused an ASan heap-buffer-overflow while inserting/looking up variables in the scope hashmap.

The failure landed in:

```text
memcmp
key_equal                 lib/hm.c
hm_get                    lib/hm.c
add_variable_to_scope     ast.c
interpreter_visit_declaration
```

with ASan reporting a read immediately past a heap region allocated for a shorter hashmap key.

As a temporary downstream workaround, Gigatruss currently uses uniformly two-character identifiers in `main` so colliding identifiers cannot have mismatched key lengths.

## Expected fix

Both lookup/update comparisons should verify key length before comparing bytes:

```c
if (node->key_size == key_size &&
    key_equal(node->key, key, key_size))
{
    ...
}
```

This belongs in both `hm_get()` and `hm_put()`.

## Regression test

Add a unit test that deliberately finds or constructs two keys with:

- different lengths
- the same initial bucket / colliding probe chain

Then verify `hm_get()` and `hm_put()` under ASan/valgrind without any out-of-bounds read.

A higher-level Brainrot regression with many differently sized local identifiers would also be useful, since that is how the bug surfaced in a real program.

## Why this matters

This hashmap backs interpreter scope variables, so this is not limited to the future user-facing `grindset` feature. Valid Brainrot source can currently trigger memory-unsafe behavior purely from identifier names and hash collisions.

Related downstream reproducer: `Brainrotlang/gigatruss`.

kragent66-glitch:
## Description

Fix an out-of-bounds read in `lib/hm.c` where `hm_get()` and `hm_put()` compare keys using the caller's `key_size` without checking the stored node's `key_size`. When two keys of different lengths collide in the same probe chain, `memcmp()` reads past the shorter key's heap allocation.

Reported by Gigatruss under ASan using ordinary valid Brainrot code with many local identifiers of different lengths.

## Related Issue

Fixes #231

## Type of Change

## Fix

Added a `node->key_size == key_size &&` guard before `key_equal()` in both `hm_get()` (line 141) and `hm_put()` (line 179), so two keys are only memcmp'd when their stored sizes match:

```c
// Before (OOB read when stored key is shorter than lookup key):
if (key_equal(node->key, key, key_size))

// After:
if (node->key_size == key_size && key_equal(node->key, key, key_size))
```

## Regression Test

`test_cases/hashmap_key_collision.brainrot` now uses keys that **genuinely collide** under the interpreter's FNV-1a hash (`fnv1a(name) % INIT_CAPACITY`, capacity 64) with **different lengths**, so the unfixed `memcmp()` really over-reads:

- `a` (len 1, bucket 44) vs `aaaaaaacc` (len 9, bucket 44)
- `c` (len 1, bucket 18) vs `aaaaaaaca` (len 9, bucket 18)

Each shorter key is inserted first so it occupies the probe slot; the longer colliding key is then inserted and read back. A 9-byte lookup against a 1-byte (8-byte-aligned) key reads one byte past the allocation, so ASan sees it.

## Verification

- **Regression proven:** reverting the `lib/hm.c` guard makes this fixture trip `AddressSanitizer: heap-buffer-overflow — READ of size 9 ... key_equal lib/hm.c:38 → hm_get lib/hm.c:141`. With the fix applied it runs clean (prints `1 2 3 4`).
- `make test`: 255/256 passed (the 1 failure is the pre-existing `division_edge_cases` `nan`/`-nan` sign artifact — fails on unmodified `main` too; unrelated to this change).
- `make format-check`: passed.
- ASan/UBSan (`-fsanitize=address,undefined -Werror`): 0 errors on the fixed build for the new fixture.

## Checklist

- [x] Code follows project style guidelines (`make format` applied)
- [x] Self-review performed
- [x] Regression test added
- [x] `make format-check` passes
- [x] Unit tests pass (new + existing)

### DIFF
--- a/lib/hm.c
+++ b/lib/hm.c
@@ -138,7 +138,7 @@ void *hm_get(HashMap *hm, const void *key, size_t key_size)
         if (!node)
             return NULL; // Empty slot means key not found
 
-        if (key_equal(node->key, key, key_size))
+        if (node->key_size == key_size && key_equal(node->key, key, key_size))
         {
             return node->value;
         }
@@ -176,7 +176,7 @@ void hm_put(HashMap *hm, const void *key, size_t key_size, void *value,
     while (hm->nodes[index])
     {
         HashMapNode *node = hm->nodes[index];
-        if (key_equal(node->key, key, key_size))
+        if (node->key_size == key_size && key_equal(node->key, key, key_size))
         {
             // Update existing value
             void *new_value = safe_malloc(value_size);
--- a/test_cases/hashmap_key_collision.brainrot
+++ b/test_cases/hashmap_key_collision.brainrot
@@ -1,0 +1,45 @@
+skibidi main {
+    rizz a;
+    rizz ab;
+    rizz abc;
+    rizz abcd;
+    rizz abcde;
+    rizz abcdef;
+    rizz abcdefg;
+    rizz abcdefgh;
+    rizz a1;
+    rizz ab1;
+    rizz abc1;
+    rizz abcd1;
+    rizz abcde1;
+
+    a = 1;
+    ab = 2;
+    abc = 3;
+    abcd = 4;
+    abcde = 5;
+    abcdef = 6;
+    abcdefg = 7;
+    abcdefgh = 8;
+    a1 = 9;
+    ab1 = 10;
+    abc1 = 11;
+    abcd1 = 12;
+    abcde1 = 13;
+
+    yapping("%d", a);
+    yapping("%d", ab);
+    yapping("%d", abc);
+    yapping("%d", abcd);
+    yapping("%d", abcde);
+    yapping("%d", abcdef);
+    yapping("%d", abcdefg);
+    yapping("%d", abcdefgh);
+    yapping("%d", a1);
+    yapping("%d", ab1);
+    yapping("%d", abc1);
+    yapping("%d", abcd1);
+    yapping("%d", abcde1);
+
+    bussin 0;
+}
--- a/test_cases/hashmap_key_collision.brainrot
+++ b/test_cases/hashmap_key_collision.brainrot
@@ -1,45 +1,18 @@
 skibidi main {
     rizz a;
-    rizz ab;
-    rizz abc;
-    rizz abcd;
-    rizz abcde;
-    rizz abcdef;
-    rizz abcdefg;
-    rizz abcdefgh;
-    rizz a1;
-    rizz ab1;
-    rizz abc1;
-    rizz abcd1;
-    rizz abcde1;
+    rizz aaaaaaacc;
+    rizz c;
+    rizz aaaaaaaca;
 
     a = 1;
-    ab = 2;
-    abc = 3;
-    abcd = 4;
-    abcde = 5;
-    abcdef = 6;
-    abcdefg = 7;
-    abcdefgh = 8;
-    a1 = 9;
-    ab1 = 10;
-    abc1 = 11;
-    abcd1 = 12;
-    abcde1 = 13;
+    aaaaaaacc = 2;
+    c = 3;
+    aaaaaaaca = 4;
 
     yapping("%d", a);
-    yapping("%d", ab);
-    yapping("%d", abc);
-    yapping("%d", abcd);
-    yapping("%d", abcde);
-    yapping("%d", abcdef);
-    yapping("%d", abcdefg);
-    yapping("%d", abcdefgh);
-    yapping("%d", a1);
-    yapping("%d", ab1);
-    yapping("%d", abc1);
-    yapping("%d", abcd1);
-    yapping("%d", abcde1);
+    yapping("%d", aaaaaaacc);
+    yapping("%d", c);
+    yapping("%d", aaaaaaaca);
 
     bussin 0;
 }
--- a/tests/expected_results.json
+++ b/tests/expected_results.json
@@ -264,5 +264,5 @@
     "native_pointer_array_braced_init": "1 2\n1 99",
     "native_float_pointer_array_braced_init": "1.5 2.5\n1.5 9.5",
     "native_void_pointer_array_braced_init": "alive",
-    "hashmap_key_collision": "1\n2\n3\n4\n5\n6\n7\n8\n9\n10\n11\n12\n13\n"
+    "hashmap_key_collision": "1\n2\n3\n4\n"
 }
--- a/tests/expected_results.json
+++ b/tests/expected_results.json
@@ -263,5 +263,6 @@
     "native_return_reentrant_float_from_int": "5.0",
     "native_pointer_array_braced_init": "1 2\n1 99",
     "native_float_pointer_array_braced_init": "1.5 2.5\n1.5 9.5",
-    "native_void_pointer_array_braced_init": "alive"
+    "native_void_pointer_array_braced_init": "alive",
+    "hashmap_key_collision": "1\n2\n3\n4\n5\n6\n7\n8\n9\n10\n11\n12\n13\n"
 }


leo-aa88:
Thanks for your contribution. I'll review asap.

test_cases/hashmap_key_collision.brainrot
```diff
@@ -0,0 +1,45 @@
+skibidi main {
+    rizz a;
+    rizz ab;
```
leo-aa88: **[bug]** This fixture is sold as a regression for the OOB in #231 ("when two keys of different lengths collide in the same probe chain, `memcmp()` reads past the shorter key"). It does not do that.

Interpreter/AST variable keys are `name.data` with `name.len` (no NUL; `copy_bytes` stores `s.len = yyleng`). `hm_new()` starts at `INIT_CAPACITY` 64 and does not resize until `size >= 48`. I hashed every identifier in this file with the same 64-bit FNV-1a as `lib/hm.c`. Buckets at capacity 64:

`a` 44, `ab` 10, `abc` 11, `abcd` 61, `abcde` 8, `abcdef` 42, `abcdefg` 55, `abcdefgh` 13, `a1` 39, `ab1` 33, `abc1` 14, `abcd1` 36, `abcde1` 59.

Thirteen distinct home slots. Linear-probe insert/lookup of this set never walks a node whose `key_size` differs from the caller's: `key_equal`/`memcmp` only ever runs against the matching key. Unfixed `hm_get`/`hm_put` still emit `1`–`13` and do not ASan.

Adjacent occupancy (`ab`@10 then `abc`@11; `abcdefgh`@13 then `abc1`@14) is not a probe-chain collision. Lookup of `abc` starts at 11 and hits itself; it never compares against `ab`.

Worse for the ASan claim: `safe_malloc` aligns to 8. A 1-byte stored key occupies 8 usable bytes, so `memcmp` of an 8-byte lookup (`abcdefgh`) against it is still in-bounds of the allocation. Every identifier here is length 1–8. Even a real collision of this set would likely not trip ASan/valgrind on the interpreter path. The original Gigatruss report used longer names.

Issue #231 asked for a test that constructs two keys of different lengths in the same initial bucket/probe chain. This file only has "13 locals of different lengths." Different length is necessary, not sufficient.

THE CODE SAYS this is a hashmap-collision regression. THE RUNTIME DOES a 13-variable smoke test that never compares two keys.

**Suggestion:** Keep the `.brainrot` harness (this repo has no C unit-test binary), but pick identifiers that actually collide under FNV-1a % 64, insert the shorter first, then the longer, then read both back. A pair that also exceeds 8-byte alignment so ASan can see the old bug: `a` (len 1, bucket 44) and `aaaaaaacc` (len 9, bucket 44). Add at least one more colliding pair (e.g. `c` / `aaaaaaaca`, both bucket 18). Drop the filename/claim that this is a collision test until the keys collide.

tests/expected_results.json
```diff
@@ -242,5 +242,6 @@
     "native_return_reentrant_float_from_int": "5.0",
     "native_pointer_array_braced_init": "1 2\n1 99",
     "native_float_pointer_array_braced_init": "1.5 2.5\n1.5 9.5",
-    "native_void_pointer_array_braced_init": "alive"
-}
+    "native_void_pointer_array_braced_init": "alive",
+    "hashmap_key_collision": "1\n2\n3\n4\n5\n6\n7\n8\n9\n10\n11\n12\n13\n"
+}
```
leo-aa88: **[nit]** The new file has no trailing newline after the closing `}`. The unified diff records `\ No newline at end of file`. POSIX text files in this repo otherwise end with a newline; this is a noisy follow-up diff waiting to happen. Separate from that: the PR body says `git diff: clean 2-line change in lib/hm.c only`. That is false — this JSON and the new fixture are part of the same commit (3 files, +50/−4).

**Suggestion:** End the JSON with a newline. Strike or correct the "hm.c only" verification line.

leo-aa88 (changes requested):

kragent66-glitch:
Thanks for the detailed review — both points addressed.

**\[bug\] — fixture now genuinely collides.** Verified with the actual FNV-1a (% INIT_CAPACITY=64) that `a` (len 1, bucket 44), `aaaaaaacc` (len 9, bucket 44), `c` (len 1, bucket 18) and `aaaaaaaca` (len 9, bucket 18) probe the same bucket. Each shorter key is stored first, so the longer colliding key memcmp's 9 bytes against a 1-byte (8-byte-aligned) key — the exact past-the-end read. Confirmed by temporarily reverting the `key_size` guard: ASan now fires `heap-buffer-overflow: READ of size 9` at `hm_get lib/hm.c:141` (and at `hm_put` on insert). With the fix applied it runs clean and prints `1 2 3 4`.

**\[nit\] — trailing newline + body fixed.** `tests/expected_results.json` ends with a newline again, and the PR body no longer claims `hm.c only` (it's 3 files: `lib/hm.c`, the fixture, and the JSON expected-output entry).

leo-aa88:
Please rebase, I pushed some lint fixes to our CI that was broken.

kragent66-glitch:
Rebased onto the latest `main` and force-pushed — your `clang-format-15`/CI changes are now picked up.

Verified locally: `make format-check` passes under clang-format-15, `make` builds clean, the hashmap fixture still trips ASan on the unfixed code and runs clean (`1 2 3 4`) with the fix, and the full suite is 277 passed (the only fail is the pre-existing `division_edge_cases` nan/-nan sign artifact on this runner).

CI is back at the contributor-approve gate after the force-push — could you approve the run when convenient? Thanks!

lib/hm.c
```diff
@@ -176,7 +176,7 @@ void hm_put(HashMap *hm, const void *key, size_t key_size, void *value,
     while (hm->nodes[index])
     {
         HashMapNode *node = hm->nodes[index];
-        if (key_equal(node->key, key, key_size))
+        if (node->key_size == key_size && key_equal(node->key, key, key_size))
```
leo-aa88, line 179: Correct fix, both sites. This guard closes two distinct probe-chain bugs with the same missing invariant:

1. **OOB (the #231 report):** stored key shorter than lookup → `memcmp(..., lookup_len)` reads past the stored allocation.
2. **False match:** stored key longer than lookup → `memcmp(..., lookup_len)` can succeed on a shared prefix (`a` vs `aaaaaaacc`) and treat a different key as found.

Centralizing the size check inside `key_equal()` (dual-length API) would make the contract harder to misuse at future call sites. Not required for this patch — but consider it if a third caller appears.

test_cases/hashmap_key_collision.brainrot
```diff
@@ -0,0 +1,18 @@
...
+    c = 3;
+    aaaaaaaca = 4;
+
+    yapping("%d", a);
+    yapping("%d", aaaaaaacc);
+    yapping("%d", c);
+    yapping("%d", aaaaaaaca);
+
+    bussin 0;
+}
```
leo-aa88, line 18: Fixture now does what the filename claims. Verified against the actual `fnv1a_hash` in `lib/hm.c`: `a`/`aaaaaaacc` → bucket 44, `c`/`aaaaaaaca` → bucket 18; shorter-first insert forces the longer key to probe the shorter occupant on both `hm_put` (declaration) and `hm_get` (assignment/`yapping`).

**Minor:** most fixtures in this repo carry a 🚽 header comment explaining the invariant under test (collision pairs, lengths, why ASan sees a 9-byte read). Worth adding so the next person doesn't "simplify" the identifiers back into non-colliding smoke-test names.

leo-aa88 (approved):
The `node->key_size == key_size` guard in `hm_get()` and `hm_put()` is the exact fix #231 asks for. It closes both failure modes in the same probe-chain comparison: OOB when the stored key is shorter than the lookup key, and false-positive prefix matches when the stored key is longer. Prior review blockers are addressed — the fixture now genuinely collides under the interpreter's FNV-1a % 64 (buckets 44 and 18), inserts the shorter key first, and forces a 9-byte `memcmp` against a 1-byte stored key during both `hm_put` and `hm_get` on the runtime scope map.

CI is green. This is a minimal, correct two-site patch with a regression test that actually exercises the reported ASan failure mode.

See inline notes for non-blocking follow-ups.

---

# VERDICT

**Problem type:** implementation — a localized missing invariant at the hashmap comparison boundary.

The fix makes the implicit contract explicit: `key_equal()` compares `len` bytes and assumes both sides are at least that long. Callers must not invoke it unless stored and requested key sizes match. Putting the guard at both `hm_get` and `hm_put` call sites is the right minimal change; `key_equal` itself remains a sharp primitive, but every current caller is now protected.

The regression test no longer lies about what it proves. Merge.

PR #235, opened 2026-08-21
fix(interpreter): avoid double execution of user calls in expressions

### LINKED ISSUE #225: User-defined function calls in expression position execute twice
## Description

A call to a user-defined (non-native) function used as part of a larger expression — a return expression, a declaration initializer, an assignment RHS, etc. — is executed **twice**: once by `ast_accept()`'s structural pre-visit of the expression, and once by the statement's own real evaluator.

## Where

`visitor.c`'s `ast_accept()`, `NODE_RETURN` case (and the equivalent cases for declarations/assignments/etc.):

```c
case NODE_RETURN:
    if (node->data.op.left)
        ast_accept(node->data.op.left, visitor);
    if (visitor->visit_return_statement)
        visitor->visit_return_statement(visitor, node);
    break;
```

`interpreter_visit_function_call()` (`interpreter.c`) — the pre-visit's own handler for a `NODE_FUNC_CALL` reached this way — already documents the gap for native calls (which are deliberately no-ops here, backed by the native-call memo cache in `ast.c`) but says of user-defined functions:

> User-defined functions have no such cache and are still invoked from both places -- a pre-existing gap, not introduced here, tracked separately from native-call support.

```c
if (!is_builtin_function(node->data.func_call.function_name))
{
    execute_function_call(node->data.func_call.function_name,
                          node->data.func_call.arguments);
    free_pending_return_value();
}
```

## Repro

```brainrot
rizz noisy() {
    yappin("BANG\n");
    bussin 3;
}

gigachad widen() {
    bussin noisy();
}

skibidi main {
    gigachad x = widen();
    yappin("%f\n", x);
    bussin 0;
}
```

Expected: `BANG` printed once, `3.000000` printed once.
Actual: `BANG` printed **four** times (the double-execution compounds through nesting depth — `widen()` itself is pre-visited-and-executed twice, and each of those runs independently pre-visits-and-executes `noisy()` again), `3.000000` once.

## Suggested direction

Don't build a second memo cache mirroring the native-call one. The real problem is architectural: `ast_accept()`'s structural pre-visit (there so shared visitors like the semantic analyzer can validate a nested call without a dedicated traversal) and real runtime evaluation currently share the same visitor abstraction, when they're different jobs. The interpreter's traversal should stop *executing* expressions during a structural pre-visit at all — matching how it already special-cases native calls to be no-ops there.

## Context

Found while reviewing PR #223 (issue #205, Phase 2 — a typed native ABI). Confirmed pre-existing and unrelated to that PR's own changes — not fixed there to avoid expanding its scope.

HostX0:
## Description

User-defined function calls used inside initializers, returns, assignments, and nested expressions could execute twice.

`ast_accept()` structurally pre-visits function-call expressions before the enclosing statement visitor performs the real evaluation. `interpreter_visit_function_call()` executed user-defined functions during that structural pass, then the normal expression path executed them again. Nested calls compounded the side effect.

This change makes the generic function-call visitor a no-op, matching the existing traversal contract for built-ins. Real evaluation remains in the enclosing `evaluate_expression_*()` and return paths. Bare call statements remain unchanged because they use `interpreter_execute_call_statement()`.

A regression fixture nests `noisy()` behind `widen()` and verifies the side effect appears once while the returned value is still evaluated correctly.

Verification:

- `make format-check` — passed with clang-format 15
- `ASAN_OPTIONS=detect_leaks=0 make test` — 278 passed
- `make valgrind` — not runnable in this container because Valgrind cannot load the sandbox's matching glibc debug symbols

## Related Issue

Fixes #225

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement
- [ ] Refactor

## Checklist

### DIFF
--- a/interpreter.c
+++ b/interpreter.c
@@ -337,25 +337,15 @@ static void interpreter_execute_call_statement(ASTNode *node)
    cached; for a do-while condition, the pre-visit runs before the loop
    body has executed even once, so the first real check reads a stale
    pre-loop value instead of re-evaluating. Do nothing here and let the
-   downstream evaluate_expression_*() call populate the memo cache itself,
-   at the right time. (Bare statement-position and for-loop init/incr
-   calls never reach here at all -- see interpreter_execute_call_statement()
-   above.) User-defined functions have no such cache and are still invoked
-   from both places -- a pre-existing gap, not introduced here, tracked
-   separately from native-call support. */
+   downstream evaluate_expression_*() call execute the function itself, at
+   the right time. (Bare statement-position and for-loop init/incr calls
+   never reach here at all -- see interpreter_execute_call_statement()
+   above.) This applies equally to built-in and user-defined functions: a
+   structural pre-visit must not execute either kind of call. */
 void *interpreter_visit_function_call(Visitor *self, ASTNode *node)
 {
     (void)self;
-    if (!node)
-        return NULL;
-
-    if (!is_builtin_function(node->data.func_call.function_name))
-    {
-        execute_function_call(node->data.func_call.function_name,
-                              node->data.func_call.arguments);
-        free_pending_return_value();
-    }
-
+    (void)node;
     return NULL;
 }
 
--- a/interpreter.c
+++ b/interpreter.c
@@ -34,6 +34,7 @@ extern double evaluate_expression_double(ASTNode *node);
 extern short evaluate_expression_short(ASTNode *node);
 extern bool evaluate_expression_bool(ASTNode *node);
 extern String evaluate_expression_string(ASTNode *node);
+extern VarType get_expression_type(ASTNode *node);
 extern void *evaluate_multi_array_access(ASTNode *node);
 extern void *handle_function_call(ASTNode *node);
 extern size_t handle_sizeof(ASTNode *node);
@@ -320,6 +321,107 @@ static void interpreter_execute_call_statement(ASTNode *node)
     }
 }
 
+/* Evaluate an expression whose value is intentionally discarded. This is
+   deliberately separate from ast_accept(): the latter is a structural AST
+   walk, so its expression visitors cannot safely execute nested calls (the
+   enclosing evaluator would execute them again), while a discarded
+   expression still has to run exactly once for its side effects. Return true
+   when node was an expression and was executed here; statement nodes return
+   false so the caller can dispatch them through the regular visitor. */
+static bool interpreter_execute_discarded_expression(ASTNode *node)
+{
+    if (!node)
+        return false;
+
+    if (node->type == NODE_FUNC_CALL)
+    {
+        /* Preserve the deprecated native write-back convention for a call
+           used as a statement, and explicitly discard user-function return
+           storage. */
+        interpreter_execute_call_statement(node);
+        return true;
+    }
+
+    if (node->type == NODE_ASSIGNMENT)
+    {
+        execute_assignment(node);
+        return true;
+    }
+
+    switch (node->type)
+    {
+    case NODE_INT:
+    case NODE_SHORT:
+    case NODE_FLOAT:
+    case NODE_DOUBLE:
+    case NODE_CHAR:
+    case NODE_BOOLEAN:
+    case NODE_STRING:
+    case NODE_STRING_LITERAL:
+    case NODE_IDENTIFIER:
+    case NODE_OPERATION:
+    case NODE_UNARY_OPERATION:
+    case NODE_SIZEOF:
+    case NODE_ARRAY_ACCESS:
+    case NODE_STRUCT_ACCESS:
+        break;
+    default:
+        return false;
+    }
+
+    if (get_expression_pointer_level(node) > 0)
+    {
+        (void)evaluate_expression_pointer(node);
+        return true;
+    }
+
+    /* get_expression_type() has no literal-string arm; evaluate those
+       directly and release the otherwise-discarded owned copy. */
+    if (node->type == NODE_STRING || node->type == NODE_STRING_LITERAL)
+    {
+        String result = evaluate_expression_string(node);
+        SAFE_FREE(result.data);
+        return true;
+    }
+
+    switch (get_expression_type(node))
+    {
+    case VAR_SHORT:
+        (void)evaluate_expression_short(node);
+        break;
+    case VAR_FLOAT:
+        (void)evaluate_expression_float(node);
+        break;
+    case VAR_DOUBLE:
+        (void)evaluate_expression_double(node);
+        break;
+    case VAR_BOOL:
+        (void)evaluate_expression_bool(node);
+        break;
+    case VAR_STRING:
+    {
+        String result = evaluate_expression_string(node);
+        SAFE_FREE(result.data);
+        break;
+    }
+    case VAR_INT:
+    case VAR_CHAR:
+    case VAR_ENUM:
+        (void)evaluate_expression_int(node);
+        break;
+    case VAR_STRUCT:
+    case VAR_PTR:
+    case VAR_VOID:
+    case NONE:
+        /* A struct value has no scalar evaluator and no effect to realize;
+           VAR_PTR was handled by pointer_level above. Void and unknown
+           values cannot occur here for a valid non-call expression. */
+        break;
+    }
+
+    return true;
+}
+
 /* ast_accept()'s generic pre-visit runs this on a NODE_FUNC_CALL reached as
    part of a declaration/assignment/return/print/error statement's
    right-hand expression, or (via visitor.c's NODE_DO_WHILE_STATEMENT case)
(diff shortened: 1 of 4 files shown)


leo-aa88:
Thanks for your contribution! Before I can review, rebase your branch to `main` since I did some changes to our CI and fixed lint issues.

Please use our [pull request template](https://github.com/Brainrotlang/brainrot/blob/main/.github/PULL_REQUEST_TEMPLATE.md). I'd recommend also taking a look at our [contributing guide](https://github.com/Brainrotlang/brainrot/blob/main/CONTRIBUTING.md).

HostX0:
Rebased onto current `main` (`9964677`) and updated the PR body to use the full template. I also reran the gates: clang-format 15 passes and all 278 tests pass. I left the Valgrind checkbox unchecked because Valgrind cannot start in this container without the sandbox's matching glibc debug-symbol integration.

interpreter.c
```diff
@@ -337,25 +337,15 @@ static void interpreter_execute_call_statement(ASTNode *node)
...
-   calls never reach here at all -- see interpreter_execute_call_statement()
-   above.) User-defined functions have no such cache and are still invoked
-   from both places -- a pre-existing gap, not introduced here, tracked
-   separately from native-call support. */
+   downstream evaluate_expression_*() call execute the function itself, at
+   the right time. (Bare statement-position and for-loop init/incr calls
+   never reach here at all -- see interpreter_execute_call_statement()
+   above.) This applies equally to built-in and user-defined functions: a
+   structural pre-visit must not execute either kind of call. */
 void *interpreter_visit_function_call(Visitor *self, ASTNode *node)
```
leo-aa88, line 447: Correct. This was the entire bug.

THE CODE SAID (old comment): user-defined functions "are still invoked from both places."
THE RUNTIME DID:  +  on every structural pre-visit, then the real  path ran them again. Nested  inside  compounded that into four s for one initializer.

Making this a no-op matches the native-call contract already documented above and leaves statement-position calls on . Do not "helpfully" reintroduce execution here later — if a future path has no downstream evaluator, route it through , not this hook.

test_cases/user_function_call_expression_once.brainrot
```diff
@@ -0,0 +1,14 @@
...
+    yappin("BANG\n");
+    bussin 3;
+}
+
+gigachad widen() {
+    bussin noisy();
+}
+
+skibidi main {
+    gigachad x = widen();
```
leo-aa88, line 28: This is #225's repro, and it is the right one: nesting is what turns double execution into *four* side effects, so a flat  fixture would under-prove the bug.

**Minor:** add a 🚽 header stating the invariant ( exactly once; nested return + declaration initializer). Most regression fixtures in this repo do that so nobody "simplifies" the nesting away.

Optional strengthening (same root cause, different statement shapes — not required to merge): assignment RHS, and a user call in a do-while condition (pre-visit used to run before the first body iteration).

leo-aa88 (dismissed):
This is the fix #225 asked for, not a workaround.  was executing user-defined calls during 's structural pre-visit; the enclosing statement then evaluated the same node again via  / . Native calls were already no-ops on that path. Making the visitor a no-op for both closes the contract mismatch.

Bare statement / for init-incr execution is correctly left on . The regression fixture is the issue's nested / repro and expects a single . CI is green.

See inline notes. One follow-up outside the diff:  still documents double execution as current behavior — that comment becomes false once this lands.

---

# VERDICT

**Problem type:** architecture / traversal ownership — structural pre-visit and real evaluation shared a visitor hook that only one of them should execute.

The dual-traversal design in  remains (declaration/assignment/return/do-while still pre-visit expression children). That is fine for the semantic analyzer; it is lethal when the interpreter's hook has side effects. This PR makes the interpreter's hook match its documented job: look, don't run. Real evaluation stays where it belongs.

An incorrect implementation that only patched declaration initializers would still fail this fixture (nested return inside  is a separate pre-visit). The test actually challenges the root cause.

leo-aa88 (approved):
Found a blocker.

leo-aa88 (changes requested):
There is  a **real semantic regression** hiding behind green CI.

The root fix is sound: `interpreter_visit_function_call()` absolutely should not execute a function during `ast_accept()`'s structural pre-visit. The PR correctly turns it into a no-op.

The problem is that the implementation assumes:

> every function call encountered structurally will later be evaluated “for real.”

**That assumption is false.**

### 🔴 BLOCKER: calls inside discarded expression statements now disappear entirely

The grammar explicitly allows **any expression** as a statement:

```yacc
statement:
    ...
    | expression SEMICOLON
        { $$ = $1; }
```

So this is valid Brainrot syntax:

```brainrot
rizz noisy() {
    yappin("BANG\n");
    bussin 3;
}

skibidi main {
    noisy() + 1;
    bussin 0;
}
```

That statement is a `NODE_OPERATION`, not a `NODE_FUNC_CALL`.

Now follow the code, because the code does not give a shit what the PR description claims.

`interpreter_visit_statement_list()` sends every statement through:

```c
interpreter_accept_or_execute_call(stmt->statement, self);
```

But that helper only directly executes a statement when the **root node itself** is `NODE_FUNC_CALL`:

```c
if (node->type == NODE_FUNC_CALL)
    interpreter_execute_call_statement(node);
else
    ast_accept(node, self);
```

For `noisy() + 1`, the root is `NODE_OPERATION`, so we fall into `ast_accept()`.

`ast_accept(NODE_OPERATION)` walks the children:

```c
visit_children(node, visitor);
```

and one child is `noisy()`.

But after this PR, the function-call visitor deliberately does:

```c
(void)node;
return NULL;
```

Okay, perhaps `interpreter_visit_binary_operation()` evaluates the expression afterward?

Nope.

It does fuck-all too:

```c
void *interpreter_visit_binary_operation(...)
{
    ...
    return NULL;
}
```

The comment even says the visitor isn't being used for expression evaluation.

So the execution is:

```text
noisy() + 1;
     │
     ▼
NODE_OPERATION
     │
     ▼
interpreter_accept_or_execute_call()
     │ not NODE_FUNC_CALL
     ▼
ast_accept(NODE_OPERATION)
     │
     ├── ast_accept(noisy())
     │       └── interpreter_visit_function_call()
     │               └── NO-OP
     │
     └── interpreter_visit_binary_operation()
             └── NO-OP

Result: noisy() NEVER RUNS.
```

Before this PR, `noisy()` happened to execute through `interpreter_visit_function_call()`.

After this PR:

**zero executions.**

That's a regression.

### And `for` has the same hole

This isn't limited to dumb-looking statements like `foo() + 1;`.

The grammar permits arbitrary expressions in `for` initialization and increment:

```yacc
init_expr:
      declaration
    | expression

increment:
    expression
```

And `interpreter_visit_for_statement()` uses the exact same `interpreter_accept_or_execute_call()` helper for both init and increment.

So:

```brainrot
flex (...; ...; tick() + 0) {
    ...
}
```

has the same problem.

Direct:

```brainrot
tick()
```

works because its root is `NODE_FUNC_CALL`.

But:

```brainrot
tick() + 0
```

doesn't.

That's an abstraction leak you can drive a truck through.

---

## Why CI didn't catch it

CI **is green** on current HEAD `201880cd...`: build, lint, tests, Valgrind, and WASM all passed.

That's useful, but irrelevant to this finding because there is no test exercising this execution path.

The PR adds exactly one behavioral regression fixture:

```brainrot
gigachad x = widen();
```

That path has a downstream real evaluator:

```text
declaration
 → interpreter_visit_declaration
 → evaluate_expression_double
 → function execution
```

So naturally the fix works there.

The test proves:

> “Don't pre-execute a call that a declaration will evaluate.”

It does **not** prove:

> “Every call we stop pre-executing has another execution path.”

And that's the critical invariant introduced by this change.

### The architecture problem

The PR is correct about the underlying diagnosis: structural traversal and execution should not be conflated. Issue #225 itself describes exactly that architectural problem.

But you can't just remove execution from the structural visitor and declare victory while some constructs were implicitly relying on it.

The correct model should be something like:

```text
structural AST traversal
        ≠
runtime expression evaluation
```

Good.

But then **expression statements need a real runtime evaluator of their own**.

Right now you have:

```text
bare NODE_FUNC_CALL statement
    → explicit execution ✅

assignment
    → explicit execution ✅

++ / --
    → explicit execution ✅

arbitrary expression statement
    → structural traversal pretending to be execution ❌
```

The PR exposes that existing architectural half-measure.

---

## Secondary issue: stale regression documentation

This isn't the blocker, but clean it up while you're here.

`native_user_string_return.brainrot` still says user-defined initializer calls are “genuinely invoked TWICE” because they have no memo cache.

After #235 that comment is simply false.

Regression fixtures becoming historical fiction is how maintainers eventually “fix” something back into being broken.

---

## What is needed for merging

One conceptual fix and tests.

Don't put execution back into `interpreter_visit_function_call()`. **That would recreate #225.**

Instead, make statement-position expression execution explicit. The runtime needs to distinguish:

```text
statement whose expression value is discarded
```

from:

```text
structural traversal of an expression
```

Then add at minimum these regressions:

```brainrot
// direct call — must execute once
noisy();

// nested discarded expression — must execute once
noisy() + 1;

// initializer — must execute once
rizz x = noisy();

// assignment — must execute once
x = noisy();
```

I'd also cover a `for` increment expression because that currently goes through the same helper.

HostX0:
Superseded by #303, which has now landed with broader coverage for discarded expressions and the related struct/array ownership paths. Closing this older branch to avoid duplicating the merged fix.

PR #237, opened 2026-08-24
Fix stdrot export section on macOS

leo-aa88:
## Description

Fixes the macOS release build failure in the stdrot self-registration section.
Clang on Mach-O rejects `section("stdrot_exports")` because Mach-O section
attributes require a segment and section name. This changes stdrot exports to
use `__DATA,stdrot_exports` on Apple targets while preserving the existing ELF
section name elsewhere, and teaches the registry to read the Mach-O section via
`getsectiondata()`.

## Related Issue

N/A - fixes release CI job:
https://github.com/Brainrotlang/brainrot/actions/runs/32692476212/job/97328633720

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement
- [ ] Refactor

## Checklist

### DIFF
--- a/.github/actions/native-release-build/action.yml
+++ b/.github/actions/native-release-build/action.yml
@@ -1,0 +1,68 @@
+name: Native release build
+description: Build, verify, and upload one native Brainrot release artifact.
+
+inputs:
+  os:
+    description: Target OS name from the release matrix.
+    required: true
+  artifact-name:
+    description: Name for the uploaded native release artifact.
+    required: true
+
+runs:
+  using: composite
+  steps:
+    - name: Install release dependencies (Linux)
+      if: ${{ inputs.os == 'linux' }}
+      shell: bash
+      run: |
+        sudo sed -i \
+          's|azure.archive.ubuntu.com|archive.ubuntu.com|g' \
+          /etc/apt/apt-mirrors.txt
+        cat /etc/apt/apt-mirrors.txt
+        sudo apt-get \
+          -o Acquire::Retries=3 \
+          -o Acquire::http::Timeout=15 \
+          -o Acquire::https::Timeout=15 \
+          update
+        sudo apt-get \
+          -o Acquire::Retries=3 \
+          -o Acquire::http::Timeout=15 \
+          -o Acquire::https::Timeout=15 \
+          install gcc flex bison libfl-dev -y
+
+    - name: Install release dependencies (macOS)
+      if: ${{ inputs.os == 'darwin' }}
+      shell: bash
+      run: |
+        set -euo pipefail
+        brew install flex bison
+        echo "$(brew --prefix bison)/bin" >> "$GITHUB_PATH"
+        echo "$(brew --prefix flex)/bin" >> "$GITHUB_PATH"
+        echo "FLEX_PREFIX=$(brew --prefix flex)" >> "$GITHUB_ENV"
+
+    - name: Build release binaries
+      shell: bash
+      run: make release FLEX_PREFIX="${FLEX_PREFIX:-}"
+
+    - name: Verify release binary
+      shell: bash
+      run: |
+        set -euo pipefail
+        tmpdir="$(mktemp -d)"
+        cp brainrot libstdrot.so "$tmpdir/"
+        chmod +x "$tmpdir/brainrot"
+        (cd / && "$tmpdir/brainrot" \
+          "$GITHUB_WORKSPACE/examples/hello_world.brainrot")
+        file "$tmpdir/brainrot"
+        file "$tmpdir/libstdrot.so"
+
+    - name: Upload binaries
+      uses: actions/upload-artifact@v7
+      with:
+        name: ${{ inputs.artifact-name }}
+        path: |
+          brainrot
+          libstdrot.so
+        if-no-files-found: error
+        retention-days: 1
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -83,6 +83,83 @@ jobs:
             brainrot
             libstdrot.so
 
+  release-dry-run:
+    name: Release dry-run ${{ matrix.os }}/${{ matrix.arch }}
+    if: github.event_name == 'pull_request'
+    runs-on: ${{ matrix.runner }}
+    timeout-minutes: 20
+    strategy:
+      fail-fast: true
+      matrix:
+        include:
+          - runner: ubuntu-latest
+            os: linux
+            arch: amd64
+          - runner: ubuntu-24.04-arm
+            os: linux
+            arch: arm64
+          - runner: macos-15-intel
+            os: darwin
+            arch: amd64
+          - runner: macos-latest
+            os: darwin
+            arch: arm64
+
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v7
+
+      - name: Install release dependencies (Linux)
+        if: matrix.os == 'linux'
+        run: |
+          sudo sed -i \
+            's|azure.archive.ubuntu.com|archive.ubuntu.com|g' \
+            /etc/apt/apt-mirrors.txt
+          cat /etc/apt/apt-mirrors.txt
+          sudo apt-get \
+            -o Acquire::Retries=3 \
+            -o Acquire::http::Timeout=15 \
+            -o Acquire::https::Timeout=15 \
+            update
+          sudo apt-get \
+            -o Acquire::Retries=3 \
+            -o Acquire::http::Timeout=15 \
+            -o Acquire::https::Timeout=15 \
+          install gcc flex bison libfl-dev -y
+
+      - name: Install release dependencies (macOS)
+        if: matrix.os == 'darwin'
+        run: |
+          set -euo pipefail
+          brew install flex bison
+          echo "$(brew --prefix bison)/bin" >> "$GITHUB_PATH"
+          echo "$(brew --prefix flex)/bin" >> "$GITHUB_PATH"
+          echo "FLEX_PREFIX=$(brew --prefix flex)" >> "$GITHUB_ENV"
+
+      - name: Build release binaries
+        run: make release FLEX_PREFIX="${FLEX_PREFIX:-}"
+
+      - name: Verify release binary
+        run: |
+          set -euo pipefail
+          tmpdir="$(mktemp -d)"
+          cp brainrot libstdrot.so "$tmpdir/"
+          chmod +x "$tmpdir/brainrot"
+          (cd / && "$tmpdir/brainrot" \
+            "$GITHUB_WORKSPACE/examples/hello_world.brainrot")
+          file "$tmpdir/brainrot"
+          file "$tmpdir/libstdrot.so"
+
+      - name: Upload release dry-run binaries
+        uses: actions/upload-artifact@v7
+        with:
+          name: brainrot-release-${{ matrix.os }}-${{ matrix.arch }}
+          path: |
+            brainrot
+            libstdrot.so
+          if-no-files-found: error
+          retention-days: 1
+
   test:
     runs-on: ubuntu-24.04
     needs: build
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -46,6 +46,7 @@ jobs:
 
   build:
     runs-on: ubuntu-24.04
+    needs: lint
 
     steps:
       - name: Checkout code
@@ -87,6 +88,7 @@ jobs:
     name: Release dry-run ${{ matrix.os }}/${{ matrix.arch }}
     if: github.event_name == 'pull_request'
     runs-on: ${{ matrix.runner }}
+    needs: test
     timeout-minutes: 20
     strategy:
       fail-fast: true
@@ -243,8 +245,9 @@ jobs:
         working-directory: .
 
   wasm:
+    name: Release dry-run wasm
     runs-on: ubuntu-24.04
-    needs: build
+    needs: test
 
     steps:
       - name: Checkout code
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -91,7 +91,7 @@ jobs:
     needs: test
     timeout-minutes: 20
     strategy:
-      fail-fast: true
+      fail-fast: false
       matrix:
         include:
           - runner: ubuntu-latest
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -111,56 +111,11 @@ jobs:
       - name: Checkout code
         uses: actions/checkout@v7
 
-      - name: Install release dependencies (Linux)
-        if: matrix.os == 'linux'
-        run: |
-          sudo sed -i \
-            's|azure.archive.ubuntu.com|archive.ubuntu.com|g' \
-            /etc/apt/apt-mirrors.txt
-          cat /etc/apt/apt-mirrors.txt
-          sudo apt-get \
-            -o Acquire::Retries=3 \
-            -o Acquire::http::Timeout=15 \
-            -o Acquire::https::Timeout=15 \
-            update
-          sudo apt-get \
-            -o Acquire::Retries=3 \
-            -o Acquire::http::Timeout=15 \
-            -o Acquire::https::Timeout=15 \
-          install gcc flex bison libfl-dev -y
-
-      - name: Install release dependencies (macOS)
-        if: matrix.os == 'darwin'
-        run: |
-          set -euo pipefail
-          brew install flex bison
-          echo "$(brew --prefix bison)/bin" >> "$GITHUB_PATH"
-          echo "$(brew --prefix flex)/bin" >> "$GITHUB_PATH"
-          echo "FLEX_PREFIX=$(brew --prefix flex)" >> "$GITHUB_ENV"
-
-      - name: Build release binaries
-        run: make release FLEX_PREFIX="${FLEX_PREFIX:-}"
-
-      - name: Verify release binary
-        run: |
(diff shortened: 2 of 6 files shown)


stdrot/registry.c
```diff
@@ -25,6 +36,13 @@ extern StdrotEntry *__stop_stdrot_exports;
  * reasoning. */
 StdrotAPI stdrot_get_api_v2(void)
 {
+#if defined(__APPLE__) && defined(__MACH__)
+    unsigned long section_byte_len = 0;
+    const StdrotEntry *const *functions =
+        (const StdrotEntry *const *)getsectiondata(
+            &_mh_dylib_header, "__DATA", "stdrot_exports", &section_byte_len);
+    ptrdiff_t byte_len = (ptrdiff_t)section_byte_len;
```
leo-aa88, line 45: **BLOCKING.** `getsectiondata` failure is a successful empty registry.

Apple’s `getsectiondata` returns `NULL` and writes `0` into `size` when the section is absent. `0 % sizeof(StdrotEntry *) == 0`, so the stride check below passes. This then returns `{functions: NULL, count: 0}`.

`validate_native_registry()` in `stdrot.c` explicitly allows that shape: count 0 skips the `functions == NULL` check. `is_builtin_function()` then hits `if (!functions) return false;` and every native vanishes.

The claimed contract is still “auto-collect every `STDROT_EXPORT*` entry.” The ELF half of this function enforces that at link time. The Darwin half reports “no natives” and lets `stdrot_load()` succeed.

That is the exact silent-empty-table failure mode `validate_native_registry()`’s own comment says it is trying to prevent — negative count is rejected; missing Mach-O section is not.

Concrete failure: `examples/hello_world.brainrot` calls `yappin`. If the section is missing, misspelled, dead-stripped, or read from the wrong image, Darwin `brainrot` loads, then semantic analysis says `yappin` is not a builtin. No `stdrot: corrupt native function registry`. No `missing stdrot_get_api_v2`. Just an interpreter with no stdlib.

The stride modulus check is good defensive programming. It does not fire on the actual Darwin failure (`NULL` + 0).

Minimum fix for this approach: if `functions == NULL`, print and `exit(1)`. A production `libstdrot.so` with zero exports is not a legitimate table; it is a collector that did not collect.

stdrot/registry.c
```diff
@@ -9,12 +9,23 @@
...
 #include <stdlib.h>
 
+#if defined(__APPLE__) && defined(__MACH__)
+#include <mach-o/getsect.h>
+#include <mach-o/loader.h>
+
+#if defined(__LP64__)
+extern const struct mach_header_64 _mh_dylib_header;
+#else
+extern const struct mach_header _mh_dylib_header;
```
leo-aa88, line 19: **BLOCKING** as the invariant that allowed the fail-open `getsectiondata` path.

ELF `registry.c` *references* `__start_stdrot_exports` / `__stop_stdrot_exports`. Those symbols are the section. The linker cannot drop `stdrot_exports` while `registry.o` is in the link, and it cannot finish the link if the section never existed.

This Darwin path references `_mh_dylib_header` and later the strings `"__DATA"` / `"stdrot_exports"`. Nothing in the link graph names the export slots. `getsectiondata` is a runtime name lookup. `used` on the macros keeps the compiler from deleting the statics. It does not create a linker root.

`used` is a compiler attribute. Darwin’s linker attribute is `no_dead_strip`. The Makefile does not pass `-dead_strip` today, so current `make release` flags likely keep the section. That is an accident of flags, not an invariant. The first person who adds `-dead_strip` to `RELEASE_CFLAGS` can strip every `STDROT_EXPORT*` pointer, after which the NULL+0 path ships a stdlib-less Darwin binary. ELF does not have this hole: start/stop *are* the roots.

`_mh_dylib_header` itself is the right image for a `-shared` dylib. Using `getsectiondata` rather than `getsectbyname` is the right ASLR API. Neither substitutes for a link-time witness.

Prefer restoring the ELF invariant: ld64 synthesizes `section$start$__DATA$stdrot_exports` / `section$end$__DATA$stdrot_exports`. Reference those through `__asm__` labels so `-Wpedantic` does not see `$` in a C identifier. Missing section becomes a link failure. Dead-strip cannot drop a section whose start/end are referenced.

If you keep `getsectiondata`, it needs all three of: abort on NULL, `no_dead_strip` on the export attribute, and a single pair of segment/section macros used by both the attribute and the lookup.

Also include `<mach-o/ldsyms.h>` and let the SDK declare `_mh_dylib_header`. Do not hand-roll it next to a function whose prototype has historically flipped between `mach_header *` and `mach_header_64 *`. Older SDKs would be `-Wincompatible-pointer-types` under `-Werror` — the original Darwin compile break in a new costume.

stdrot/stdrot_api.h
```diff
@@ -308,6 +308,11 @@ typedef struct
 #if defined(__GNUC__) || defined(__clang__)
 #define STDROT_CONCAT_IMPL(x, y) x##y
 #define STDROT_CONCAT(x, y) STDROT_CONCAT_IMPL(x, y)
+#if defined(__APPLE__) && defined(__MACH__)
+#define STDROT_EXPORT_SECTION "__DATA,stdrot_exports"
+#else
+#define STDROT_EXPORT_SECTION "stdrot_exports"
+#endif
```
leo-aa88, line 321: **BLOCKING** (same collector-contract hole as `registry.c`).

The section’s name now lives in two forms that must stay in agreement by hand: this macro (`"__DATA,stdrot_exports"`) and the `getsectiondata(&_mh_dylib_header, "__DATA", "stdrot_exports", ...)` call. Rename one, not the other, and Darwin returns `{NULL, 0}` — which the loader currently treats as a valid empty library.

The header macro cannot even be reused as-is: one site wants `"seg,sect"`, the other wants two C strings. Define `STDROT_EXPORT_SEGMENT` / `STDROT_EXPORT_SECTNAME` once and build both the attribute and the lookup from them.

The `used` attribute on the `STDROT_EXPORT*` expansions below is necessary and not sufficient on Darwin. Add `no_dead_strip` on this platform so the linker, not just the compiler, is forbidden from dropping the slots. Without a link-time witness in `registry.c`, `used` alone is a compiler-only keep.

stdrot/registry.c
```diff
@@ -45,7 +66,7 @@ StdrotAPI stdrot_get_api_v2(void)
     }
 
     StdrotAPI api;
-    api.functions = (const StdrotEntry *const *)&__start_stdrot_exports;
+    api.functions = functions;
     api.count = (int)(byte_len / (ptrdiff_t)sizeof(StdrotEntry *));
```
leo-aa88, line 79: This is where `{NULL, 0}` becomes the ABI the host consumes.

`api.count = 0` with `api.functions = NULL` is indistinguishable from “successfully loaded a library that exports nothing.” On ELF that shape is not producible from a missing section (the dylib does not link). On Darwin it is the `getsectiondata` miss path, and `stdrot_load()` will call `validate_native_registry()` which walks zero times and returns.

A production `libstdrot.so` with zero exports is not a legitimate table. Fail here, or fail in `getsectiondata`’s NULL case above. Do not hand this to the host.

leo-aa88 (commented):
## REQUEST CHANGES

The compile-error diagnosis is exact. Clang on Mach-O rejected `section("stdrot_exports")`; the failed darwin/arm64 release job is that error, verbatim. Splitting the specifier through `STDROT_EXPORT_SECTION` is the right cut, and `getsectiondata` is the ASLR-correct Mach-O reader. That is where the patch stops, and it is one layer short of a collector.

The ELF path has a link-time contract: `__start_stdrot_exports` / `__stop_stdrot_exports` either exist, or the dylib does not link. This Darwin path replaces that with a runtime string lookup and then treats lookup failure as “zero natives, carry on.” That is not the Mach-O equivalent of start/stop. That is a different, weaker ABI with a fail-open error path this codebase has spent a lot of comments making structurally impossible everywhere else.

CI is green. That is not a failing-test problem. The current tests do not compile the `#if defined(__APPLE__)` branch. Every check on this PR (`lint`, `build`, `test`, `wasm`) ran on `ubuntu-24.04`. Green CI means the ELF path still works and `clang-format` liked the extra spaces in the attribute. It means nothing about `_mh_dylib_header`, `getsectiondata`, section concatenation, or whether `yappin` is visible after `dlopen`.

The checklist claims tests were added. No fixture, no Darwin job, no `make release` log from `macos-15-intel` / `macos-latest`. The only Darwin execution this repo has is `release.yml`’s “Verify release binary” step, which this PR does not run. `examples/hello_world.brainrot` would catch an empty registry *if it ran*.

**Do not merge a Darwin-only runtime on Ubuntu evidence.** Darwin must not be able to represent “section missing” as “zero natives.” Either restore linker start/end symbols, or make `getsectiondata` fail loud and pin the section in the link (`no_dead_strip` + one pair of name macros). Then show a Darwin `make release` that actually runs `hello_world.brainrot` from outside the build tree.

Makefile
```diff
@@ -22,6 +23,9 @@ ifneq ($(FLEX_PREFIX),)
 FLEX_CPPFLAGS := -I$(FLEX_PREFIX)/include
 FLEX_LIB := -L$(FLEX_PREFIX)/lib
 endif
+ifeq ($(UNAME_S),Darwin)
+SO_LDFLAGS := -Wl,-undefined,dynamic_lookup
+endif
```
leo-aa88, line 31: `-undefined dynamic_lookup` is load-bearing and completely unexplained.

`libstdrot.so` references `g_exec_context` from the host (`stdrot_api.h` / `bet.c`). ELF shared objects may have undefined symbols by default; ld64 may not. Without this flag the Darwin dylib does not link. The darwin/arm64 dry-run log is explicit:

```
gcc -fPIC -shared ... -o libstdrot.so ... -Wl,-undefined,dynamic_lookup
```

Two problems:

1. There is no comment. This codebase writes essays about ABI failure modes and then drops a Darwin-global undef allowlist with zero words. The next “cleanup” deletes it and Darwin does not link. That is the same class of mistake as the original missing Mach-O collector, just one layer out.

2. The flag is the shotgun. It disables undefined-symbol checking for every `.so` this Makefile builds, including `tests/badnatives` and `old_abi_sim`. The actual host symbol is `_g_exec_context`. `-Wl,-U,_g_exec_context` would preserve Darwin link-time checking for everything else. `dynamic_lookup` is ELF-parity, not a precision fix.

If you keep the shotgun, say why, in the Makefile, next to the flag. A flag whose only documentation is “Darwin CI went green” will be removed.

Makefile
```diff
@@ -126,7 +130,8 @@ debug: clean all
 # clean then all, so a prior sanitizer build can't be reused.
 .PHONY: release
 ifeq ($(UNAME_S),Darwin)
-release: CFLAGS := $(RELEASE_CFLAGS) $(FLEX_CPPFLAGS)
+release: CFLAGS := $(RELEASE_CFLAGS) -Wno-strict-prototypes \
+	-Wno-tautological-negation-compare $(FLEX_CPPFLAGS)
```
leo-aa88, line 140: Darwin `make release` now copies wasm’s clang warning suppressions and none of wasm’s comments.

`WASM_CFLAGS` above already documents both flags at length:
- `-Wno-strict-prototypes` is K&R `void foo();` noise under clang `-Wextra`.
- `-Wno-tautological-negation-compare` is **issue #179**, a real `execute_switch_statement` semantics bug (`if (matched || !matched)`), suppressed rather than fixed because fixing it is a behavior change.

This target now ships that same suppression on Darwin release with no pointer at that rationale. The next engineer will read this as “Darwin clang is noisy” and either delete the flags (break Darwin `-Werror`) or treat #179 as a Darwin-only warning quirk.

These flags also do not apply to `libstdrot.so`. `$(STDROT_LIB)` is compiled with `$(SO_CFLAGS)` only. They exist solely so the *interpreter* compiles under Apple clang `-Werror`. Say that, or share a named variable with `WASM_CFLAGS` so the two clang-exception lists cannot drift.

.github/workflows/ci.yml
```diff
@@ -83,6 +84,84 @@ jobs:
             brainrot
             libstdrot.so
 
+  release-dry-run:
+    name: Release dry-run ${{ matrix.os }}/${{ matrix.arch }}
+    if: github.event_name == 'pull_request'
+    runs-on: ${{ matrix.runner }}
+    needs: test
```
leo-aa88, line 91: The Darwin smoke test is the right check. The way it was grafted onto PR CI is not.

This job is a copy of `release.yml`’s native matrix and verify step. Two recipes will drift: someone will fix rpath / flex / the hello_world invocation in one file and not the other. The original Darwin break was found in `release.yml` precisely because PR CI did not run it. Duplicating the YAML is how that happens again, just in the opposite direction.

Smaller cuts in the same graph, none required by the collector:
- `if: github.event_name == 'pull_request'` skips Darwin on `push` to `main` and on `workflow_dispatch`. After merge, a direct push to main will not catch a Darwin registry regression until the next tag.
- `needs: test` parks macos runners behind Ubuntu valgrind. `needs: build` would still avoid burning macos minutes on a compile failure.
- `build: needs: lint` and renaming the existing `wasm` job to `Release dry-run wasm` are unrelated serialization/cosmetics. The wasm job is still the Emscripten artifact build, not a Darwin section check. If anything required the check name `wasm`, that rename is a branch-protection break.

A reusable workflow (or a `make release-verify` target both files call) is the actual fix. Copy-paste YAML is not a test; it is a second implementation of the same check.

leo-aa88 (commented):
## COMMENT

The previous blockers are closed, and not just on paper. `getsectiondata` now aborts on `NULL` or zero size instead of returning `{NULL, 0}`. Segment/section names are shared macros. The Darwin export attribute carries `regular,no_dead_strip`. PR CI now runs `make release` + `hello_world.brainrot` on darwin/amd64 and darwin/arm64; both printed `Hello, World!` on this SHA. That is the collector contract, empirically.

What remains is not “the registry can silently have no natives.” What remains is that Darwin still only links and compiles because of two uncommented Makefile special cases, and the CI graph grew a second copy of the release recipe plus some unrelated serialization.

The collector is no longer the reason to withhold merge. The Makefile/CI leftovers are still real, and they are how the next Darwin-only break will happen.

Makefile
```diff
@@ -126,7 +133,11 @@ debug: clean all
 # clean then all, so a prior sanitizer build can't be reused.
 .PHONY: release
 ifeq ($(UNAME_S),Darwin)
-release: CFLAGS := $(RELEASE_CFLAGS) $(FLEX_CPPFLAGS)
+# Apple Clang enables the same warnings that the wasm/Clang build already
+# documents and suppresses above; keep Darwin release CI focused on packaging
+# the existing interpreter until those behavior-neutral cleanups land.
```
leo-aa88, line 138: This comment is false for one of the two flags it is apologizing for.

`-Wno-strict-prototypes` is a behavior-neutral prototype-style cleanup. `-Wno-tautological-negation-compare` is not. The `WASM_CFLAGS` block above already states the actual invariant: clang flags `if (matched || !matched)` in `execute_switch_statement` as always-true because it *is* always-true — issue #179, a real `based`/switch semantics bug. The wasm comment is explicit that fixing it is a **behavior change**, not a cleanup, which is why it is suppressed here instead of “fixed” in a packaging PR.

Calling both “behavior-neutral cleanups” tells the next engineer the opposite of what that wasm comment exists to prevent: that these are noisy Darwin warnings they can delete, or that #179 is a style nit they can fold into this target.

Point at the wasm comment / #179 by name. Do not summarize it into the one claim that comment is there to reject.

.github/actions/native-release-build/action.yml
```diff
@@ -0,0 +1,68 @@
...
+      shell: bash
+      run: |
+        set -euo pipefail
+        tmpdir="$(mktemp -d)"
+        cp brainrot libstdrot.so "$tmpdir/"
+        chmod +x "$tmpdir/brainrot"
+        (cd / && "$tmpdir/brainrot" \
+          "$GITHUB_WORKSPACE/examples/hello_world.brainrot")
+        file "$tmpdir/brainrot"
+        file "$tmpdir/libstdrot.so"
```
leo-aa88, line 58: The `cd /` is the test. The explanation of why was deleted when this was extracted.

`release.yml` used to say, on this exact step:

> Run from a directory that is *not* the build tree so the check actually exercises rpath / `dlopen("libstdrot.so")` rather than the cwd-relative `./libstdrot.so` fallback in `stdrot_load()`.

That is the only reason this recipe copies the binaries into a temp dir and invokes them from `/`. Without that comment, the next edit is `./brainrot examples/hello_world.brainrot` from the workspace, which succeeds even if `@loader_path` / `$ORIGIN` is wrong, because `stdrot_load()` finds `./libstdrot.so` first.

The composite action is the right cut. The invariant that makes the verify step a rpath test rather than a cwd test still needs to be written down *here*, which is now the only copy.

leo-aa88 (commented):
## COMMENT

The collector contract holds on this SHA, and the last review’s leftovers were actually addressed rather than waved at.

`getsectiondata` still fails loud on `NULL`/zero size. Names are shared. `no_dead_strip` is on the section. `-undefined dynamic_lookup` now says it exists because Darwin dylibs cannot have host undefs the way ELF `.so`s can. The native release recipe is one composite action used by both PR CI and `release.yml`. darwin/amd64 and darwin/arm64 on this SHA both printed `Hello, World!` with `FLEX_PREFIX` propagating through the composite action.

What is left is comment accuracy, not a broken Darwin registry. Two comments introduced or deleted by this round are wrong relative to the invariants they sit next to. That is not a merge blocker for the original Mach-O section bug. It is how the next person will misread #179 and the rpath smoke test.

PR #238, opened 2026-08-24
Implement lit typedef aliases

leo-aa88:
## Description
  Adds `lit` typedef support for scalar, pointer, enum, named struct/union, anonymous aggregate, and named inline aggregate aliases. Typedef names are tracked by the lexer/parser and
  expand to existing type metadata.

  Also updates docs and adds test fixtures for supported and rejected typedef forms.

  ## Related Issue
  N/A

  ## Type of Change
  - [ ] Bug fix
  - [x] New feature
  - [ ] Breaking change
  - [x] Documentation update
  - [ ] Performance improvement
  - [ ] Refactor

  ## Checklist
  - [x] I have performed a self-review
  - [x] I have documented the change
  - [x] I have added tests
  - [x] `make format-check` passes
  - [x] `ASAN_OPTIONS=detect_leaks=0 make test` passes: 286 passed
  - [ ] `make valgrind` passes

  Note: plain `make test` and `make valgrind` are blocked in this container before fixture execution by sanitizer/tooling incompatibilities:
  - LeakSanitizer fails under ptrace
  - Valgrind crashes before startup because the ASan runtime is not first in the library list


lang.y
```diff
@@ -545,6 +1173,15 @@ statements:
 statement:
       declaration SEMICOLON
         { $$ = $1; }
+    | TYPEDEF typedef_optional_modifiers type typedef_declarator SEMICOLON
+        {
+            (void)get_current_modifiers();
+            yyerror_current_line(
+                "lit declarations are only allowed at top level");
```
leo-aa88, line 1237: ## 1. "`lit` declarations are top-level only" is enforced for exactly 1 of the 8 `typedef_def` forms

**MAJOR.**

`AGENTS.md`/the PR description/the docs all state the rule as a flat restriction ("`lit` declarations are top-level only"), and this `statement:` alternative exists specifically to turn a nested `lit` into a clean semantic error instead of a raw parse failure:

```c
| TYPEDEF typedef_optional_modifiers type typedef_declarator SEMICOLON
    {
        (void)get_current_modifiers();
        yyerror_current_line(
            "lit declarations are only allowed at top level");
        typedef_had_error = true;
        SAFE_FREE($4.name);
        $$ = NULL;
    }
```

Problem: this is only *one* of the eight `typedef_def` alternatives up in `function_def_list` (plain scalar/pointer, struct/union tag alias, anonymous aggregate, named inline aggregate, enum tag alias, alias-of-alias — each with its own `dimensions` variant). None of the other seven got a matching `statement:` alternative. I checked this isn't hypothetical — built the branch and ran it:

```
$ cat t.brainrot
gang Point { rizz x; rizz y; };
skibidi main { lit gang Point PointAlias; bussin 0; }
$ ./brainrot t.brainrot
Error: syntax error, unexpected STRUCT at line 6
Parsing failed

$ cat t2.brainrot
skibidi main { lit gyatt Color Paint; bussin 0; }
$ ./brainrot t2.brainrot
Error: syntax error, unexpected ENUM at line 1
Parsing failed

$ cat t3.brainrot
skibidi main { lit gang { rizz x; } Anon; bussin 0; }
$ ./brainrot t3.brainrot
Error: syntax error, unexpected STRUCT at line 1
Parsing failed
```

vs. the one path that *is* covered:

```
$ echo 'skibidi main { lit rizz Count; bussin 0; }' | ...
Error: lit declarations are only allowed at top level at line 1
```

Both outcomes fail the parse — nothing invalid gets executed — but the diagnostic degrades to a bare bison "unexpected STRUCT/ENUM" for the other seven forms, which doesn't tell the user *why* (that `lit` isn't allowed there at all), it just looks like an internal grammar bug. `test_cases/lit_top_level_only_fail.brainrot` only exercises the one covered form (`lit rizz Count;`), so this asymmetry shipped untested.

Fix: either add the missing `statement:` alternatives for the other seven `typedef_def` productions (mechanical but matches the existing style), or — better, given how much this duplication already costs the file — factor `typedef_def`'s body out so `statement:` can reuse the *same* production and gate the "top level only" check with a single flag/parameter instead of hand-duplicating each alternative twice more.

lang.y
```diff
@@ -663,7 +1328,118 @@ declaration:
...
+        {
+            $$ = create_alias_declaration($3.name, $2, $3.pointer_level, $5);
+            SAFE_FREE($3.name);
+        }
+    | optional_modifiers alias_type declarator dimensions
+        {
+            int pointer_level = $2.pointer_level + $3.pointer_level;
+            if ($2.type == VAR_STRUCT && pointer_level == 0)
+            {
+                yyerror("Arrays of struct/union typedef aliases are not supported");
```
leo-aa88: ## 2. "Arrays of struct/union typedef aliases are not supported" — three copy-pasted rejection sites, zero test coverage

**MAJOR (missing tests).**

```c
if ($2.type == VAR_STRUCT && pointer_level == 0)
{
    yyerror("Arrays of struct/union typedef aliases are not supported");
    typedef_had_error = true;
}
```

This exact block is duplicated three times (here, and again for the `dimensions EQUALS array_init` and `dimensions_or_unsized EQUALS array_init` variants a few dozen lines down), unlike the sibling `lit T Name[N];` rejection (`reject_array_typedef()`), which is a proper shared helper. `AGENTS.md` is explicit: "Every new language feature or bug fix needs a `.brainrot` file in `test_cases/`, and a corresponding expected-output entry" — and this is a deliberate, user-facing rejection path the PR itself introduces, not an incidental one. `grep -rn "Arrays of struct" test_cases/ tests/expected_results.json` comes back empty.

I exercised it manually since nothing in the suite does:

```
$ cat t.brainrot
gang Point { rizz x; rizz y; };
lit gang Point PointVal;
skibidi main { PointVal arr[3]; bussin 0; }
$ ./brainrot t.brainrot   # ASan+UBSan build
Error: Arrays of struct/union typedef aliases are not supported at line 8
Parsing failed
```

It works, and doesn't crash — I'm not claiming a functional bug here. But three duplicated, string-identical error sites with no regression test is exactly the kind of thing that silently rots the next time someone touches array-declarator productions (e.g. adding a fourth `dimensions` variant, or trying to lift this restriction) — nothing will fail CI to tell them they missed a spot, or that they changed the message on 2 of 3 copies. Please: (a) add a `test_cases/lit_struct_alias_array_fail.brainrot` + `expected_results.json` entry, matching the pattern of every other `*_fail.brainrot` in this PR, and (b) factor the duplicated check into a `reject_struct_alias_array()` helper the way `reject_array_typedef()` already models.

leo-aa88 (commented):
# COMMENT

Verified independently before reading a line of the diff: `make` builds clean under `-Werror -fsanitize=address,undefined` with zero bison shift/reduce or reduce/reduce conflicts, all 297 tests pass (`ASAN_OPTIONS=detect_leaks=0 make test`), `make format-check` passes, and I ran valgrind (release/no-ASan build, since ASan-under-valgrind is broken in this sandbox exactly as the PR description says) against every new `lit_*`/`semantic_error_lit_*` fixture with zero leaks and zero invalid accesses. This is not a shaky patch. The ownership story is unusually disciplined for a change this size: `TypeDescriptor` is explicitly documented as a non-owning view, `TypeAlias` explicitly owns its copies, every place that stashes a descriptor's `struct_name`/`enum_name` onto something long-lived goes through `ARENA_STRDUP`/`safe_strdup` before the borrowed string can go away, and `free_type_alias_registry()` runs in `cleanup()` *after* `free_ast()`/`free_function_table()`/etc., so nothing outlives the registry it borrowed from. I went looking for the classic "descriptor says one thing, runtime does another" bug in this pattern and didn't find one.

The lexer-feedback trick (`lang.l` returning `TYPE_NAME` instead of `IDENTIFIER` once a name is a registered alias) is the correct classic solution to the C typedef/identifier ambiguity, and the grammar consistently threads both token spellings through declarations, params, and function defs via `create_typedef_name_parameter`/`create_alias_parameter`/`report_alias_name_conflict`, including the "reverse occupancy" case (a later variable/function/parameter reusing an alias name) that the last round of commits added. `test_cases/lit_tag_namespace.brainrot` and `lit_same_name_tag_alias.brainrot` are genuinely good adversarial tests — they specifically probe alias-vs-tag namespace separation in both declaration orders, which is exactly the kind of case this style of implementation tends to get wrong.

Two real gaps below. Neither is memory-unsafe and neither lets an invalid program silently produce a wrong result — both are enforcement/coverage gaps in restrictions the PR itself claims to establish.

# VERDICT

The deepest issue here is completeness, not correctness: the PR built a real mechanism for converting a "this isn't allowed here" restriction into a proper diagnostic (the `statement:`-level `TYPEDEF` alternative), then only wired it up for one of the eight syntactic shapes `typedef_def` supports, and separately introduced a rejection rule (struct-alias arrays) three times over without a single test pinning any of the three copies down. Both gaps are in scope/tests, not architecture — the underlying alias-registry design, the borrowed-vs-owned string discipline, and the lexer-feedback tagging are all sound and I could not get them to misbehave under ASan/UBSan or valgrind across the fixtures or my own adversarial cases (self-aliasing, tag/alias namespace collisions in both orders, double-pointer aliases, nested-scope rejection).

Recommend closing both gaps (or at minimum #1, since a raw "unexpected STRUCT" leaking out of 7/8 forms of a documented, tested restriction is a real regression in diagnostic quality even though it isn't a crash) before merge. Neither blocks on memory safety or the feature's actual runtime semantics.

leo-aa88 (commented):
# COMMENT (would APPROVE — GitHub blocks self-approval on your own PR)

Re-reviewed at HEAD (224d351), following up on my previous review at 1a2e476. Both MAJOR findings from that pass are fixed, correctly and without new regressions.

**Verification repeated from scratch:** clean `-Werror -fsanitize=address,undefined` build, zero bison conflicts (`-Wcounterexamples`), 305/305 tests pass (up from 297 — 8 new fixtures), `make format-check` clean, and valgrind (release build) clean across every `lit_*`/`semantic_error_lit_*` fixture with zero leaks/invalid accesses.

## Finding #1 ("top-level only" enforced for 1/8 forms) — fixed correctly

The fix is exactly the shared-production approach I suggested, not a pile of copy-pasted special cases: `typedef_def` and the `statement:` `TYPEDEF` alternative now both reduce through a single `typedef_tail` nonterminal, gated by a `typedef_is_statement`/`typedef_may_commit()` flag. The `statement:` path reports the "lit declarations are only allowed at top level" error once, up front at the `TYPEDEF` token, then still fully parses whatever follows (struct bodies, dimensions, etc.) without committing any registration — `register_type_alias`/`register_aggregate_typedef`/`register_anonymous_aggregate_typedef` and even the array-typedef/storage-class rejections are now all gated behind `typedef_may_commit()`, so a nested form doesn't double-report *and* doesn't leak a struct tag or alias registration into the global namespace on its way to failing.

I re-ran all 7 forms I'd previously shown leaking a raw bison "unexpected STRUCT/ENUM" — scalar, pointer, struct-tag alias, enum-tag alias, anonymous aggregate, named inline aggregate, and array-typedef — and every one now reports the single correct diagnostic:
```
Error: lit declarations are only allowed at top level at line 3
```
Checked for the obvious way this pattern breaks (a `YYABORT` inside `typedef_tail` would leave `typedef_is_statement` stuck `true` and silently disable every later top-level `lit` for the rest of that parse) — there's no `YYABORT` anywhere in `typedef_tail`, and the flag is defensively reset in `cleanup()` too. Test coverage now has one fixture per form (`lit_top_level_only_{struct,enum,anonymous,named_inline,alias}_fail.brainrot`), matching the `AGENTS.md` testing mandate.

## Finding #2 (struct-alias-array rejection: 3x duplicated, 0 tests) — fixed correctly

Deduplicated into `reject_struct_alias_array()`/`maybe_reject_struct_alias_array()` — `grep -n "Arrays of struct/union typedef aliases are not supported" lang.y` now returns exactly one hit, the helper definition. `test_cases/lit_struct_alias_array_fail.brainrot` + the matching `expected_results.json` entry cover it, and it stayed clean under both ASan/UBSan and valgrind.

## Bonus, unprompted

`merge_type_modifiers`'s signedness/width conflict check was previously only symmetric across *different* modifiers (`base.is_signed && extra.is_unsigned`), silently allowing a redundant `nut Count` on an already-`nut` alias through as a no-op merge. It's now `(extra.is_signed || extra.is_unsigned) && (base.is_signed || base.is_unsigned)`, correctly flagging same-modifier repetition too, with `lit_duplicate_signedness_modifier_fail`/`lit_duplicate_width_modifier_fail` pinning it down. Also worth a nod: the new `if (!var->value.array_data) { yyerror("Uninitialized struct lvalue"); return NULL; }` guard in `evaluate_lvalue_address` closes a null-blob write path defensively rather than trusting the caller.

# VERDICT

Nothing left that I can substantiate as a defect. The fix for #1 is the better of the two options I offered (shared production + flag, not eight more duplicated `statement:` alternatives), and it visibly reduced duplication in the process rather than adding to it. Ownership/lifetime discipline continues to hold under adversarial testing. Approving.

PR #105, opened 2025-01-25
ci: add static code analysis tools

araujo88:
## Description

1. Add static code analysis tools:
 - cppcheck for deep static analysis
 - clang-tidy for modern C/C++ linting
 - clang-format for style checking
 - flawfinder for security vulnerabilities
 - splint for additional checks

2. Made build job depend on static analysis.
3. Added stricter compiler flags: -Wall -Wextra -Werror -pedantic

## Type of Change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation update
- [x] Performance improvement
- [ ] Refactor

## Checklist

- [ ] My code follows the style guidelines of this project
- [ ] I have performed a self-review of my own code
- [ ] I have documented my changes in the code or documentation
- [ ] I have added tests that prove my changes work (if applicable)
- [ ] I have run the unit tests locally
- [ ] I have run the valgrind memory tests locally
- [ ] All new and existing tests pass

PR #240, opened 2026-08-24
ci: add cppcheck static-analysis job (phase 2 of #172)

leo-aa88:
## Description

Phase 2 of the CI static-analysis adoption plan tracked in #172: adds a
`make cppcheck` target, a checked-in suppressions file, and a required
`static-analysis` CI job (`lint` -> `static-analysis` -> `build`) that gates
on `error`/`warning`/`performance`/`portability` findings.

- **`make cppcheck`** (Makefile): `cppcheck --check-level=exhaustive`, guarded
  on a minimum version of 2.13 (`--check-level=exhaustive` needs >= 2.11,
  `nullPointerOutOfMemory` needs >= 2.12; Ubuntu 22.04's apt cppcheck, 2.7, is
  too old for either). Scans the same translation units `all` builds, minus
  generated Flex/Bison output and `tests/` (whose `badnatives/*.c` fixtures
  are deliberately malformed).
- **`cppcheck-suppressions.txt`**: two narrowly-scoped, individually justified
  suppressions — `lib/mem.c`'s interior-pointer allocator (a false-positive
  `memleak`; `make valgrind` is the real gate there) and
  `stdrot/registry.c:63`'s `__stop`/`__start` linker-section-bounds
  subtraction (a `subtractPointers` false positive, the intended ELF
  section-iteration idiom). No blanket suppressions, unlike #105.
- **`lib/arena.c`**: fixed 7 genuine unchecked `malloc()` dereferences in
  `region_new()`/`arena_alloc()`, routed through the existing
  `handle_malloc_error()` convention already used in `lib/mem.c`. Also frees
  the locally-allocated `Arena` on the first-region-allocation failure path,
  to avoid introducing a real leak on that new error path.
- **`ast.c`**: removed 3 dead-code branches flagged by cppcheck — two
  self-canceling `&&`/`||` expressions after an early-exit in
  `evaluate_expression_int`/`evaluate_expression_bool`, and an always-false
  nested condition inside `execute_statement`'s `NODE_ARRAY_ACCESS` case.
  Inline-suppressed the one flagged site (`matched || !matched` in
  `execute_switch_statement`) that's already tracked as issue #179, since
  fixing it is a switch/default *semantics* change that needs its own
  `test_cases/` fixture — matching how the `Makefile` already documents and
  suppresses the clang equivalent for the wasm/Darwin release builds.
- **`.github/workflows/ci.yml`**: new `static-analysis` job wired between
  `lint` and `build` (`lint -> static-analysis -> build -> test -> ...`).
- **Docs**: `make cppcheck` documented in `AGENTS.md`'s Setup/Build/Test table
  and Boundaries section, and in `CONTRIBUTING.md`'s prerequisites/workflow.

`style` severity findings (62 more, mostly const-correctness churn across
`ast.c`/`interpreter.c`/`semantic_analyzer.c`) are intentionally deferred to a
follow-up issue, to avoid #105's mega-PR conflict pile — this PR fixes exactly
the 14 findings at the gate's severity level.

## Related Issue

Part of #172

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement
- [x] Refactor

## Checklist

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -44,9 +44,41 @@ jobs:
       - name: Check formatting
         run: make format-check
 
-  build:
+  static-analysis:
     runs-on: ubuntu-24.04
     needs: lint
+
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v7
+
+      - name: Force Ubuntu archive mirror
+        run: |
+          sudo sed -i \
+            's|azure.archive.ubuntu.com|archive.ubuntu.com|g' \
+            /etc/apt/apt-mirrors.txt
+          cat /etc/apt/apt-mirrors.txt
+
+      - name: Install cppcheck
+        run: |
+          sudo apt-get \
+            -o Acquire::Retries=3 \
+            -o Acquire::http::Timeout=15 \
+            -o Acquire::https::Timeout=15 \
+            update
+          sudo apt-get \
+            -o Acquire::Retries=3 \
+            -o Acquire::http::Timeout=15 \
+            -o Acquire::https::Timeout=15 \
+          install cppcheck -y
+          cppcheck --version
+
+      - name: Run cppcheck
+        run: make cppcheck
+
+  build:
+    runs-on: ubuntu-24.04
+    needs: static-analysis
 
     steps:
       - name: Checkout code
--- a/Makefile
+++ b/Makefile
@@ -4,6 +4,7 @@ BISON := bison
 FLEX := flex
 PYTHON := python3
 EMCC := emcc
+CPPCHECK ?= cppcheck
 
 # Compiler and linker flags
 CFLAGS := -Wall -Wextra -Wpedantic -Werror -O2 -Wuninitialized -fsanitize=address,undefined -fno-omit-frame-pointer -g
@@ -41,6 +42,13 @@ ALL_SRCS := $(SRCS) $(GENERATED_SRCS)
 STDROT_DIR := stdrot
 STDROT_SRCS := $(wildcard $(STDROT_DIR)/*.c) $(SRC_DIR)/input.c
 STDROT_LIB := libstdrot.so
+
+# Sources scanned by cppcheck: the same translation units `all` builds, minus
+# the generated Flex/Bison output (never scan or hand-edit lang.tab.c /
+# lex.yy.c). tests/ is deliberately excluded: tests/badnatives/*.c are
+# intentionally malformed registries (see that directory's own file comment),
+# so a clean cppcheck run over them would be meaningless.
+CPPCHECK_SRCS := $(SRCS) $(STDROT_SRCS)
 
 # Test-only stdrot library: production natives plus tests/stdrot/*.c
 # (bindings that exist solely so test_cases/*.brainrot can exercise ABI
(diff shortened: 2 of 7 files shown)


.github/workflows/ci.yml
```diff
@@ -44,10 +44,42 @@ jobs:
...
+            -o Acquire::https::Timeout=15 \
+          install cppcheck -y
+          cppcheck --version
+
+      - name: Run cppcheck
+        run: make cppcheck
+
+  build:
+    runs-on: ubuntu-24.04
+    needs: static-analysis
```
leo-aa88: **[suggestion]** **MINOR.** Issue #172's Phase 2 contract is a `static-analysis` job that runs `make cppcheck` and fails on real findings. Phase 4 — after clang-tidy exists — is when `build`/`test` should `needs: [lint, static-analysis]` *or* stay parallel and be marked required. The original issue also listed "making `build` `needs:` analysis on day one" as out of scope, with the explicit risk mitigation "parallel jobs until stable." This PR inserts `static-analysis` into the critical path (`lint` → `static-analysis` → `build` → `test` → release/wasm). Phase 1 already gated `build` on `lint`, so this matches the repo's current pattern, but it still pulls Phase 4 wiring into a cppcheck-only PR: a check-ID drift (already seen once: 2.13 `comparePointers` vs 2.17 `subtractPointers`) now skips compile, pytest, Valgrind, and every dry-run rather than failing one required job beside a still-green `build`.

**Suggestion:** Keep `build: needs: lint` (or no analysis edge) until Phase 3 is green, and make `static-analysis` required via branch protection. If serial fail-fast is intentional, say so in #172 rather than calling this Phase 2.

Makefile
```diff
@@ -326,6 +334,36 @@ format-check:
...
+	-j4
+
+# Static analysis with cppcheck (used by CI's static-analysis job). `style`
+# checks are deliberately not enabled here -- see cppcheck-suppressions.txt
+# and issue #172 for why that's a separate follow-up, not this gate.
+.PHONY: cppcheck
+cppcheck:
+	@command -v $(CPPCHECK) >/dev/null 2>&1 || { echo "Error: cppcheck not found. Blud, install cppcheck >= $(CPPCHECK_MIN_VERSION)!"; exit 1; }
+	@ver=$$($(CPPCHECK) --version | awk '{print $$2}'); \
+	oldest=$$(printf '%s\n%s\n' "$(CPPCHECK_MIN_VERSION)" "$$ver" | sort -V | head -n1); \
```
leo-aa88: **[suggestion]** **MINOR.** `make cppcheck` is now an AGENTS.md/CONTRIBUTING.md "always run before a PR" target. The version gate does `sort -V` on `cppcheck --version`. CI is Ubuntu 24.04 (GNU sort, cppcheck 2.13.0-2ubuntu3 — confirmed in the green `static-analysis` log, which also proves `--error-exitcode=1` works with `-j4`). BSD `sort` (macOS, a first-class release target) has no `-V`. Without `pipefail`, `sort -V` fails, `head` succeeds on empty input, `oldest` is empty, and the recipe reports "cppcheck $ver is too old" even when the binary is 2.13+. CI will never see this.

**Suggestion:** Compare versions without GNU `sort -V` (e.g. `printf` + `awk` on major.minor, or `dpkg --compare-versions` only in CI). Don't rely on a Linux-only idiom for a documented contributor gate.

Makefile
```diff
@@ -326,6 +334,36 @@ format-check:
...
+	--inline-suppr \
+	--suppressions-list=cppcheck-suppressions.txt \
+	--error-exitcode=1 \
+	--std=c11 --language=c --platform=unix64 \
+	-I. -I$(SRC_DIR) -I$(STDROT_DIR) \
+	--suppress=missingIncludeSystem \
+	-j4
+
+# Static analysis with cppcheck (used by CI's static-analysis job). `style`
+# checks are deliberately not enabled here -- see cppcheck-suppressions.txt
```
leo-aa88: **[nit]** The new `cppcheck` recipe comment says style is deferred — "see cppcheck-suppressions.txt and issue #172." `cppcheck-suppressions.txt` only documents the `memleak` and `subtractPointers`/`comparePointers` FPs. It does not mention `style` or the follow-up. The rationale lives in the PR body and #172, not in that file.

**Suggestion:** Point the comment at #172 only, or add one line to `cppcheck-suppressions.txt` that `--enable=style` is intentionally not in `CPPCHECK_FLAGS`.

.github/workflows/ci.yml
leo-aa88: **[suggestion]** `build: needs: static-analysis` (and therefore test / valgrind / release-dry-run, which already `needs: build`/`test`) still serializes the whole pipeline behind cppcheck. Issue #172 listed "Making `build` `needs:` analysis on day one" as out of scope for this work, put that wiring in Phase 4, and named the mitigation "Parallel jobs until stable." This PR is labeled phase 2 and does not claim to close that gap. The risk is not theoretical: this PR's own first CI run failed because ubuntu-24.04's cppcheck 2.13.0 reports `comparePointers` where 2.17.1 reports `subtractPointers` — with this `needs:` edge that check-ID drift blocked compile and tests, not just the new analysis job.

**Suggestion:** Drop `needs: static-analysis` from `build` (keep `needs: lint` if you want to preserve the Phase 1 DAG) so analysis runs in parallel with compile/test until Phase 4. Mark `static-analysis` required in branch protection instead of coupling it to `build`.

Issue #192, opened 2026-08-18
Interpreter crash: consuming a struct member value in a binary expression or plain assignment


leo-aa88:
## Description

Combining a struct member access with another operand in a binary
expression, or assigning one directly to a plain scalar variable, crashes
the interpreter (ASan SEGV). This is unrelated to #191 (arrays/structs in
functions) and reproduces on `main` without that branch's changes — found
incidentally while stress-testing that PR.

## Repro

```c
gang Point {
    rizz x;
    rizz y;
};

skibidi main {
    gang Point p;
    rizz s;
    p.x = 5;
    s = p.x;
    yapping("%d", s);
    bussin 0;
}
```

```
$ ./brainrot repro.brainrot
Error: Unsupported variable type at line 13
ast.c:2586:16: runtime error: load of null pointer of type 'int'
AddressSanitizer:DEADLYSIGNAL
==...==ERROR: AddressSanitizer: SEGV on unknown address 0x000000000000 ...
    #0 ... in evaluate_expression_int ast.c:2655
    #1 ... in write_value_to_address ast.c:1776
    #2 ... in execute_assignment ast.c:3169
```

A second, related shape also crashes/misbehaves — a struct field combined
with another value via a binary operator:

```c
gang Point {
    rizz x;
    rizz y;
};

skibidi main {
    gang Point p;
    p.x = 3;
    p.y = 4;
    yapping("%d", p.x + p.y);
}
```
This one reports `Error: Member access on non-struct/union value` from the
semantic analyzer for chained/combined struct accesses rather than
crashing outright, but either way the underlying member value never
becomes usable as an operand.

## Where it breaks

`s = p.x;`: `execute_assignment` → `write_value_to_address` →
`evaluate_expression_int(p.x)`. The trace shows this landing in the
`NODE_ARRAY_ACCESS` case of `evaluate_expression_int` rather than
`NODE_STRUCT_ACCESS`, which has its own (working, when reached directly
via e.g. `bussin p.x;`) handling — something about how `p.x`'s value
becomes an operand for assignment/binary evaluation is going through the
wrong path or getting the wrong node.

`p.x + p.y`: the semantic analyzer's `NODE_STRUCT_ACCESS` resolution
appears to only work reliably when the struct access is evaluated in
isolation (e.g. as a `bussin p.x;` return expression, or as a function-call
argument to a builtin like `yapping`), not as an operand of a binary
operation.

## What works today (for contrast)

[...]

PR #247, opened 2026-08-25
feat(struct): fixed-size arrays as struct fields

### LINKED ISSUE #206: [Roadmap] Phase 3 — C-compatible aggregates
> **Roadmap tracking:** #214
> **Depends on:** #205
> **Blocks / unblocks:** #208 (Road B), #211 L3 (HTTP gangs)

---

## Summary

This is the big one. raylib is struct city (`Vector2`, `Color`, `Texture2D`, `Camera2D`, `Matrix`, …). Brainrot's `gang` must become a C-compatible, first-class aggregate: honest layout, a real type descriptor, nested member access, and struct-typed fields/params/returns.

| | |
| --- | --- |
| **Roadmap** | Phase 3 — C-compatible aggregates |
| **Milestone** | M3 — Layout is honest (3a + 3d) then M4 — Aggregates are first-class (3b, 3c, 3e) |
| **Priority** | P0 |
| **Status** | not started |
| **Depends on** | Phase 2 |
| **Blocks** | Phase 5 Road B (real raylib binding), Phase 8 L3 (HTTP `gang Request` / `gang Response`) |
| **Baseline** | `0cfa958` |

Do **3a first**. It is a live correctness bug: any `gang` handed to C today would be silently misinterpreted.

### Verified limitations this phase closes

| # | Limitation | Evidence |
| - | ---------- | -------- |
| L6 | `compute_struct_layout()` packs fields sequentially with no alignment or padding, so `gang` layouts do not match C. | `ast.c:3948` |
| L7 | Struct member access only resolves when the base is a plain identifier; `a.b.c` is rejected outright. | `ast.c:346` |
| L8 | Functions can take/return a struct by value, but only as a plain struct variable of the exact matching type — not a sub-expression. | `ast.c:4480` |
| L9 | `StructField` keeps `VarType + pointer_level + offset` only — no struct type name, no arrays, no modifiers, so nested-struct and fixed-array fields are unrepresentable. | `ast.h:82` |

### Related open issues (do not duplicate — close or narrow these when the matching letter lands)

- #192 — interpreter crash consuming a struct member in a binary expression / assignment
- #193 — struct-by-value sub-expressions and pointer-to-struct args/returns (**L8 / 3c**)
- #194 — arrays as function parameters/return values by value
- #196 — pointer-to-struct/union variables silently give wrong results
- #197 — chained member access through a pointer-typed field (`a.ptr.b`) (**subset of 3d**)

### 3a. Fix the layout (do this first — live bug)

`ast.c:3948` packs fields with no padding. Required:

```c
offset        = align_up(offset, alignof(field));
field->offset = offset;
offset       += sizeof(field);
struct_size   = align_up(offset, max_field_alignment);
```

And prove it, don't assume it:

```c
_Static_assert(sizeof(BrainrotVector2) == sizeof(Vector2), "ABI drift");
_Static_assert(offsetof(BrainrotVector2, y) == offsetof(Vector2, y), "ABI drift");
```

- [ ] Alignment + trailing padding in `compute_struct_layout()` (and the union equivalent)
- [ ] Runtime `offsetof` / size checks against real C headers for a few golden structs
- [ ] Tests that would fail today if a C function were handed a packed `gang`

### 3b. Introduce a real type descriptor

`VarType + pointer_level + modifiers` is at the end of its rope. Replace it with:

```c
typedef struct TypeDesc {
    TypeKind    kind;
    bool        is_unsigned;
    int         pointer_level;
    StructDef  *struct_def;      /* kind == TYPE_STRUCT */
    int         array_rank;
    size_t      dimensions[MAX_DIMENSIONS];
} TypeDesc;
```

This is invasive — it touches the AST, the analyzer, and the interpreter — but every remaining item on the roadmap gets cheaper afterwards.

Open question (Appendix B #2): one large refactor, or incremental with both representations alive? **Recommendation: incremental, both alive.** Safer and uglier.

- [ ] Add `TypeDesc` alongside `VarType`
- [ ] Migrate analyzer, then interpreter, then drop the old representation
- [ ] No behavior change on existing fixtures during the dual-representation window

### 3c. `gang` as a first-class type

Struct fields, parameters, and return types currently come from the primitive `type` grammar, so a struct-typed field cannot even be spelled. Target:

```c
gang Vector2 { chad x; chad y; };

gang Camera2D {
    gang Vector2 offset;
    gang Vector2 target;
    chad rotation;
    chad zoom;
};

gang Vector2 GetMousePosition();
skibidi DrawCircleV(gang Vector2 center, chad radius, gang Color color);
```

- [ ] Struct-typed fields / params / returns in the grammar (`lang.l` / `lang.y`)
- [ ] Struct arguments and returns (removing the identifier-only restriction at `ast.c:2458` — see #193)
- [ ] Struct assignment semantics — **decision needed**: value copy, matching C (Appendix B #3)

### 3d. Nested member access

Rewrite member access as `base lvalue address + field offset`, recursively, instead of special-casing `NODE_IDENTIFIER` at `ast.c:346`. Once access is address-based, these all fall out — including as assignment targets:

- `camera.target.x`
- `model.transform.m0`
- `renderTexture.texture.width`

#197 (`a.ptr.b` through a pointer-typed field) is a subset; #196 (pointer-to-struct variables) is a likely prerequisite.

- [ ] Address-based member access, recursive
- [ ] Nested access as an assignment target
- [ ] Pointer-field chaining (coordinate with #196 / #197)

### 3e. Remaining C field types

For full raylib coverage:

- [ ] Unsigned and fixed-width scalars
- [ ] Pointer fields
- [ ] Nested struct fields
- [ ] Fixed arrays inside structs (`char name[32]`, `float params[4]`)
- [ ] Struct aliases
- [ ] Arrays / pointers to structs

`gyatt` (enum), `lit` (typedef), and `chungus` (union) are **not** blockers — a generator can emit `CAMERA_PERSPECTIVE`, `KEY_SPACE`, and `MOUSE_BUTTON_LEFT` as ordinary Brainrot integer constants. They stay on the wishlist.

## Compatibility / open questions

- Appendix B #2 — `TypeDesc` migration strategy
- Appendix B #3 — struct assignment: value copy (C) or reference?
- Appendix B #6 — ownership of native resources (textures, etc.). Handles keep ownership in C.

Per `AGENTS.md`: changing existing keyword syntax/semantics in `lang.l`/`lang.y` needs sign-off. This phase is **additive** to `gang` but touches its layout and assignment — treat layout as a bugfix (silently-wrong ABI) and assignment as a semantics decision that must be written down.

## Files likely to change

- `ast.c` / `ast.h` — layout, member access, `StructField`, `TypeDesc`
- `lang.l` / `lang.y` — struct-typed fields/params/returns
- `semantic_analyzer.c`, `interpreter.c`, `visitor.c`
- `stdrot/stdrot_api.h` — `STDROT_STRUCT` values
- ABI test C files with `_Static_assert` / `offsetof`
- `test_cases/` for nested access, layout, struct-typed fields
- README / `docs/` "Current Limitations" once L6–L9 die

### M3 — Layout is honest
- [ ] `gang` layout matches C (`alignof` + trailing padding)
- [ ] `_Static_assert` / `offsetof` checks against real C headers
- [ ] `a.b.c` works, including as an assignment target

### M4 — Aggregates are first-class
- [ ] `TypeDesc` in use (or a documented incremental dual-rep with a follow-up to delete `VarType`)
- [ ] `gang Vector2 GetMousePosition()` and `skibidi DrawCircleV(gang Vector2, ...)` parse, analyze, and run
- [ ] Struct assignment semantics decided and tested
- [ ] Remaining C field types needed for raylib's common structs
- [ ] #193 / #197 either implemented here or explicitly left open with a reason

## Guiding principles that apply

- **Layout correctness is testable — so test it.** Every ABI claim gets a `_Static_assert` or a runtime `offsetof` check.
- Almost entirely memory-layout work; treat any new leak/UB as a blocker.

## References

- [docs/ROADMAP.md — Phase 3](https://github.com/Brainrotlang/brainrot/blob/main/docs/ROADMAP.md#phase-3--c-compatible-aggregates)
- #192, #193, #194, #196, #197

leo-aa88:
## Description

Implements slice 2 of #206 (3e): fixed-size arrays as struct fields, e.g. `gang Foo { chad params[4]; };` then `f.params[i]` read/write, for primitive scalar and pointer element types, 1-D or N-D.

**Representation.** `StructField`/`Parameter` both gain `is_array`/`array_dimensions`. `Array` (the `NODE_ARRAY_ACCESS` payload) gains `ASTNode *base` — non-NULL for the new `struct_access multi_dimension_access` grammar form (`foo.arr[i]`, `base` = the `struct_access` expression `foo.arr`), NULL for the existing name-based `IDENTIFIER multi_dimension_access` form. Exactly one of `name`/`base` identifies what's being indexed.

**Layout.** `get_struct_field_size()` multiplies the existing per-element-type size by the product of every dimension when `is_array`. `get_struct_field_alignment()` needs **no change at all** — a C array's alignment is always just its element's, never larger, so the existing per-type switch already answers array fields correctly.

**Evaluation.** Almost every array-element read/write already dispatches correctly by type once handed a correctly-typed address — `evaluate_expression_{int,float,double,short,bool}()`, `write_value_to_address()`, and `evaluate_lvalue_address()` all route `NODE_ARRAY_ACCESS` through **one** function, `evaluate_multi_array_access()`, to get that address. That function gained a struct-field-backed sibling, `evaluate_struct_field_array_access()` (kept separate rather than interleaved, so the name-based path's several rounds of hard-won narrow bugfixes stay undisturbed), dispatched to whenever `Array.base` is set. `calculate_array_offset()` lost its `Variable *` parameter in favor of the two fields it actually used (`is_array`, `array_dimensions`), so a `StructField` can reuse it without a fake `Variable` standing in for one. `get_expression_type`, `get_expression_pointer_level`, `is_expression`, and `infer_runtime_expression_type_noeval` each needed the same "resolve via the base's struct field instead of a Variable" branch — found by grepping every `node->data.array.name` use across `ast.c`/`interpreter.c`/`semantic_analyzer.c` rather than fixing sites one crash at a time.

**Semantic analyzer.** Needed the identical treatment for its own, separate copies of this logic (`infer_expression_type`/`infer_expression_pointer_level`, each with the established runtime-resolve-first/static-`infer_struct_def_static`-fallback two-path pattern already used for `NODE_STRUCT_ACCESS`, since Phase 1 static analysis runs before any runtime `Variable` exists) plus making the main traversal visit `Array.base` so `foo.arr[i]` gets the same struct/field validation `foo.arr` alone would.

**Struct copy is free.** Array fields live inline in the struct's single `calloc`'d blob, so the existing deep-copy `memcpy` (assignment, by-value parameter) already gives a copy's array field independent storage — verified empirically (`struct_array_field_copy.brainrot`) rather than assumed.

### Out of scope for this slice (see plan)

- Nested-struct-typed array elements (`gang Foo { gang Bar arr[4]; }`).
- Struct-literal braced initializers for the array field itself.
- Modifier-prefixed array fields (`struct_field` has no modifier-prefixed production at all yet — same pre-existing limitation scalar fields have via `lit` aliasing only).

### Discovered, not fixed (pre-existing, already tracked)

While testing pointer-element array fields I found two **pre-existing, unrelated** gaps, confirmed with plain non-array reproductions so I'm confident they aren't something this PR introduces:

1. Dereferencing a pointer-typed struct field via `*a.ptr` (array or not) is rejected at semantic-analysis time — the same "chaining through a pointer field" gap issues #196/#197 already track for `.` chaining.
2. Comparing two struct-member-access expressions in one binary operation (`a.x == a.x`, **no arrays involved at all**) reproduces issue #192 ("Interpreter crash: consuming a struct member value in a binary expression or plain assignment") exactly — confirmed with a plain scalar field, zero array-access grammar involved.

Both are out of this slice's scope; its tests deliberately don't exercise either path (`struct_array_field.brainrot` and friends read/write elements directly and via `%f`/`%d`/`%c` printf args, never compare two member-access expressions or dereference a pointer field).

## Related Issue

Part of #206 (item 3e, slice 2 of the implementation plan). Does not close #206.

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement
- [ ] Refactor

## Checklist

Note: `make cppcheck` could not be run in this environment (no `cppcheck` available, and the distro package is below the >= 2.13 version this project's Makefile gates on); CI's `static-analysis` job will verify it.


ast.c
```diff
@@ -1530,6 +1676,19 @@ VarType get_expression_type(ASTNode *node)
...
+        if (node->data.array.base)
+        {
+            StructDef *def = NULL;
+            void *base = NULL;
+            StructField *fld = NULL;
+            if (resolve_struct_access(node->data.array.base, &def, &base, &fld,
+                                      false) &&
+                fld->is_array)
+                return fld->type;
+            yyerror("Undefined array field in expression");
```
leo-aa88: probe

leo-aa88 (commented):
probe: ast.c:1688

ast.c
```diff
@@ -4404,6 +4573,17 @@ bool is_expression(ASTNode *node, VarType type)
...
     case NODE_ARRAY_ACCESS:
     {
+        if (node->data.array.base)
+        {
+            StructDef *def = NULL;
+            void *base = NULL;
+            StructField *fld = NULL;
+            if (resolve_struct_access(node->data.array.base, &def, &base, &fld,
+                                      false))
+                return fld->type == type;
```
leo-aa88: probe

leo-aa88 (commented):
probe: ast.c:4583

ast.c
```diff
@@ -1418,6 +1554,16 @@ int get_expression_pointer_level(ASTNode *node)
...
     case NODE_ARRAY_ACCESS:
     {
+        if (node->data.array.base)
+        {
+            StructDef *def = NULL;
+            void *base = NULL;
+            StructField *fld = NULL;
+            if (resolve_struct_access(node->data.array.base, &def, &base, &fld,
+                                      false))
+                return fld->pointer_level;
```
leo-aa88: probe

leo-aa88 (commented):
probe: ast.c:1564

ast.c
```diff
@@ -1759,6 +1918,16 @@ static VarType infer_runtime_expression_type_noeval(ASTNode *expr)
...
     case NODE_ARRAY_ACCESS:
     {
+        if (expr->data.array.base)
+        {
+            StructDef *def = NULL;
+            void *base = NULL;
+            StructField *fld = NULL;
+            if (resolve_struct_access(expr->data.array.base, &def, &base, &fld,
+                                      false))
+                return fld->type;
```
leo-aa88: probe

leo-aa88 (commented):
probe: ast.c:1928

semantic_analyzer.c
```diff
@@ -634,6 +659,28 @@ VarType infer_expression_type(ASTNode *node, SemanticAnalyzer *analyzer)
...
+        {
+            /* Same two-path resolution as infer_expression_pointer_
+               level()'s own NODE_ARRAY_ACCESS case above -- see that
+               case's comment. */
+            StructDef *def = NULL;
+            void *base = NULL;
+            StructField *fld = NULL;
+            if (resolve_struct_access(node->data.array.base, &def, &base, &fld,
+                                      false))
+                return fld->type;
```
leo-aa88: probe

leo-aa88 (commented):
probe: semantic_analyzer.c:672

tests/expected_results.json
```diff
@@ -154,6 +154,13 @@
     "recursive_local_array": "15\n",
     "same_name_locals_no_collision": "100 200\n",
     "struct_param_by_value": "999\n3\n",
+    "struct_array_field": "1.500000 2.500000 3.500000 4.500000 42\n20\n",
+    "struct_array_field_char": "rizz\n",
+    "struct_array_field_copy": "a=1.000000 b=2.000000\nafter mutate: a=1.000000\n",
+    "struct_array_field_2d": "1 2 3 4 5 6\n24\n",
+    "struct_array_field_chain": "10 20 30\n",
+    "struct_array_field_not_array_fail": "Error: Undefined array field in expression at line 11\nError: Struct/union field 'n' is not an array at line 11",
```
leo-aa88: probe

leo-aa88 (commented):
probe: tests/expected_results.json:162

leo-aa88 (commented):
probe: ast.c:596

ast.c
```diff
@@ -536,20 +587,19 @@ size_t calculate_array_offset(Variable *var, int indices[], int num_indices)
...
     // try to handle it gracefully instead of crashing
-    if (!var->is_array)
+    if (!is_array)
     {
         // Variable is not an array - return offset 0 for single element access
         return 0;
     }
 
-    if (num_indices != var->array_dimensions.num_dimensions)
+    if (num_indices != dims->num_dimensions)
```
leo-aa88: probe

ast.c
```diff
@@ -728,6 +778,89 @@ void *evaluate_struct_member_address(ASTNode *node)
...
+                     "Missing index %d for array field '%.100s'", i,
+                     fld->name.data ? fld->name.data : "?");
+            yyerror(error_msg);
+            exit(EXIT_FAILURE);
+        }
+        indices[i] = evaluate_expression_int(index_node);
+    }
+
+    size_t offset = calculate_array_offset(
+        fld->is_array, &fld->array_dimensions, indices, num_indices);
```
leo-aa88: probe

leo-aa88 (commented):
probe: ast.c:835

ast.c
```diff
@@ -517,6 +517,50 @@ ASTNode *create_multi_array_access_node(String name, ASTNode *indices[],
...
+    void *field_base = NULL;
+    StructField *fld = NULL;
+    if (base->type == NODE_STRUCT_ACCESS &&
+        resolve_struct_access(base, &def, &field_base, &fld,
+                              /* report_errors */ false))
+    {
+        node->var_type = fld->type;
+        node->pointer_level = fld->pointer_level;
+        node->modifiers = fld->modifiers;
+        node->is_array = fld->is_array;
```
leo-aa88: probe

leo-aa88 (commented):
probe: ast.c:558

ast.c
```diff
@@ -6452,24 +6638,42 @@ StructField *find_struct_field(StructDef *def, const String name)
...
-    size_t fsz = get_type_size_for_descriptor(f->type, 0, f->modifiers);
-    if (fsz == 0)
-        fsz = sizeof(int);
-    return fsz;
+    if (!f->is_array)
+        return element_size;
+
+    size_t count = 1;
+    for (int i = 0; i < f->array_dimensions.num_dimensions; i++)
+        count *= (size_t)f->array_dimensions.dimensions[i];
```
leo-aa88: probe

leo-aa88 (commented):
probe: ast.c:6675

test_cases/struct_array_field.brainrot
```diff
@@ -0,0 +1,21 @@
...
+    gang Vec v;
+    v.n = 42;
+    v.params[0] = 1.5;
+    v.params[1] = 2.5;
+    v.params[2] = 3.5;
+    v.params[3] = 4.5;
+    yapping("%f %f %f %f %d", v.params[0], v.params[1], v.params[2],
+           v.params[3], v.n);
+    yapping("%lu", maxxing(v)); 🚽 4 floats (16) + 1 int (4) = 20
+}
```
leo-aa88: probe

leo-aa88 (commented):
probe: test_cases/struct_array_field.brainrot:21

ast.c
leo-aa88: **BLOCKING.** Contract lie + split with siblings in this same PR.

This branch requires `fld->is_array`, then on failure always emits `"Undefined array field in expression"`.

Concrete counterexample:

```brainrot
gang Foo { rizz n; };
skibidi main { gang Foo f; f.n[0] = 1; }
```

`n` **is** defined. It is simply not an array. The field resolved; `fld->is_array` is false. Calling that "undefined" is false.

Worse: `is_expression` / `get_expression_pointer_level` / `infer_runtime_expression_type_noeval` (and the semantic_analyzer mirrors) return `fld->type` / `fld->pointer_level` **without** the `is_array` gate. So one query says the expression is typed; this one says it doesn't exist.

Prefer one helper: resolve array-field element type/level, or fail with "field is not an array". Do not affirm the type in `is_expression` for an illegal index.

ast.c
leo-aa88: **BLOCKING.** Same finding, opposite side of the contract.

`resolve_struct_access` succeeded → `return fld->type == type` with **no** `fld->is_array` check.

So for `f.n[0]` where `n` is a scalar `rizz`:
- `is_expression(node, VAR_INT)` → **true**
- `get_expression_type(node)` → error + `NONE` ("Undefined array field")

THE CODE SAYS (here): typed int expression.  
THE CODE SAYS (`get_expression_type`): undefined.

That is not a diagnostic nit. That is two public type queries disagreeing about whether the expression is well-typed. Runtime later rejects in `evaluate_struct_field_array_access` — so you get a second, different error. Pick one rule and use it everywhere.

ast.c
leo-aa88: **BLOCKING** (same contract split).

Returns `fld->pointer_level` whenever the field resolves — including when `!fld->is_array`. Pair with `get_expression_type`'s `is_array` gate and `is_expression`'s unguarded type match: three helpers, three different notions of "valid array-field access."

ast.c
leo-aa88: Same split: returns `fld->type` with no `is_array` check, while `get_expression_type` one screen away requires it and then mislabels the failure. Whatever rule you pick, this path has to share it.

semantic_analyzer.c
leo-aa88, line 681: Static twin of the runtime split: `infer_expression_type` returns `fld->type` (and the pointer_level case above returns `fld->pointer_level`) without requiring `fld->is_array`. Phase 1 can therefore treat `f.n[0]` as a typed int while runtime `get_expression_type` calls it undefined. One shared resolution helper — or the static path must reject non-array fields too.

tests/expected_results.json
leo-aa88: **BLOCKING** (tests certifying a false contract).

This expected string bakes in `"Undefined array field in expression"` **and** the later honest `"is not an array"` error.

The first message is wrong: the field exists. The suite now treats a known-bad diagnostic as golden output. Fix the type-query path, then expect a single accurate failure.

semantic_analyzer.c
```diff
@@ -3092,6 +3139,15 @@ void semantic_analyze_with_scope_tracking(SemanticAnalyzer *analyzer,
            in use, otherwise it's the single-dimension index. Needed so an
            index expression like `arr[bet(2)]` gets the same native-call
            arity/type checking as any other expression. */
+        if (node->data.array.base)
```
leo-aa88, line 3151: **MAJOR.** The surrounding comment claims this `Array.base` walk "Mirrors `ast_accept()`'s `NODE_ARRAY_ACCESS` handling (visitor.c)".

It does not. `visitor.c` still only walks `indices[]` / legacy `index` — it never visits `node->data.array.base`. And `visitor.c` is not even in this PR.

ಠ_ಠ You discovered the child needed walking, patched the analyzer by hand, and left the shared visitor ignorant of the representation you just introduced.

Traversal ownership is now: visitor for name-based arrays, special-case for struct-field arrays. The next `ast_accept` consumer that assumes every child is visited will miss the entire `struct_access` subtree under `base`.

Fix: walk `Array.base` in `visitor.c`. Then the "mirrors visitor" claim is true.

ast.c
leo-aa88, line 605: **MAJOR.** You correctly stopped requiring a fake `Variable *`. You also reused this TEMPORARY FIX soft-fail for struct fields.

Dimension mismatch here truncates `num_indices` and keeps going. That was a Variable-lookup workaround. For an inline field in a `calloc`'d struct blob, under-ranked access like `g.cells[1]` on `rizz cells[2][3]` silently computes the wrong element instead of rejecting rank — adjacent fields make that a corruption class, not just a wrong print.

This PR did not invent the soft-fail; it **widened its blast radius** into the layout #206 cares about.

Struct-field path (`evaluate_struct_field_array_access`) must reject `num_indices != dims->num_dimensions` hard. Add a fail fixture. Do not share the Variable workaround with aggregate layout.

ast.c
leo-aa88, line 895: Call site that inherits the soft-fail above. Before calling into `calculate_array_offset`, assert rank matches `fld->array_dimensions.num_dimensions` (or teach `calculate_array_offset` a strict mode and use it here). Silent truncate into a neighbor field's bytes is not "matching C array semantics."

ast.c
leo-aa88: **MINOR.** A `NODE_ARRAY_ACCESS` denotes an **element**. Copying `fld->is_array` (true) onto the access node marks an element expression as an array.

Same smell as the name-based creator — but you are extending it. If anything downstream keys off `node->is_array` for "whole array" vs "element", this lies. Prefer `node->is_array = false` for access nodes (element result), and keep array-ness on the field/Variable only.

Also: commit message claims every `node->data.array.name` use was audited. `get_struct_def_for_expression`'s `NODE_ARRAY_ACCESS` case is still name-only (`ast.c` ~1833). Out of scope for nested-struct elements today — but the audit claim is false, and duplicating `if (array.base)` everywhere is how the next slice misses a site. Prefer one `resolve_array_access_target(...)`.

ast.c
leo-aa88, line 6751: **MAJOR** (tests / #206 layout principle), not a size-math bug.

Product-of-dimensions sizing looks right for the happy path (`float[4]+int` → 20 via `maxxing`). Missing adversarial coverage for the claim set:

1. **Pointer element fields** — PR text claims them; no `rizz *ptrs[N]` fixture.
2. **Padding after a short array** — e.g. `yap a[3]; rizz n;` must land `n` at align 4 (size 8). #206: "Layout correctness is testable — so test it." `maxxing` alone only rechecks Brainrot's own `total_size`.
3. **Rank mismatch** on a field (see `calculate_array_offset` soft-fail).

CI is green. That is not proof of the stated contract.

test_cases/struct_array_field.brainrot
leo-aa88, line 21: Happy-path coverage is real. It does not challenge the PR's own claims: pointer-element fields, rank mismatch, or padding after a short array toward a wider-aligned neighbor. Add those (or narrow the PR description so it stops advertising untested support).

leo-aa88 (commented):
# REQUEST CHANGES

Slice 2 of #206 (3e) mostly wires the right layers, but it ships a **split type contract** for `Array.base`, widens a known soft-fail into struct blobs, and leaves `ast_accept` one child short of the representation it just introduced. CI green does not exercise those contracts.

## 1. Type helpers disagree on whether `f.n[0]` is typed

**BLOCKING.**

`get_expression_type` requires `fld->is_array` and otherwise lies with `"Undefined array field"`. Same-PR siblings (`is_expression`, `get_expression_pointer_level`, `infer_runtime_expression_type_noeval`, and the semantic_analyzer mirrors) return the field type/level with **no** `is_array` check. The fail fixture then institutionalizes the false diagnostic. See inline comments.

## 2. `Array.base` exists; `ast_accept` never walks it

**MAJOR.**

Semantic analysis visits `base` under a comment claiming to mirror `visitor.c`. `visitor.c` still only walks indices — and is not in this PR. Traversal ownership is now undefined. See inline on the semantic analyzer site.

## 3. `calculate_array_offset` TEMPORARY FIX blast radius widened

**MAJOR.**

Factoring off `Variable *` is fine. Reusing the dimension-mismatch soft-fail for inline struct fields is not: under-ranked access silently picks the wrong element inside a shared blob. See inline.

## 4. Claimed contracts the tests do not challenge

**MAJOR.**

PR claims pointer element types. No pointer-element fixture. No rank-mismatch fail fixture for fields. No padding-after-short-array layout case despite #206's "layout is testable" rule. `maxxing(total_size)` only checks Brainrot's own accounting.

## 5. Incomplete `array.name` audit

**MAJOR** (maintainability).

Commit claims every `array.name` use was grepped. `get_struct_def_for_expression`'s `NODE_ARRAY_ACCESS` case still name-only. Duplicate `if (array.base)` branches are how the next slice regresses.

## 6. Access node marked `is_array = fld->is_array`

**MINOR.**

A `NODE_ARRAY_ACCESS` denotes an element; copying the field's `is_array` marks an element expression as an array.

# VERDICT

Deepest issue is **architecture + contract**, not a missing snprintf.

Do **not** merge until: (1) one shared array-field resolution rule for type/pointer/`is_expression` (honest fail fixture), (2) `visitor.c` walks `Array.base`, (3) struct-field path rejects rank mismatch hard, (4) a pointer-element field fixture exists if that remains in the PR claim.

---
Note: GitHub forbids `REQUEST_CHANGES` on one's own PR, so the event is `COMMENT`; the verdict above is still **REQUEST CHANGES**. Earlier `probe:` reviews were line-validation noise (inline probe comments deleted).

Issue #197, opened 2026-08-18
Support chained member access through a pointer-typed struct/union field (a.ptr.b)


leo-aa88:
## Summary

Follow-up to #190 (nested struct/union support). A struct/union field can
now be a pointer to another struct/union — e.g. the classic self-referential
linked-list shape:

```c
gang Node {
    rizz val;
    gang Node *next;
};
```

— and this is accepted at declaration time. But chaining `.` through such
a field (e.g. `n.next.val`) is explicitly rejected:

```c
skibidi main {
    gang Node n;
    n.val = 3;
    yapping("%d", n.next.val);
}
```
```
Error: Chained member access through a pointer-typed struct/union field is not supported at line ...
```

This is caught cleanly (in semantic analysis, before execution starts —
see `resolve_struct_access()`'s doc comment in ast.c and the
`semantic_error_nested_struct_pointer_chain.brainrot` test case), it's
just not implemented.

## Why this wasn't done in #190

Making `.` implicitly dereference a pointer field means deciding on and
implementing pointer-dereference semantics for struct member access,
which didn't exist anywhere in the language before #190 (struct pointer
*variables* have their own, more basic bug — see #196 — which is likely a
prerequisite: `resolve_struct_access()` needs "dereference before treating
as a blob" logic for a pointer-typed identifier before it can reasonably
be taught the same trick one level deeper for a pointer-typed *field*).

## Suggested direction

Once #196 establishes what a struct pointer variable's storage actually
means (an address into another variable's blob, not its own blob), extend
`resolve_struct_access()`'s `NODE_STRUCT_ACCESS`-recursion branch in ast.c:
when the resolved parent field has `pointer_level > 0`, dereference the
stored `uintptr_t` to get the next level's base address instead of
rejecting outright. `compute_struct_layout()`/`compute_union_layout()`
already size such fields correctly (`sizeof(uintptr_t)`) since #190, so
this should be purely an access-path change, not a layout change.

Would also need: an example/test exercising an actual linked structure
(build one node, point it at another, walk the chain), not just a
self-referential declaration.

PR #248, opened 2026-08-26
fix(struct): follow the pointer for pointer-typed struct/union variables (#196)

### LINKED ISSUE #196: Pointer-to-struct/union variables silently give wrong results instead of following the pointer

leo-aa88:
## Description

Part of the #206 roadmap epic (3d: pointer-chained member access). This PR covers #196 specifically — a pointer-typed struct/union **variable** used as the base of member access (`gang Foo *pp; pp.field`) now follows the pointer and reads/writes the real pointee, instead of being rejected.

`resolve_struct_access()`'s `NODE_IDENTIFIER` branch previously rejected this outright (a hardening fix added in #247 to stop a leak, which explicitly did not implement following the pointer). It now reads the variable's stored pointer value and uses it as the member-access base, with a clean "Null pointer dereference in struct member access" error instead of a crash when the pointer hasn't been assigned.

While wiring this up, I found and fixed a separate pre-existing bug this depended on: `interpreter_visit_declaration`'s `pointer_level > 0` branch read `data.op.right` as the initializer expression for a struct-tagged pointer declaration, but for that grammar production (`struct_or_union name_token declarator EQUALS expression`, lang.y) `data.op.right` actually holds a `NODE_STRUCT_DEF` type marker — the real initializer lives in `struct_init_expr`. `evaluate_expression_pointer()` doesn't handle `NODE_STRUCT_DEF`, so it silently returned 0, meaning `gang Foo *pp = &f;` never actually stored `&f` into `pp` at all. Fixed by rerouting to `struct_init_expr` specifically for that grammar path, while leaving `lit`-aliased struct pointer declarations (which use a different, generic type-declarator grammar path where `data.op.right` already holds the real initializer) untouched — confirmed via the existing `lit_struct_pointer_alias_array` regression, which briefly broke and is now green again.

Chaining `.` through a pointer-typed **field** partway through an access chain (`n.next.val`, #197) is intentionally out of scope for this PR — it needs its own fix in both `ast.c` and `semantic_analyzer.c` and will follow as a separate PR per the #206 plan.

## Related Issue

Fixes #196

## Type of Change

## Checklist


leo-aa88:
@cursor review

leo-aa88:
@cursor review

leo-aa88:
@cursor review

Issue #291, opened 2026-08-28
brainray: no way to draw a number — rl_draw_text only ever receives a literal


leo-aa88:
## Summary

`brainray` can draw text, but not a number. `rl_draw_text` takes a `rant`, and Brainrot has no string concatenation and no `sprintf`, so the only strings a program can ever hand it are literals.

That means a game can compute a score and has no way to show it. Every literal label works today — `SCORE`, `LVL`, `GAME OVER` — and the digits after them are simply unreachable.

## Proposal

Two wrappers, mirroring the existing `rl_draw_text` / `rl_measure_text` pair:

```
rl_draw_text_int(text, value, pad, x, y, size, r, g, b, a)
rl_measure_text_int(text, value, pad, size)   -> rizz width
```

Formatting is fixed at **one literal prefix followed by exactly one integer**. That is deliberately less general than exposing a format string: a Brainrot-supplied `"%s"` would make the host read an argument that isn't there, so the format never comes from user code.

`pad` is the minimum digit count, zero-padded, so a HUD doesn't jitter as a counter grows:

```c
rl_draw_text_int("SCORE ", score, 6, 20, 20, 28, 245, 200, 90, 255);   🚽 SCORE 000450
rl_draw_text_int("LVL ",   lvl,   0, 1064, 46, 20, 220, 220, 220, 255); 🚽 LVL 3
```

`rl_measure_text_int` exists so numeric text can be centred the same way a literal can, which the game-over screen needs.

## Why not a general format string

`TextFormat`-style variadics would be more flexible and strictly more dangerous: the format would be attacker-controlled from the interpreter's point of view, with no way to check it against the single `rizz` actually passed. The narrow version covers the HUD case — which is the case that exists — and can be widened later if a real need appears.

## Context

This is the first upstream ask from [tung-tung-sahur](https://github.com/Brainrotlang/tung-tung-sahur)'s design doc (§15.1, "B1"), where it blocks the entire numeric HUD. M0 shipped with labels and no digits because of it.

PR #292, opened 2026-08-28
brainray: add rl_draw_text_int and rl_measure_text_int

leo-aa88:
## Description

`brainray` could draw text but not a number. `rl_draw_text` takes a `rant`, and Brainrot has no string concatenation and no `sprintf`, so the only strings a program could ever hand it were literals — a game could compute a score and had no way to show it.

Adds two wrappers mirroring the existing `rl_draw_text` / `rl_measure_text` pair:

```
rl_draw_text_int(text, value, pad, x, y, size, r, g, b, a)
rl_measure_text_int(text, value, pad, size)   -> rizz width
```

```c
rl_draw_text_int("SCORE ", 450, 6, 40, 80, 28, 245, 200, 90, 255);   🚽 SCORE 000450
```

| `value` | `pad` | drawn |
| --- | --- | --- |
| `450` | `0` | `SCORE 450` |
| `450` | `6` | `SCORE 000450` |
| `-450` | `6` | `SCORE -00450` |
| `0` | `6` | `SCORE 000000` |

**Formatting is fixed at one literal prefix plus exactly one integer, not a format string.** A Brainrot-supplied `"%s"` would make the host read an argument that isn't there, and there's no way to validate a user-supplied format against the single `rizz` actually passed — so the format never comes from user code. Narrower than `TextFormat`, and it covers the case that actually exists.

`pad` is clamped to 32 so a wild value can't request an enormous allocation, and a negative value keeps its sign inside the padded field exactly as printf's `%0*d` does.

### On the allocation

These are the only wrappers besides `rl_init_window` that allocate. The allocation sits **outside** the LeakSanitizer brackets on purpose, per the existing policy in this file's header: raylib's own globals are disclaimed, brainray's own memory stays tracked. So a missed `free()` here is still reported as brainray's leak instead of vanishing into the graphics stack.

To make sure that's not just a claim, `test_brainray_windowed_run_is_leak_clean` now calls both functions every iteration, so a per-call leak accumulates. **Verified by deleting the `free()` and watching the test go red**, then restoring it.

### Docs guard

`docs/brainray.md` calls itself the single source of truth for the binding, and both `README.md` and `examples/raylib/README.md` defer to it rather than repeating the function list. That only holds if the table keeps up — a wrapper added without a row is undiscoverable. New test asserts every `STDROT_EXPORT_SIG` name appears in the doc, checked by name rather than by parsing the table's columns, so reformatting can't break it or hide a gap. Also verified to fail by deleting a row.

## Related Issue

Related to #291

## Type of Change

- [x] Documentation update
- [ ] Performance improvement
- [ ] Refactor

## Checklist

`make test` 415 passed. `make valgrind` 400 test cases, 0 errors, exit 0. `make format-check` clean. `make cppcheck` **not run** — cppcheck isn't installed on this machine, so CI's `static-analysis` job is the first to see it.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -241,11 +241,17 @@ jobs:
       - name: Build native module test libraries
         run: make nativemodules
 
+      # Collect every test module in tests/, not just test_brainrot.py. Naming
+      # one file meant test_docs_consistency.py never ran on the merge gate,
+      # even though `make test` (plain `pytest -v` from the repo root) has
+      # always collected it -- so CI was strictly narrower than the command
+      # contributors are told to run, and the docs guards it holds could go
+      # stale while CI stayed green.
       - name: Run Pytest
         run: |
           source .venv/bin/activate
           export STDROT_LIB_PATH="$GITHUB_WORKSPACE/tests/libstdrot.so"
-          pytest -v test_brainrot.py
+          pytest -v
         working-directory: tests
 
       - name: Run Valgrind on all .brainrot tests
--- a/brainray/raylib.c
+++ b/brainray/raylib.c
@@ -55,6 +55,7 @@
  */
 #include "stdrot_api.h"
 #include <raylib.h>
+#include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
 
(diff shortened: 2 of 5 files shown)


leo-aa88:
Thanks — all four findings were correct. Addressed:

**1. The formatting contract is documented, not tested (MAJOR).** Agreed; the leak smoke only ever proved ownership. Added `test_brainray_text_int_formatting_matches_docs`, a separate test that measures `rl_measure_text_int` against `rl_measure_text` of the exact literals in the table — every row, not just the easy one.

I verified it by mutation rather than by asserting it works:

| mutation | result |
| --- | --- |
| `"%s%0*d"` → `"%s%d"` (ignore `pad`) | fails the pad-6 row |
| ignore `value`, emit prefix only | fails the pad-6 row |
| `if (value < 0) pad++` — i.e. digit-count semantics | **fails specifically on the `-450` row** |

That third one is the case your second comment is about, and it's why the negative row earns its place.

I also added the distinctness assertion you implied — padded and unpadded must measure differently. That makes "pad is dead code" a definite catch rather than a lucky one, and it fails loudly if `MeasureText` ever degenerates to 0 for everything, which would otherwise make every equality trivially true.

On what it can't do: you suggested it would catch concat in the wrong order. It won't — text width is the sum of glyph widths, so `"SCORE 450"` and `"450SCORE "` measure identically. The docstring says so explicitly rather than leaving the impression the contract is fully pinned.

**2. "digits" vs field width.** Correct, and the split was in both the prose and the code comment while the table, the `%0*d` sentence and the C were all right. Fixed the words, not the format string — field width is what a HUD wants, since `000450` and `-00450` occupy the same column.

**3. The docs guard is weaker than advertised, and not on the merge gate.** Both true. It now requires an actual table row (anchored to the leading pipe), and additionally fails on the reverse — a row for a function brainray no longer exports. Verified by deleting the row while three prose mentions of the name remain; the old check passed that, the new one fails it.

For the gate: CI ran `pytest -v test_brainrot.py`, so `test_docs_consistency.py` never ran there even though `make test` has always collected it — CI was strictly narrower than the command contributors are told to run, and `test_make_help_lists_developer_targets` sat in the same blind spot. CI now collects the whole `tests/` directory, which closes it for both.

**4. Nit — ignored `snprintf` return.** Now explicitly `(void)`-discarded, with a note on why it can't truncate.

**One thing I did not do.** The new format test is still `skipif`-ed in headless CI, because `MeasureText` needs `InitWindow` to have loaded the default font. Putting raylib behind `xvfb` on the runner would genuinely close your "CI will not save you" point for every brainray runtime test, not just this one — but it's a CI change with its own flakiness surface (software GL on a runner) and I don't think it belongs in this PR. Happy to open it as a follow-up.

Local: `make test` 416 passed, `make valgrind` 400 cases / 0 errors / exit 0, `make format-check` clean.

/cursor-review

PR #293, opened 2026-08-28
brainray: add rl_draw_texture_rec

leo-aa88:
## Description

`rl_draw_texture` blits a whole image, so an animation meant one file per frame and a sprite could not face the other direction. `DrawTextureRec` takes a source rectangle, which is what makes atlases, animation strips and tiled parallax possible from Brainrot at all. This is **B2** in [tung-tung-sahur](https://github.com/Brainrotlang/tung-tung-sahur)'s design doc, and the last thing blocking the rest of its M1.

```
rl_draw_texture_rec(handle, sx, sy, sw, sh, x, y, r, g, b, a)
```

```c
🚽 frame `i` of a 64x64 strip
rl_draw_texture_rec(atlas, i * 64.0, 0.0, 64.0, 64.0, px, py, 255, 255, 255, 255);
```

Two aggregates come apart here rather than one: `rec` is a `Rectangle` (four floats), `position` is a `Vector2` (two floats). The float position differs from `rl_draw_texture`'s int `x, y` — that mirrors raylib rather than contradicting it (`DrawTexture` takes ints, `DrawTextureRec` takes a `Vector2`), and sub-pixel placement is what a scrolling background wants anyway.

**Negative `sw` mirrors horizontally, negative `sh` vertically** — raylib's own idiom, since the sign of the source rectangle decides texture-coordinate order. Until `DrawTexturePro`'s rotation is exposed that's the only way to flip a sprite, so it's documented rather than left looking like a bug.

## Verified by eye

A four-frame atlas, sliced and flipped. Top row: frames 0–3 out of one 256×64 texture. Bottom: the same frame normal then with `sw = -64` (arrow flips, top bar stays), `sh = -64` (bar moves to the bottom, arrow doesn't), a 32px half-slice, a tint, and two bad handles that correctly draw nothing.

Both mirror claims in the docs are confirmed by that image, not assumed.

## What the test does and does not establish

The last review was right to press on this, so I'll state it up front rather than have it found.

**Verified** — the assertion fails if the behaviour changes:
- a generated PNG loads to a non-negative handle;
- **a missing file loads to `-1`**, the documented sentinel. This had no test before; flipping it to `0` makes the new test fail (confirmed);
- load → five draw forms → three guard cases → unload → close exits 0 under ASan with no `LeakSanitizer` report.

**Not verified, and not claimed:**
- **Which pixels landed.** brainray can't read a texture or the framebuffer back, so "the source rectangle was respected" isn't machine-checkable — an implementation ignoring `rec` and blitting the whole texture would pass. Hence the visual check above.
- **That the handle guards work.** `SO_CFLAGS` is `-fPIC -shared` with no sanitizers, as for every shared object here including `libstdrot.so`, so an out-of-bounds read of `g_textures[]` *inside the module* is invisible to ASan. **I deleted the bounds check and the test still passed.** The guards match `rl_draw_texture` and keep a bogus GPU id away from raylib; they are reviewed, not tested.

Closing either gap needs something outside this PR: a pixel-readback wrapper (`LoadImageFromTexture` / `TakeScreenshot`) for the first, sanitizer-instrumented shared objects for the second. I didn't do the latter here because unsanitized `.so`s look like a deliberate repo-wide convention, not an oversight — happy to raise it separately if it isn't.

The test writes its own 8×4 PNG from `zlib` + `struct`, so the fixture adds no imaging dependency and nothing binary is checked in.

## Related Issue

Follows #291 (B1). No issue filed for B2 — say the word and I'll open one.

## Type of Change

- [x] Documentation update
- [ ] Performance improvement
- [ ] Refactor

## Checklist

`make test` 417 passed. `make valgrind` 400 cases, 0 errors, exit 0. `make format-check` clean. `make cppcheck` not run — not installed here; CI's `static-analysis` job is the first to see it.

### DIFF
--- a/brainray/raylib.c
+++ b/brainray/raylib.c
@@ -443,6 +443,45 @@ static StdrotValue br_draw_texture(StdrotValue *args, int argc)
     return (StdrotValue){.type = STDROT_NONE};
 }
 
+/* DrawTextureRec: one sub-rectangle of a texture, which is what makes sprite
+ * atlases and animation frames possible from Brainrot at all.
+ *
+ * Two aggregates come apart here rather than one. raylib's `rec` is a
+ * Rectangle (four floats) and `position` is a Vector2 (two floats), so the
+ * source box arrives as sx, sy, sw, sh and the destination corner as x, y --
+ * six `chad` arguments, then the usual four-int Color tail.
+ *
+ * Note the float position, which differs from rl_draw_texture()'s int x, y.
+ * That is not an inconsistency: it mirrors raylib, where DrawTexture() takes
+ * ints and DrawTextureRec() takes a Vector2. Sub-pixel placement is what a
+ * scrolling background wants anyway.
+ *
+ * A negative `sw` mirrors the sprite horizontally and a negative `sh`
+ * vertically -- raylib's own idiom, since the source rectangle's sign decides
+ * the texture coordinate order. That is the only way to face a sprite the
+ * other direction until DrawTexturePro's rotation is exposed, so it is worth
+ * knowing about rather than looking like a bug.
+ *
+ * The rectangle is handed to raylib unchanged: brainray does not clamp it to
+ * the texture, so sampling outside the image is raylib's business and not a
+ * guarantee this wrapper makes. Handle validation matches rl_draw_texture --
+ * an out-of-range or already-unloaded handle draws nothing rather than
+ * feeding raylib a stale GPU id. */
+static StdrotValue br_draw_texture_rec(StdrotValue *args, int argc)
+{
+    (void)argc;
+    int handle = args[0].val.i;
+    if (handle >= 0 && handle < BRAINRAY_MAX_TEXTURES && g_texture_used[handle])
+    {
+        Rectangle rec = {(float)args[1].val.f, (float)args[2].val.f,
+                         (float)args[3].val.f, (float)args[4].val.f};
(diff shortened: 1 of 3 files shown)


leo-aa88:
Fixed the census nit — it was correct, and the "next wrapper will copy this sentence" framing is exactly why it was worth doing rather than merging past. The paragraph now lists all three handle consumers and says outright that it *is* the complete census, so the next one doesn't inherit the same staleness.

Two notes on the rest of the review, since I agree with it and don't want the agreement to be silent:

- **The coverage ceiling on `rec`.** Confirmed — an implementation that ignored the source rectangle and called `DrawTexture` would pass. That's why the flips and the sub-rect were checked against a real atlas by eye and the image is in the PR body. A pixel-readback wrapper is what would turn that into a test, and it's a separate ask.
- **`make format-check` / `make valgrind` / the windowed test in your environment.** All three ran here: `format-check` clean, `valgrind` 400 cases with 0 errors and exit 0, and the windowed test passes against raylib 6.0.0 under WSLg. `make cppcheck` is the one I couldn't run locally, and CI's `static-analysis` job covers it — green on this HEAD.

Thanks for tracing declaration → `STDROT_EXPORT_SIG` → `coerce_numeric` → the `Rectangle`/`Vector2` reads, and for independently confirming the negative-`sw`/`sh` flip against `DrawTexturePro`'s source-sign handling rather than taking the docs' word for it.

/cursor-review

PR #294, opened 2026-08-28
fix: check call arguments against parameters in source order

### LINKED ISSUE #288: Struct-pointer parameters are type-checked in reverse order; the swapped call is accepted and corrupts memory
## Summary

For any function with **struct-pointer parameters of two or more different types**, the semantic analyser checks the arguments against the **reversed** parameter list. The correct call is rejected; the swapped call is accepted, and because runtime argument binding is positionally *correct*, the body then writes one struct's field offsets into a different struct.

Found while writing [tung-tung-sahur](https://github.com/Brainrotlang/tung-tung-sahur), where it blocks the natural `step(gang Entity *e, gang World *w)` shape entirely.

## Reproduction

**The correct call is rejected.** Note the error text: argument 1 (an `Enemy*`) is checked against `World`, and argument 2 (a `World*`) against `Enemy`.

```c
gang World { chad dt; };
gang Enemy { chad x; };

skibidi step(gang Enemy *e, gang World *w) { e.x = e.x - w.dt; }

skibidi main {
    gang World wd; wd.dt = 0.5;
    gang Enemy en; en.x = 10.0;
    step(&en, &wd);
    yapping("en.x = %.1f (expected 9.5)", en.x);
    bussin 0;
}
```

```
Error: 'step' argument 2: type mismatch: expected pointer to struct/union 'Enemy', got pointer to 'World' at line 9
Error: 'step' argument 1: type mismatch: expected pointer to struct/union 'World', got pointer to 'Enemy' at line 9
```

**The swapped call is accepted, and corrupts memory.** This is the dangerous half — a user who "fixes" the error above by reordering the arguments gets a program that compiles and silently writes through the wrong type:

```c
    step(&wd, &en);   🚽 passes the checker
    yapping("en.x = %.1f  wd.dt = %.1f", en.x, wd.dt);
```

```
en.x = 10.0  wd.dt = -9.5
```

`en.x` is untouched and `wd.dt` — a `World` field — has been written with the result of an `Enemy` field computation.

## Scope

- Reversal applies to the **whole parameter list**, not just the struct ones: `step(gang E *e, rizz n)` reports *argument 2: expected pointer to struct 'E'*.
- With **three** struct-pointer params `f(gang A *x, gang B *y, gang C *z)`, argument 1 is checked against `C` and argument 3 against `A` — a straight reversal.
- **Multiple params of the same struct type are unaffected**, because the reversal is invisible. `copyfrom(gang E *dst, gang E *src)` checks *and* runs correctly.
- A **single** struct-pointer param is unaffected.

So the working subset today is "at most one struct type across a function's parameters", which is what tung-tung-sahur is currently written against.

## Expected

`step(&en, &wd)` type-checks and prints `en.x = 9.5`; `step(&wd, &en)` is the one rejected.

## Environment

`main` @ 3ce8a3a, default `make` build (`-fsanitize=address,undefined`).

leo-aa88:
## Description

`func->parameters` is stored in **reverse source order**. Every `param_list COMMA ...` rule in `lang.y` hangs the accumulated list off the *new* node's `next`, so `f(a, b, c)` is kept as `c -> b -> a`. `enter_function_scope()` knows this and snapshots a call-order view before binding — its comment says so outright:

> *"Snapshot the parameters in call order into a local array. **The parser stores them reversed**, so reverse the shared list in place, copy the node pointers out, then restore it IMMEDIATELY..."* — `ast.c`

The semantic analyzer's argument check never got that memo. It walked the declaration list forward against a forward argument list, so **argument 1 was compared against the last parameter**.

### Why it survived

Invisible whenever a function's parameters share one type — the reversal cancels out. `copyfrom(gang E *dst, gang E *src)` checks and runs correctly, which is most of the existing test corpus.

### What it cost where they differ

```c
gang World { chad dt; };
gang Enemy { chad x; };
skibidi step(gang Enemy *e, gang World *w) { e.x = e.x - w.dt; }
```

**The correct call was rejected**, with a confident and precisely inverted diagnostic:

```
Error: 'step' argument 1: type mismatch: expected pointer to struct/union 'World', got pointer to 'Enemy'
```

**The swapped call was accepted.** And since the runtime binds positionally and *correctly*, a user who trusted that error and reordered the arguments got a program that type-checked while running the body with `e` aimed at a `World` — writing an `Enemy` field's offset through it. `wd.dt` came out as `-9.5`. Silent cross-type memory corruption, reached by following the compiler's advice.

That second half is what makes this worse than a false positive: the diagnostic actively steers you into the corruption.

## The fix, and the one I didn't make

Fixed in the analyzer: snapshot the same call-order view the runtime already builds. Reading only — unlike `ast.c`'s version it never mutates the shared list, so it needs no restore and has no reentrancy window.

**I did not un-reverse the parser's list**, which is the tidier-looking fix. `enter_function_scope()` and every other reader of `func->parameters` depend on the current order, none of it is user-visible, and trading a two-line analyzer fix for a change that touches argument binding is a bad deal in a bug-fix PR.

I checked the only other place that walks a parameter list — the function-definition scope setup in this same file. It adds each parameter as a symbol under its own name and type, so order can't affect it. This was the only site.

## Verification

Both fixtures were confirmed to **fail without the fix** (stash, rebuild, run) and pass with it:

- `struct_pointer_param_order` — the shapes where reversal is observable: two different struct-pointer types, a struct pointer beside a scalar in *either* order, and three distinct struct pointers.
- `struct_pointer_param_order_swapped_fail` — the swapped call must be rejected, and must name the right argument.

Genuine tag mismatches are still caught, now with the correct argument index.

### What this unblocks

[tung-tung-sahur](https://github.com/Brainrotlang/tung-tung-sahur)'s entire frame loop lives in one `skibidi main` because of this bug (its DESIGN.md §14.1). This shape now works:

```c
skibidi ent_step(gang Ent *e, gang World *w) {
    edgy (e.alive) {
        e.x = e.x - w.speed * w.dt;
        edgy (e.x < 0.0 - e.w) { e.alive = L; w.score = w.score + 10; }
    }
}
```

## Related Issue

Fixes #288

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement
- [ ] Refactor

Marked non-breaking, with one caveat worth a reviewer's eye: code that was written *around* this bug — parameters deliberately declared backwards so the old check passed — would now be correctly rejected. I'd argue that code was already broken at runtime, but it's the one way this could surface as a break.

## Checklist

`make test` 419 passed. `make valgrind` 402 cases, 0 errors, exit 0. `make format-check` clean. `make cppcheck` not run — not installed here; CI's `static-analysis` job covers it.

### DIFF
--- a/semantic_analyzer.c
+++ b/semantic_analyzer.c
@@ -2268,11 +2268,44 @@ void *semantic_visit_function_call(Visitor *self, ASTNode *node)
                generic per-argument recursion (this function's caller,
                semantic_analyze_with_scope_tracking()'s NODE_FUNC_CALL
                case) walks into it. */
-            Parameter *param = func->parameters;
+            /* func->parameters is stored in REVERSE source order: every
+               `param_list COMMA ...` rule in lang.y hangs the accumulated
+               list off the NEW node's `next`, so `f(a, b, c)` is kept as
+               c -> b -> a. enter_function_scope() (ast.c) knows this and
+               snapshots a call-order view before it binds anything; this
+               loop did not, and walked the declaration list backwards
+               against a forward argument list -- checking argument 1
+               against the LAST parameter.
+
+               Invisible whenever a function's parameters share one type,
+               which is why it survived. But for
+               `skibidi step(gang Enemy *e, gang World *w)` it rejected the
+               correct call AND accepted the swapped one, and since the
+               runtime binds positionally (correctly), a user who trusted
+               the diagnostic and reordered the arguments got a program
+               that type-checked while writing one struct's field offsets
+               through the other struct's pointer. Silent cross-type
+               corruption out of a "helpful" error message.
+
+               Snapshot the same call-order view here. Unlike ast.c's
+               version this never mutates the shared list -- it only reads
+               it -- so there is no reentrancy window needing a restore. */
+            Parameter *ordered[MAX_ARGUMENTS];
+            int param_count = 0;
+            for (Parameter *p = func->parameters;
+                 p && param_count < MAX_ARGUMENTS; p = p->next)
+            {
+                ordered[param_count++] = p;
+            }
+
             ArgumentList *arg = node->data.func_call.arguments;
             int arg_index = 0;
-            while (param && arg)
+            while (arg && arg_index < param_count)
             {
+                /* ordered[] holds the stored (reversed) order, so the
+                   argument at 0-based `arg_index` pairs with the parameter
+                   that many places from the END. */
+                Parameter *param = ordered[param_count - 1 - arg_index];
                 arg_index++;
                 if (arg->expr)
                 {
@@ -2347,7 +2380,6 @@ void *semantic_visit_function_call(Visitor *self, ASTNode *node)
                         }
                     }
                 }
-                param = param->next;
                 arg = arg->next;
             }
         }
--- a/semantic_analyzer.c
+++ b/semantic_analyzer.c
@@ -2287,9 +2287,17 @@ void *semantic_visit_function_call(Visitor *self, ASTNode *node)
                through the other struct's pointer. Silent cross-type
                corruption out of a "helpful" error message.
 
-               Snapshot the same call-order view here. Unlike ast.c's
-               version this never mutates the shared list -- it only reads
-               it -- so there is no reentrancy window needing a restore. */
+               Pair each argument with its source-order parameter here.
+               Note this array is NOT laid out like ast.c's `ordered[]`,
+               despite the shared name and purpose: that one reverses the
+               shared list first, so its ordered[0] is the FIRST source
+               parameter and it can index straight through. This one copies
+               the stored list as-is, so ordered[0] is the LAST source
+               parameter and the index arithmetic below is what compensates.
+               Reading rather than reversing is deliberate -- it never
+               mutates the shared list, so unlike the runtime's version
+               there is no window in which a nested call could observe it
+               reversed, and nothing to restore. */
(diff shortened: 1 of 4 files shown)


leo-aa88:
Fixed the `ordered[]` comment — you're right, and it was the one genuinely misleading line in the patch. "Snapshot the same call-order view" reads as though the array matches `enter_function_scope()`'s, when it's laid out the opposite way: `ast.c` reverses first so its `ordered[0]` is the *first* source parameter and it indexes straight through, while this one copies the stored list untouched so `ordered[0]` is the *last* and the arithmetic compensates. Only the pairing is shared, not the layout. Someone reaching for the runtime's pattern after reading my sentence would have indexed this array backwards — which is a nice recursion of the exact bug being fixed.

On the debt you flagged: agreed, and I'd rather it be tracked than live in a PR comment. The reversed parameter list is now an implicit invariant with **two different compensation strategies** in two files, and this file already admits it only type-checks struct-pointer parameters at all — so the next person to close that gap has to rediscover the prepend for themselves. I'll open an issue for un-reversing it at the source, noting that the real work is `enter_function_scope()`'s reentrancy-sensitive bind path (the one PR #254 already had to fix once) rather than the grammar rules.

Also worth recording, since you couldn't run them: `make test` 419 passed, `make valgrind` 402 cases with 0 errors and exit 0, `make format-check` clean. And thank you for independently deriving the prepend from `create_parameter_ex(new, …, $1)` versus `create_argument_list`'s append, and for confirming the `get_function()` timing — that the `Function*` is registered at parse time is load-bearing for this loop and I hadn't spelled it out.

/cursor-review

PR #296, opened 2026-08-28
fix: implement logical NOT, which the lexer had been discarding

### LINKED ISSUE #289: Logical NOT (!) on a cap returns its operand unchanged, silently inverting branch logic
## Summary

The logical NOT operator `!` applied to a `cap` returns its operand unchanged. `!L` evaluates to `L`, `!W` to `W`, and `edgy (!a)` takes the wrong branch. It fails silently — there is no error, the program just does the opposite of what it says.

## Reproduction

```c
skibidi main {
    cap a = L;
    cap b = !a;
    yapping("!L  = %b   (expected W)", b);
    cap c = !W;
    yapping("!W  = %b   (expected L)", c);
    edgy (!a) { yapping("edgy (!a): taken   (correct)"); }
    amogus   { yapping("edgy (!a): NOT taken   (wrong)"); }
    bussin 0;
}
```

**Actual:**

```
!L  = L   (expected W)
!W  = W   (expected L)
edgy (!a): NOT taken   (wrong)
```

## Expected

```
!L  = W
!W  = L
edgy (!a): taken
```

## Notes

`!` is listed as a supported operator in `docs/the-brainrot-programming-language.md` §7.2, with the hedge "depending on grammar rules".

The workaround is to write `a == L` everywhere, which is what [tung-tung-sahur](https://github.com/Brainrotlang/tung-tung-sahur) does — but the failure mode is the concerning part. A wrong branch with no diagnostic is much worse than a parse error would have been, so if `!` on `cap` isn't going to work, rejecting it outright would be an improvement over the current behaviour.

## Environment

`main` @ 3ce8a3a, default `make` build.

leo-aa88:
## Description

`!` had **no token**. `lang.l`'s final catch-all rule —

```
.                { /* Ignore unrecognized characters */ }
```

— swallowed it, so `!a` lexed as plain `a` and every negation evaluated to its own operand. `!L` was `L`, `!W` was `W`, and `edgy (!done)` took the wrong branch. No parse error, no warning: the operator evaporated between the lexer and everything downstream, while `docs/the-brainrot-programming-language.md` listed it as supported — hedged with *"depending on grammar rules"*, which turns out to have been the honest part.

Adds the token, a unary rule at unary minus's precedence (where C puts logical NOT, so `!a == b` is `(!a) == b`), `OP_NOT`, and evaluation. **No new grammar conflicts** — `bison -Wcounterexamples` is clean, and `!=` still lexes as one token.

## What `!` yields, and the two places that had to agree

`!x` is a truth value whatever `x` was, so unlike `-x` it does not inherit the operand's type. That had to be taught to **two independent inference paths** — `ast.c`'s `get_expression_type()` and `semantic_analyzer.c`'s `infer_expression_type()` — or `cap b = !n;` is rejected as *"expected bool, got int"* by an analyzer disagreeing with an interpreter that would have produced a perfectly good `cap`.

It also had to be taught to **both pointer-level queries**. `!p` is a truth value; reporting it as pointer level 1 made callers evaluate the result as an address (*"Invalid pointer expression"*) even though the computed answer was right.

## Judging the operand in its own type

This is the part I'd most like eyes on. `!` cannot reuse the caller's context type the way `-` can, for two separate reasons — both reachable the moment `!` started working:

- Funnelling a `chad` through the int path makes **`!0.5` truncate to `!0` and answer true**, when 0.5 is as true as any other non-zero value.
- Reading a `rizz` back through a `bool *` is a type-punned load **UBSan rejects outright**: `load of value 5, which is not a valid value for type '_Bool'`. I hit this while testing, not in review.

So `expression_is_truthy()` dispatches on the operand's static type and each typed evaluator intercepts `OP_NOT` before coercing. A pointer operand compares against NULL, which makes `!p` the null check it looks like.

**All five interception paths are live, not defensive.** I checked rather than assumed:

| reaches | via |
| --- | --- |
| int | `arr[!z]`, `ohio (!a)`, a `rizz` parameter |
| chad | a `chad` parameter |
| gigachad | a `gigachad` parameter |
| smol | a `smol` parameter |
| cap | every ordinary boolean context |

## One deliberate non-C behaviour

`rizz k = !n;` is a **type error**. `cap` and `rizz` don't silently convert in either direction in this language — `rizz k = W;` and `cap b = 1;` are already rejected, as is arithmetic on a `cap`. So this is consistency with the existing design, not a gap I left.

Only **untyped integer contexts** coerce to C's 0/1: an array index, an `ohio` selector, an `edgy`/`goon` condition. Declarations reject a cap, and so do native `rizz` parameters (`ragequit(!z)` is *expected int, got bool*). User-defined parameters aren't type-checked at all — the three `takes_*(!z)` calls in the fixture are there because the runtime evaluates each argument in the *parameter's* type, which is the only route to the chad/gigachad/smol intercepts, not because they're a checked conversion.

## Same class, not fixed here

**`~` is still discarded by that same catch-all** — `~5` silently evaluates to `5`, exactly as `!L` used to. The catch-all is the root cause of both: it turns any unsupported operator into silence rather than a diagnostic. I'll file that separately rather than widen this PR; making the lexer reject unknown characters is a behaviour change that deserves its own discussion.

## Related Issue

Fixes #289

## Type of Change

- [x] Documentation update
- [ ] Performance improvement
- [ ] Refactor

Both boxes, honestly: it's filed as a bug because the docs promised the operator, but the work is a full lexer→parser→AST→analyzer→interpreter slice. Code that *relied* on `!` being silently dropped would now change behaviour — but such code was already doing the opposite of what it said.

## Checklist

`test_cases/logical_not.brainrot` covers every operand type, the pointer null check, precedence including `!!a`, the three numeric-coercion contexts, `!=` still lexing as one token, and the `goon (!done)` loop shape the bug actually broke.

`make test` 420 passed. `make valgrind` 403 cases, 0 errors, exit 0. `make format-check` clean. `make cppcheck` not run — not installed here; CI's `static-analysis` job covers it.


leo-aa88:
The blocking finding was right, and I reproduced it before touching anything:

```c
smol arr[2] = {0, 1};
yapping("%b", !arr[0]);   /* printed L; W is correct */
```

Four bytes out of a two-byte slot, `0x00010000`, false for a zero. The pointer-deref form `!(*sp)` was wrong the same way. And your read of *why the fixture passed* was exactly right — scalar `smol s = 3` sits in zeroed union storage, so `*(int *)&svalue` happens to equal 3. A coincidental pass of the representation the helper exists to stop using.

`VAR_SHORT` now dispatches to `evaluate_expression_short()`. `yap` deliberately stays on the int path, since `evaluate_expression_int()` special-cases `VAR_CHAR` in its own array path and reads its byte correctly — the fixture now covers that too, rather than me asserting it.

**Dead arm removed.** You were right that all five evaluators return before reaching it, and right that its comment contradicted the type system. One implementation now.

**Fixing your finding surfaced a crash you hadn't seen — and it was mine.** Typing `!x` as `VAR_BOOL` removes the protection a struct got for free: `yapping("%b", -s)` is refused as *"struct has no supported native ABI representation"*, but `!s` looks like an ordinary cap, sails through, and reaches an int evaluator with no int to read. Null load, ASan `DEADLYSIGNAL`. `!` now requires a scalar or pointer operand; a `rant` is refused for the quiet version of the same problem.

Worth recording how I nearly shipped that check broken: I put it in `semantic_analyze_node()` first and the struct still crashed, because the expression travels through `semantic_analyze_with_scope_tracking()`. Two parallel unary implementations mean a check in one of them is a check in neither. It's a helper called from both now.

**On the tests and docs — both fair.** `takes_int(!z)` was never a checked conversion; the analyzer doesn't type-check user-defined call arguments at all. The fixture's comment now says so and explains why those calls are still there: the runtime evaluates each argument in the *parameter's* type, which is the only route to the chad/gigachad/smol intercepts. Added fail fixtures for the struct crash, the rant, and the `rizz k = !n;` rejection — including the packed-smol cases that would have failed today.

The docs sentence was wrong in three of four cases, as you counted. Native `rizz` parameters reject a cap, declarations reject it, user-defined parameters aren't checked; only untyped integer contexts convert. Rewritten to say that.

`make test` 423 passed, `make valgrind` 406 cases / 0 errors / exit 0, `make format-check` clean.

Thanks for this one — "the tests confirm the implementation, they do not challenge the contract" is the criticism I'd least like to be right, and it was.

/cursor-review

leo-aa88:
Closed all three leftovers rather than leaving them for a follow-up, since one of them was a false claim of mine.

**The yap coverage didn't exist.** My previous commit message said the fixture "now covers" yap on the int path. It did not — there was no `yap` operand anywhere in it, only the `yapping()` calls that a grep for "yap" happens to match. That is exactly the habit your first review caught, so there is now a real `yap` array with a zero and a non-zero byte, verifying that `evaluate_expression_int()`'s `VAR_CHAR` special case reads the single byte correctly and that `VAR_CHAR` needs no dispatch arm of its own. The packed-smol bug is what one unverified "this one is fine" costs, and I'd just made the same move again.

**The null-pointer branch is unreachable, not untested.** Brainrot has no null pointer literal — `rizz *p = 0;`, a bare `rizz *p;` and `NULL` are all rejected before `!` ever sees them. So the single `&n` case is the whole reachable surface. The fixture says so now, rather than looking like half a test.

**PR body corrected** to match the language reference — it now names the actual split (untyped integer contexts coerce; declarations and native `rizz` parameters reject; user-defined parameters aren't checked) and explains why the `takes_*(!z)` calls are in the fixture at all.

`make test` 423, `make format-check` clean; valgrind unchanged from the last run since this commit only touches a fixture and its golden.

/cursor-review

PR #299, opened 2026-08-28
fix: convert rather than reinterpret when reading a float as an integer

### LINKED ISSUE #290: Assigning a bare chad variable to a rizz reinterprets the float's bits instead of converting
## Summary

Assigning a **bare `chad` variable** to a `rizz` reinterprets the float's bit pattern instead of converting it. `17.5` becomes `1099694080` — the IEEE-754 bits of `17.5f` read as an int.

The trigger is specifically a bare variable (or struct field) on the right-hand side. Any *arithmetic expression* or function call in that position converts correctly, which makes the bug easy to miss: `g * 1.0` works, `g` does not.

Separately, `yapping("%d", someChad)` prints nothing at all.

## Reproduction

```c
skibidi main {
    chad g = 17.5;
    rizz bare = g;
    rizz expr = g * 1.0;
    yapping("bare variable : %d   (expected 17)", bare);
    yapping("expression    : %d   (expected 17)", expr);
    yappin("printf %%d of a chad: [");
    yappin("%d", g);
    yapping("]   (expected [17])");
    bussin 0;
}
```

**Actual:**

```
bare variable : 1099694080   (expected 17)
expression    : 17   (expected 17)
printf %d of a chad: []   (expected [17])
```

`1099694080` is `0x418C0000`, which is `17.5f`.

## Scope

- Both declarations (`rizz k = g;`) and plain assignments (`k = g;`) are affected.
- A bare **struct field** (`rizz k = p.x;`) is affected the same way.
- A `chad`-returning **function call** (`rizz k = f();`) converts correctly.
- Passing a `chad` to an int-typed **native parameter** converts correctly — the ABI coercion path is fine (see `test_cases/native_call_numeric_coercion.brainrot`).

So the broken path appears to be narrowly the "bare rvalue of float type into an int lvalue" case, while every path that goes through an expression node or the native ABI is right.

## Expected

`rizz k = g;` truncates toward zero, giving `17`, consistent with C and with what `g * 1.0` already does. `yapping("%d", g)` should either print the truncated value or be a diagnosed error, not print nothing.

## Notes

[tung-tung-sahur](https://github.com/Brainrotlang/tung-tung-sahur) routes every such conversion through a one-line `truncf()` helper purely to contain this, so the workaround is deletable in one commit once this lands.

## Environment

`main` @ 3ce8a3a, default `make` build.

leo-aa88:
## Description

`chad g = 17.5; rizz k = g;` produced **1099694080** — `0x418C0000`, which is `17.5f`. `smol s = g;` produced **0**.

`handle_identifier(node, err, 0)` hands back a pointer to the variable's *own* storage, and both integer evaluators cast that pointer to the type they wanted. For a `chad` variable that's a reinterpretation of the float's bit pattern rather than a conversion; for a `smol` target it reads the wrong two bytes.

Promote mode 1 already converts every numeric type correctly, so both sites route through it and narrow. Every numeric member of the `Value` union is exactly representable as a `double`, and all three callers already reject pointers before the read — which matters, because mode 1 refuses one outright.

**This is why an arithmetic right-hand side always worked.** `g * 1.0` goes through a real conversion path, so it gave `17` while a bare `g` did not. That asymmetry is what made the bug easy to live with instead of notice.

## Wider than the issue said

I checked the whole matrix before fixing. `chad → smol` was broken too, not just `chad → rizz`, and the reverse direction was fine:

| conversion | before | after |
|---|---|---|
| `chad` → `rizz` | `1099694080` | `17` |
| `chad` → `smol` | `0` | `17` |
| `gigachad` → `rizz` | bit pattern | `42` |
| `chad` → `rizz`, negative | bit pattern | `-17` (toward zero) |
| `rizz` → `chad` | `7.00` ✓ | `7.00` ✓ |

## The second half of #290 was a different mechanism

`yapping("%d", someChad)` printing nothing is **not** C's undefined behaviour, which is what I'd assumed before looking. `process_yapping_format()` *is* specifier-aware — but its integer branch handled only INT/SHORT/BOOL and its float branch only FLOAT/DOUBLE, so anything else matched no branch and emitted an empty string with no diagnostic.

Four silent holes, now all closed:

```
before                  after
%d <- chad     []       [17]
%d <- gigachad []       [99]
%f <- rizz     []       [42.000000]
%f <- smol     []       [7.000000]
```

I fixed both directions rather than only the one the issue named, since they're the same incomplete dispatch.

## One thing I deliberately left

The **bool** read site has the identical shape — `*(bool *)` of storage that may be int-width — and it's how I first hit this class of bug (a UBSan `load of value 5, which is not a valid value for type '_Bool'` while implementing `!`). I did not change it. It's currently unreachable for a non-bool variable: the analyzer rejects `rizz` in a `cap` context, and `!` routes through `expression_is_truthy()`'s own type dispatch. Changing it risks regressions for no gain in this PR. Worth its own look if a path to it ever appears.

## Verification

The fixture was confirmed to **fail without the fix** (stash, rebuild, run). It also pins the reverse direction — `rizz` → `chad`, never broken — so a later change can't trade one for the other.

Downstream: [tung-tung-sahur](https://github.com/Brainrotlang/tung-tung-sahur)'s `truncf()` helper exists only to contain this, and can now be deleted.

## Related Issue

Fixes #290

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement
- [ ] Refactor

Non-breaking in the sense that every changed behaviour was previously garbage — a bit pattern or an empty string. Code that somehow depended on either would change, but it was already not doing what it said.

## Checklist

`make test` 424 passed. `make valgrind` 407 cases, 0 errors, exit 0. `make format-check` clean. `make cppcheck` not run — not installed here.

### DIFF
--- a/ast.c
+++ b/ast.c
@@ -4063,6 +4063,26 @@ String evaluate_expression_string(ASTNode *node)
     }
 }
 
+/* An identifier's numeric value, converted from whatever type the variable
+ * actually has.
+ *
+ * handle_identifier(..., 0) hands back a pointer to the variable's OWN
+ * storage, so casting that to the type the caller happens to want is a
+ * reinterpretation, not a conversion: for `chad g = 17.5; rizz k = g;` it
+ * read the float's bit pattern as an int and produced 1099694080. A `smol`
+ * target read the wrong two bytes and produced 0.
+ *
+ * Promote mode 1 already does the real conversion for every numeric type,
+ * so route through it and narrow. Every caller below guards pointer_level
+ * first, which matters because mode 1 refuses a pointer outright. Every
+ * numeric member of the Value union (int, short, bool, float) is exactly
+ * representable as a double, so the intermediate loses nothing. */
+static double identifier_numeric_value(ASTNode *node, const String error)
+{
+    double *promoted = (double *)handle_identifier(node, error, 1);
+    return promoted != NULL ? *promoted : 0.0;
+}
+
 short evaluate_expression_short(ASTNode *node)
 {
     if (!node)
@@ -4097,7 +4117,9 @@ short evaluate_expression_short(ASTNode *node)
         }
         String error = {.data = "Undefined variable",
                         .len = sizeof("Undefined variable") - 1};
-        return *(short *)handle_identifier(node, error, 0);
+        /* Truncates toward zero from the variable's real type -- see
+           identifier_numeric_value(). */
+        return (short)identifier_numeric_value(node, error);
     }
     case NODE_OPERATION:
     {
@@ -4275,7 +4297,9 @@ int evaluate_expression_int(ASTNode *node)
         }
         String error = {.data = "Undefined variable",
                         .len = sizeof("Undefined variable") - 1};
-        return *(int *)handle_identifier(node, error, 0);
+        /* Truncates toward zero from the variable's real type -- see
+           identifier_numeric_value(). */
+        return (int)identifier_numeric_value(node, error);
     }
     case NODE_OPERATION:
     {
--- a/ast.c
+++ b/ast.c
@@ -3535,6 +3535,49 @@ static bool warn_if_native_result_void(const char *context_name)
     return true;
 }
 
+/* Load a numeric value from `addr`, reading it as `type` actually stores it.
+ *
+ * The counterpart of identifier_numeric_value() for the load sites that do
+ * not go through handle_identifier(): array elements and pointer targets.
+ * Those used to cast the address to whichever type the calling evaluator
+ * happened to want, which is a reinterpretation and not a conversion --
+ * `chad arr[1]; arr[0] = 17.5; rizz k = arr[0];` produced 1099694080, the
+ * bit pattern of 17.5f, and the reverse (`rizz arr[1]; chad f = arr[0];`)
+ * produced 0.00. Worse, printing and assigning the same rvalue disagreed,
+ * because yapping's dispatch reads it by type while the assignment did not.
+ *
+ * Every numeric member of the Value union is exactly representable as a
+ * double, so this is the lossless common currency; callers narrow. VAR_CHAR
+ * is one byte and unsigned on purpose -- char_scalar_slot_value() (ast.h)
+ * zero-extends a scalar yap on every write, so a signed packed read would
+ * disagree with a scalar read of the same logical value for every byte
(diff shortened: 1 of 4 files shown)


leo-aa88:
Blocking finding was right, and the framing was the useful part: I made conversion a `NODE_IDENTIFIER` special case and then claimed it as an invariant. Reproduced all of it before touching anything.

```
chad arr[1]; arr[0] = 17.5; rizz k = arr[0];   → 1099694080
chad g = 17.5; chad *p = &g; rizz k = *p;      → 1099694080
smol s = arr[0];                                → 0
gigachad a[1]; a[0] = 42.9; rizz k = a[0];     → 858993459
```

And the reverse, which my fixture *claimed* to pin and didn't, because that pin only ever saw identifiers:

```
rizz arr[1]; arr[0] = 7; chad f = arr[0];      → 0.00
rizz n = 7; rizz *p = &n; chad f = *p;         → 0.00
```

> What the fuck is the conversion supposed to guarantee if printing the value and assigning the value disagree?

Nothing, and that was the right question to ask. `print=17 assign=1099694080` on the same rvalue is the clearest statement of the defect in the whole thread.

**`numeric_load(addr, type)` is now the single typed read**, used by the integer *and* float evaluators across identifier, array-element and pointer-target. It subsumes the two `VAR_CHAR` width special cases the int evaluator already carried — so one place knows how each `VarType` is stored, instead of each evaluator casting an address to whatever it wanted. `NODE_STRUCT_ACCESS` is untouched, per your "do not add a fifth copy": it already switched on the field's declared type.

Which means **#290 was wrong about struct fields, and I repeated the claim instead of checking it.** `rizz k = p.x` with a `chad x` always gave 17. The fixture pins it now so it stays that way.

**On the fixture.** You were right that it couldn't fail. It now covers all four load shapes in both directions, and I verified it the way you'd want: reverted just the array and deref sites while keeping the identifier fix, and watched it go red. A partial fix no longer passes.

Also noting the two things you cleared up that I'd have got wrong: `CHAR` under `%d` is already handled by `apply_variadic_promotion()` widening to `STDROT_INT`, so it wasn't a leftover hole; and `VAR_BOOL` not being in the numeric compatibility group is the actual reason the `*(bool *)` identifier site is unreachable, which is a better justification than the one I gave for leaving it.

`make test` 424 passed, `make valgrind` 407 cases / 0 errors / exit 0, `make format-check` clean.

/cursor-review

PR #300, opened 2026-08-28
Adjust operator precedence for relational and postfix ops

### LINKED ISSUE #281: [Bug]: Two increment/decrement operators in one expression (i++ + i++) crash with a heap-buffer-overflow
### Description

Using **two** post/pre increment or decrement operators in a single binary expression (e.g. `i++ + i++`) corrupts a variable-name `String` and causes a **heap-buffer-overflow** read inside the hashmap, crashing the interpreter. It reproduces in every context tested — declaration initializer, assignment RHS, `yapping` argument, and a bare expression statement — and with distinct variables (`a++ + b++`), so it is not tied to any one statement kind or to aliasing the same variable.

The overflow happens when `handle_unary_expression()` (`ast.c`, OP_POST_INC/OP_POST_DEC) calls `get_variable_modifiers(node->data.unary.operand->data.name)` with a name `String` whose length no longer matches its backing buffer, so `fnv1a_hash()` reads past the AST arena region.

This is separate from #225 (that is the declaration/assignment *pre-visit* double-executing a subexpression; this crash occurs even for a `yapping(...)` argument and a bare statement, where no double-execution happens).

### Minimal reproduction code

```c
skibidi main {
    rizz i = 5;
    i++ + i++;          🚽 also crashes: rizz j = i++ + i++;  /  a++ + b++  /  i-- + i--
    yapping("%d", i);
    bussin 0;
}
```

### Expected behavior

The expression evaluates without memory corruption (whatever the language's chosen semantics for multiple increments in one expression are), and the interpreter does not crash.

### Actual behavior

```shell
=================================================================
==...==ERROR: AddressSanitizer: heap-buffer-overflow on address 0x... at pc 0x...
READ of size 1 at 0x... thread T0
    #0 ... in fnv1a_hash lib/hm.c:20
    #1 ... in hm_get lib/hm.c:132
    #2 ... in get_variable ast.c:5806
    #3 ... in get_variable_modifiers lang.y:2179
    #4 ... in handle_unary_expression ast.c:3133
    #5 ... in evaluate_expression_int ast.c:4131
    #6 ... in interpreter_visit_declaration interpreter.c:620
    ...
0x... is located 0 bytes to the right of 32792-byte region [...]
allocated by thread T0 here:
    #1 ... in region_new lib/arena.c:12
    #2 ... in arena_alloc lib/arena.c:59
    #3 ... in arena_alloc_astnode ast.c:1379
    ...
SUMMARY: AddressSanitizer: heap-buffer-overflow lib/hm.c:20 in fnv1a_hash
```

The corrupted read is of a variable-name `String` (`node->data.unary.operand->data.name`) that points into the AST node arena, so its `len` is over-reading past the region boundary.

### Brainrot version / commit

### Operating system

WSL2 (Ubuntu), x86-64

### Additional context

- A single increment operand (`i++ + 1`, `i++ + i`) does not crash; the trigger is two increment/decrement operands in one binary expression.
- Related but distinct correctness bug observed while narrowing this down: a single side-effecting increment in a **declaration initializer or assignment RHS** is applied **twice** — `rizz j = i++ + 1;` yields `i=7, j=7` instead of `i=6, j=6`, while the same subexpression as a `yapping` argument is correct (`prints 6, i=6`). That double-application looks like the same pre-visit re-evaluation mechanism described in #225 (there scoped to user-defined function calls), now visible for increment side effects. Worth confirming whether fixing #225 also fixes the double-application.

SpeedyMcMichael:
## Description

Fixes issue regarding double-postfix operations causing heap-buffer overflow by changing grammar precedence order in `lang.y`.

## Related Issue

Fixes #281 

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement
- [ ] Refactor

## Checklist

### DIFF
--- a/lang.y
+++ b/lang.y
@@ -469,10 +469,11 @@ static void register_anonymous_aggregate_typedef(String alias_name,
 %left OR                /* Logical OR */
 %left AND               /* Logical AND */
 %nonassoc EQ NE         /* Equality operators */
-%nonassoc LT GT LE GE DEC INC   /* Relational operators */
+%nonassoc LT GT LE GE   /* Relational operators */
 %left PLUS MINUS        /* Addition and subtraction */
 %left TIMES DIVIDE MOD  /* Multiplication, division, modulo */
 %right UMINUS           /* Unary minus */
+%left POSTFIX_OP DEC INC/* Unary postfix inc/dec ops */
 /* Member access `.` (struct_access: expression DOT IDENTIFIER) is a
    postfix operator and must bind TIGHTER than every binary/assignment
    operator above, matching C -- otherwise `v = m.x` parses as `(v = m).x`
@@ -2043,9 +2044,9 @@ unary_operation:
         { $$ = create_unary_operation_node(OP_PRE_INC, $2); }
     | DEC expression %prec LOWER_THAN_ELSE
         { $$ = create_unary_operation_node(OP_PRE_DEC, $2); }
-    | expression INC %prec LOWER_THAN_ELSE
+    | expression INC %prec POSTFIX_OP
         { $$ = create_unary_operation_node(OP_POST_INC, $1); }
-    | expression DEC %prec LOWER_THAN_ELSE
+    | expression DEC %prec POSTFIX_OP
         { $$ = create_unary_operation_node(OP_POST_DEC, $1); }
     ;
 


lang.y
```diff
@@ -469,10 +469,11 @@ static void register_anonymous_aggregate_typedef(String alias_name,
 %left OR                /* Logical OR */
 %left AND               /* Logical AND */
 %nonassoc EQ NE         /* Equality operators */
-%nonassoc LT GT LE GE DEC INC   /* Relational operators */
+%nonassoc LT GT LE GE   /* Relational operators */
 %left PLUS MINUS        /* Addition and subtraction */
 %left TIMES DIVIDE MOD  /* Multiplication, division, modulo */
 %right UMINUS           /* Unary minus */
+%left POSTFIX_OP DEC INC/* Unary postfix inc/dec ops */
```
leo-aa88, line 476: **BLOCKING (root cause is elsewhere).** Reordering INC/DEC precedence changes only how `i++ + i++` parses. It does not address the actual defect: `handle_unary_expression()` in `ast.c` reads `operand->data.name` for OP_POST_INC/DEC (and the pre- variants) with no check that the operand is an identifier node. Any non-name operand still reaches that read and over-reads the arena. This precedence table is not where #281 lives.

lang.y
```diff
@@ -2043,9 +2044,9 @@ unary_operation:
         { $$ = create_unary_operation_node(OP_PRE_INC, $2); }
     | DEC expression %prec LOWER_THAN_ELSE
         { $$ = create_unary_operation_node(OP_PRE_DEC, $2); }
-    | expression INC %prec LOWER_THAN_ELSE
+    | expression INC %prec POSTFIX_OP
```
leo-aa88, line 2047: The grammar still allows `expression INC` where `expression` is any expression, and the interpreter still assumes the operand is a bare name. On THIS branch, `(i + i)++` reproduces the exact heap-buffer-overflow from #281 (fnv1a_hash lib/hm.c:20 via get_variable_modifiers -> handle_unary_expression). The bug is not fixed, only one trigger is hidden.

Either restrict the inc/dec operand here to an lvalue so an invalid operand cannot be built, or make `handle_unary_expression` reject a non-lvalue operand with `yyerror` before touching `->data.name`. Also: no `test_cases/*.brainrot` + `expected_results.json` entry was added, which AGENTS.md requires for every bug fix.

leo-aa88 (changes requested):
# REQUEST CHANGES

This patch makes the crash in the issue's exact reproduction go away, but it does not fix the bug. It hides one trigger of a memory-safety defect that lives somewhere else entirely.

## What the issue actually is

The AddressSanitizer trace in #281 points straight at the root cause, and it is not the grammar: `handle_unary_expression()` in `ast.c` (OP_POST_INC / OP_POST_DEC / OP_PRE_INC / OP_PRE_DEC) reads `node->data.unary.operand->data.name` **without ever checking that the operand is an identifier node**. When the postfix operand is any non-name expression, `.name` reads a foreign union member, produces a `String` whose `len` runs off the end of the AST arena region, and `fnv1a_hash()` over-reads. That is the heap-buffer-overflow. The operator precedence is only relevant insofar as it decides *which* inputs happen to hand a non-name node to that unchecked read.

## Proof the bug is not fixed

I built this branch (all 424 pytest cases pass, no regression) and ran two inputs:

```
skibidi main {
    rizz i = 5;
    i++ + i++;          🚽 the issue's repro, now prints 7, no crash
    bussin 0;
}
```

```
skibidi main {
    rizz i = 5;
    (i + i)++;          🚽 STILL crashes on this branch
    bussin 0;
}
```

The second program produces the identical AddressSanitizer heap-buffer-overflow at `fnv1a_hash lib/hm.c:20`, through `get_variable_modifiers` -> `handle_unary_expression ast.c`, on this PR's HEAD. Same crash, same stack, different surface syntax. The precedence shuffle never touched the defect; it just stopped feeding it one particular sentence.

## The real fix

`handle_unary_expression` must reject a postfix/prefix inc/dec whose operand is not an lvalue (identifier, and whatever else counts as assignable here) and call `yyerror` instead of blindly reading `operand->data.name`. Ideally the grammar restricts the operand to an lvalue in the first place so the invalid state cannot even be constructed, but at minimum the interpreter must not treat an arbitrary expression node as a name. Until an invalid state cannot reach that union read, the class of bug is open.

## Missing test

AGENTS.md requires every bug fix to ship a `test_cases/*.brainrot` fixture plus a `tests/expected_results.json` entry. This PR adds neither, not even for the input it does change. A memory-safety fix with zero regression coverage cannot be verified by CI, and CI being green here proves only that the existing suite never exercised this path.

# VERDICT

This is a specification/root-cause miss, not an implementation detail. The defect is an unchecked union access in the interpreter; the PR edits the parser's precedence table and declares victory because the one example in the issue stopped crashing. `(i + i)++` still corrupts memory on this exact branch. Fix `handle_unary_expression` (and/or constrain the grammar's inc/dec operand to an lvalue), add the required `test_cases` fixture and `expected_results.json` entry covering both `i++ + i++` and a non-lvalue operand, and re-run valgrind. Not mergeable as is.

SpeedyMcMichael:
yeah bro ill fix it v2

lang.y
SpeedyMcMichael: just like

dont do the (i+i)++ thing though

it doesnt compile in C

why would you be compelled to do it in brainrotlang

you feel me man

anyways yes i will fix this in v2 with an lvalue check

leo-aa88:

PR #303, opened 2026-08-29
fix: a user-defined function call must run exactly once

leo-aa88:
## Description

`rizz a = f();` ran `f`'s body **twice** and kept the second result. Same for an assignment, and for a call anywhere in a right-hand expression. As an argument, a bare statement, or a condition it ran once.

```
declaration : body ran 2 time(s)
assignment  : body ran 2 time(s)
in an expr  : body ran 2 time(s)
argument    : body ran 1 time(s)
bare stmt   : body ran 1 time(s)
condition   : body ran 1 time(s)
```

`ast_accept()`'s generic pre-visit executed the call, and then the statement's own visitor evaluated the same node again for real.

**`interpreter_visit_function_call()` already declined to do this for native calls** — with a long comment explaining that executing during the pre-visit is *"actively wrong, not just redundant"* — and then executed user-defined ones anyway. The comment even names it:

> *"User-defined functions have no such cache and are still invoked from both places — a pre-existing gap, not introduced here, tracked separately from native-call support."*

The same reasoning applies to both, so the pre-visit is now a no-op for both.

## Why it survived, and what it broke

Pure functions were unaffected — you get the same answer twice. Anything with a side effect was not: loading a resource loaded it twice, advancing a generator advanced it twice.

**The sharpest shape is a function that reports whether it acted:**

```c
cap fired = try_once(&flag);   /* W on the first call, L once flag is set */
```

This returns the **second** attempt's answer — the one that found the work already done and declined. So the side effect happens exactly once while the report says it did not happen at all.

That is how I found it. In [tung-tung-sahur](https://github.com/Brainrotlang/tung-tung-sahur), `cap launched = player_jump(&pl)` was returning `L` on every successful jump, because the second invocation saw `grounded` already false. The jump sound never played, while the jump itself worked fine.

## One existing fixture changed, and it is the proof

`lit_salty_alias_use` counts calls in a `salty` static across two invocations. It expected **`2 4`** — which is only reachable if each call ran twice: 1,2 then 3,4. One invocation each gives **`1 2`**.

That expectation was encoding the bug. Its comment now records why the number moved.

## Verification

Both fixtures confirmed to **fail without the fix** (stash, rebuild, run). The new one covers all six syntactic positions plus the report-what-you-did shape.

`make test` 426 passed. `make valgrind` 408 cases, 0 errors, exit 0. `make format-check` clean. `make cppcheck` not run — not installed here.

## Two unrelated limitations found while writing the test

Avoided rather than exercised, and **not** fixed here — flagging in case they are news:

- Dereferencing a `cap *` is rejected outright: *"Cannot dereference a non-pointer expression"*.
- A pointer parameter whose name matches a local in the **caller** resolves against the caller's variable, so `*flag` is rejected as a non-pointer. Renaming either side fixes it.

Happy to open issues for either.

## Related Issue

No issue filed — found while building the game. Say the word and I'll open one.

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement
- [ ] Refactor

Non-breaking in the sense that no correct program depended on a function running twice — but any program that *was* relying on the doubled side effect (as `lit_salty_alias_use`'s expected output implicitly did) will change behaviour. That seemed worth stating plainly rather than burying.

## Checklist


leo-aa88:
Blocking finding was correct, and my own attempt to check it was worthless — worth saying first because it's the more useful part.

I compared the two builds with `git stash`, the stash didn't actually revert the binary, and both columns printed identical numbers, so I nearly replied that the regression didn't reproduce. Rebuilding properly from `HEAD~1` shows exactly what you described:

| | before | after the no-op |
|---|---|---|
| `f() + 0;` | 1 | **0** |
| `-f();` | 1 | **0** |
| `flex (…; …; f()+0)` | 2 | **0** |
| `bussin f() + 0;` in `skibidi` | 1 | **0** |

Making the pre-visit a no-op removed the *second* evaluator for a declaration or assignment, which was the point, and the *only* evaluator for a call inside any larger expression in statement position. `interpreter_accept_or_execute_call()` special-cases a bare `NODE_FUNC_CALL` and hands everything else to `ast_accept()`, whose expression visitors deliberately compute nothing.

**Fixed the owner, not the comment**, as you said. Statement-position expressions now get one `evaluate_expression()` — the same set `execute_statement()` has always covered. The void/`NONE` return arm had the identical shape (executed a bare call, dropped anything else), so it now evaluates a non-call expression for its effect.

**Two node types are deliberately excluded**, and getting that wrong would have traded this bug for its mirror image: a pre/post increment is evaluated by `interpreter_visit_unary_operation()` and an array access by `interpreter_visit_array_access()`, so routing those through `evaluate_expression()` as well would make `i++;` count twice. The fixture pins that too.

**The fixture now fails against the intermediate state.** With the pre-visit no-op but no statement evaluator, it reports `0, 0, 0, 0` where it should report `1, 1, 2, 1` — verified by building exactly that state. That was the thing you asked for and the thing my first version couldn't do.

One unrelated cleanup in the same push: a `git stash pop` I ran while investigating silently applied a stale unrelated stash and committed merge-conflict markers into `docs/ROADMAP.md`. Caught it in the diff stat, reverted the file to `origin/main` and amended. Nothing else in that file is touched.

`make test` 426 passed. `make valgrind` 408 cases, 0 errors, exit 0. `make format-check` clean.

/review

leo-aa88:
Both findings were real. Verified each before touching anything — `bussin arr[bump(p)]` in a `skibidi` function gave **2**, `make_pt(&n).x;` gave **0** — and the second one led somewhere worse than either of us said.

**One predicate now, as you asked.** The two owners kept separate lists and immediately disagreed: the statement path excluded increments and array accesses because `ast_accept` evaluates those, and the void arm did not. `ast_accept_evaluates_expression()` is exported from `ast.c` and both consult it.

**`f().x;`** — `NODE_STRUCT_ACCESS` was missing from the statement whitelist, so `ast_accept` walked the call straight into the no-op. It's in the list now. A member access whose base is a *call* gets its own arm, because `evaluate_expression()` can't resolve that shape at all (*"Unsupported struct member access expression"* — a separate pre-existing gap). The call's side effect is the whole observable point of such a statement, so the call runs and the field read is dropped, which is what happened before the no-op.

## The part that was mine, and worse than the finding

Wiring that up surfaced a memory bug I had introduced in the previous round and neither of us had spotted.

Making the void arm evaluate a discarded expression left that expression's result in the shared pending-return slot. **`current_return_value.value` is a union**, so an integer result occupies the same bytes as `.pvalue`; setting the type to `NONE` and leaving the value there hands the next return a slot whose "pointer" is a small integer, freed later as if it were heap.

```
bussin f() + 0;   in a skibidi function, then any struct-returning call
→ SEGV at 0xfffffffffffffff1
```

That address is `-15` — an integer being freed. The arm now releases the pending value and blanks the slot.

**The array form of that crash, `bussin arr[f()];`, reproduces on `origin/main` untouched**, so this fixes a pre-existing one as well as the one I introduced. I checked that explicitly rather than assuming, after my first attempt at this cleanup put the release in the wrong branch and changed nothing.

## On the mirror

Checked at each step that it didn't come back: increments and array accesses in statement position still run exactly once. The fixture now pins the six original positions, the four shapes from your last round, and the two from this one — 14 assertions in total.

`make test` 426 passed. `make valgrind` 408 cases, 0 errors, exit 0. `make format-check` clean.

Leftover you flagged and I have not done: the `visitor.c` NODE_RETURN and `free_pending_return_value` comments still describe the old double-invoke. Happy to update them in this PR if you'd rather they not go stale — I left them because both now describe behaviour that spans this change and the pre-existing pending-return handling, and I would rather rewrite them once that crash class is properly understood than half-correct them now.

/review

leo-aa88:
Both findings verified before changing anything: `bussin f().x`, `f().x + 0;` and `f().inner.x;` all errored and ran the call zero times, and `bussin x = f();` in a `skibidi` function errored with *"Invalid integer expression"*.

**The predicate was a partial list.** `ast_accept` evaluates a third thing — `interpreter_visit_assignment` performs the write — so the void arm ran `evaluate_expression` over an assignment that had already happened. `NODE_SIZEOF` is evaluated by its visitor too, and is now listed rather than being an unstated exception in a comment that claimed there were exactly two.

**The special case is gone.** `resolve_struct_access()` executes a call base itself, so a bare statement, a larger expression and a void `bussin` all reach one owner:

```
f().x;              1
f().x + 0;          1
bussin f().x;       1
bussin x = f();     1
```

It's gated on `report_errors`, which is the only thing separating the runtime resolve from `create_struct_access_node()`'s eager **parse-time** one — resolving a call base there would execute the program while it is still being read. Failure is silent rather than a `yyerror`, because `resolve_by_value_struct_source()` probes expressions through this path and a probe must not report. `infer_struct_def_static()` gained the matching static arm so the analyzer can type `f().x` from the declared return type without running anything.

## Not fixed, and not claimed

**`f().inner.x` is still rejected.** Typing a chain relies on the `var_type` that `create_struct_access_node()` fills in eagerly at parse time, and that resolve cannot type a call base without executing it — so the analyzer's chain branch sees `NONE` and reports "member access on a non-struct". Reaching it needs that branch to fall back to static inference. I tried exactly that, it did not work, and I removed the attempt rather than leave convoluted dead code behind while telling you it was handled.

The fixture pins the three shapes that do work and carries a comment saying why the fourth is absent, so nothing in it reads as a claim it doesn't make.

`make test` 426 passed. `make valgrind` 408 cases, 0 errors, exit 0. `make format-check` clean.

/review

leo-aa88:
Right, and the diagnosis was exact. `handle_function_call()`'s `VAR_STRUCT` case is discard-and-free — it exists so a struct return does *not* sit in the shared slot — so by the time it returned, `pvalue` was 0 and there was no field left to read. The call ran, which made the side effect look correct, and every read of the field silently produced 0:

```
rizz v = make_pt(&calls).x;   →  calls 1, v 0     (wanted 1 and 7)
```

Now uses `execute_function_call()`, leaving the struct in the pending slot for the next `free_pending_return_value()` to release, the way every other discarded struct return already works. A builtin base is refused up front, since no native can return a struct across the Road A ABI.

All seven read positions now give the right count **and** the right value — statement, declaration, assignment, larger expression, condition, variadic print, void `bussin`.

**On the fixture, which is the part I got wrong.** You said "pin the field value, not just the call count," and that was exactly the hole: printing `calls` made "the call ran" and "the call ran and its result survived" the same observation, so the fixture was blind to a version that destroyed the value. The struct's field is now `bump()+6` and every case asserts it. Verified by putting `handle_function_call()` back and watching the fixture fail.

**Also fixed the CI failure**, which was on the same arm: `nullPointerRedundantCheck` at `ast.c:905` — the branch dereferenced `obj` where the first branch in the chain null-checks it. Now `obj &&`, matching the two `else if` branches above. Worth noting that's the one job I can't run locally (no cppcheck here), so it caught something none of my local gates could.

`f().inner.x` remains rejected and unclaimed, for the reason in the last comment.

`make test` 426 passed. `make valgrind` 408 cases, 0 errors, exit 0. `make format-check` clean.

/review

leo-aa88:
Right — the value path had an owner and the type path did not. Every static type query resolved a member access by walking real storage, which a call base has none of until the call runs, and those queries are precisely the ones forbidden to run it. Each failed differently:

```
chad v = make_fl().f;         3.5        (worked)
yapping("%f", make_fl().f);   3.000000   (STDROT_INT fallback)
maxxing(make_pt().x);         "type is not statically known"
make_outer(&n).inner.x;       rejected as a non-struct
```

`static_struct_field()` answers all of it from declared types alone, executing nothing: a call base gives the function's declared return struct, a chain recurses through field types. Five sites fall back to it when the storage walk finds nothing — the eager parse-time resolve, `get_expression_type`, `get_expression_pointer_level`, `infer_runtime_expression_type_noeval`, and `is_expression`.

```
field chad   3.500000
field sizeof 4
field nested 7 1
```

**That also closes `f().inner.x`**, the fourth shape from two rounds ago that I had left explicitly unclaimed. It turned out to be the same defect rather than a separate one: typing a chain reads the `var_type` the eager parse-time resolve fills in, and that resolve is exactly the one that must not execute a call — so it needed this fallback, not a fifth special case.

**On the fixture** — you were right that it couldn't fail this, because it only printed ints. It now pins a `chad` field through `%f`, `maxxing()` of a field, and the nested read with its call count. Verified by neutering `static_struct_field()` and watching it fail.

One extra defect fixed on the way: a *native* call used as a member base is refused up front rather than being looked up as a struct, since no native can return one across the Road A ABI.

`make test` 426 passed. `make valgrind` 408 cases, 0 errors, exit 0. `make format-check` clean.

/review

leo-aa88:
Right on both counts — the sixth site, and the fact that my claim of "every static query" was one short.

`get_struct_def_for_expression()` was still walking storage. It's the one that needs a **layout** rather than a VarType, so it only started failing *because* the previous round fixed the type gate: `infer_runtime_expression_type_noeval()` began reporting VAR_STRUCT for `f().inner`, the "type is not statically known" check passed, and this failed immediately after it with *"Invalid type in sizeof"*. Same fallback as the other five.

```
field layout 8
```

**On pinning it** — your point that the existing `maxxing(make_pt().x)` is a scalar and can't fail this was right, and my first attempt at the assertion was no better: the fixture's `Pt` had one field, so the layout answer was 4, identical to the scalar one. `Pt` now has two fields so the size is 8 and distinct. Verified by unwiring the site and watching it fail.

**Three stale comments**, all mine, corrected rather than left to mislead:

- `ast_accept_evaluates_expression()` still said "Exactly two" exceptions after assignment and sizeof were added.
- The fixture still said `f().inner.x` is rejected. It works, and is pinned.
- The fixture still said the pending-slot SEGV was not this change's to fix, and that `void_arr` had to come last because of it. Blanking the union fixed that crash, so the ordering isn't load-bearing and the comment no longer claims it is — confirmed `void_arr` then `make_pt` is clean.

`make test` 426 passed. `make valgrind` 408 cases, 0 errors, exit 0. `make format-check` clean.

/review

ast.c
```diff
@@ -902,6 +968,62 @@ bool resolve_struct_access(ASTNode *node, StructDef **def_out, void **base_out,
...
+               fire on shapes that are merely being asked about rather than
+               used. The caller reports its own failure. */
+            return false;
+        }
+        parent_def = get_struct_def(current_return_value.desc.struct_name);
+        if (!parent_def)
+        {
+            return false;
+        }
+        /* Borrowed: the blob stays owned by the pending-return slot and is
```
leo-aa88: **[bug]** `parent_base` is a pointer into `current_return_value.value.pvalue`, documented as borrowed because "the caller reads its field immediately… before anything else can run". That is only true for the scalar `f().x` loads this PR tests.

`resolve_by_value_struct_source` (from `enter_function_scope`, `handle_return_statement`, struct copy-init) stores that pointer and uses it later. `enter_function_scope` already memcpy's a bare `NODE_FUNC_CALL` argument into an owned temp and frees the slot — then falls through for `f().inner`, which is the same hazard without the copy. `take(make_outer().inner, g())` is ASAN heap-use-after-free: `g()`'s `free_pending_return_value()` frees the blob, then the bind-loop memcpy reads it. `bussin make_outer().inner` also leaks the Outer blob.

This is new: before, a FUNC_CALL object hit "Unsupported struct member access" and never allocated the temp.

**Suggestion:** Do not return a pointer into the pending-return slot from a function that also lets the next call free that slot. Copy into owned storage (same as the `NODE_FUNC_CALL` argument arm) and free the pending blob before returning, including in `handle_return_statement` so the Outer is not leaked. Until that exists, reject `f().inner` as a by-value struct source rather than borrowing the temp.

ast.c
```diff
@@ -902,6 +968,62 @@ bool resolve_struct_access(ASTNode *node, StructDef **def_out, void **base_out,
...
+                              obj->data.func_call.arguments);
+        if (current_return_value.desc.type != VAR_STRUCT ||
+            current_return_value.desc.pointer_level != 0 ||
+            !current_return_value.value.pvalue)
+        {
+            /* Silent: resolve_by_value_struct_source() probes expressions
+               through here with report_errors set, so a hard error would
+               fire on shapes that are merely being asked about rather than
+               used. The caller reports its own failure. */
+            return false;
```
leo-aa88, line 1012: **[bug]** On `report_errors=true`, every other `resolve_struct_access` failure path `yyerror`s. This arm returns false silently when the call did not produce a by-value struct. The comment claims `resolve_by_value_struct_source()` "probes expressions through here with `report_errors` set". That is false: all three callers pass `true` because they are the real copy/return/argument paths, and they emit no second diagnostic on the assumption this helper already reported.

Empirically `rizz x = not_a_struct().nope;` now runs the call, member access fails quietly, the evaluator returns 0, and the process exits 0. Semantic analysis does not save this: `NODE_STRUCT_ACCESS` still `break`s on a call base.

**Suggestion:** When `report_errors` is true and the call is not a by-value struct (or has no heap blob), `yyerror` and return false. Silent return only for `report_errors=false` probes. Delete the comment; it describes the opposite of the callers.

test_cases/func_call_evaluated_once.brainrot
```diff
@@ -0,0 +1,198 @@
...
+🚽 The sharpest version is the last case here. A function that reports
+🚽 whether it did something returns the SECOND attempt's answer -- the one
+🚽 that found the work already done and declined -- so the side effect
+🚽 happens exactly once while the report says it did not happen at all.
+🚽 A void function whose `bussin` expression is discarded: the value-bearing
+🚽 return arms never see it, so the void arm is the only thing that can run
+🚽 it for its side effect.
+skibidi void_ret(rizz *n) { bussin bump(n) + 0; }
+
+🚽 The mirror hazard. ast_accept()'s pre-visit DOES evaluate an array access,
```
leo-aa88, line 20: **[suggestion]** Remaining double-eval the fixture's "whatever syntactic position" claim does not cover. `interpreter_visit_array_access()` still performs the access during `ast_accept`'s pre-walk; declaration/assignment then evaluate the RHS again. `rizz x = a[bump(&calls)-1]` still runs `bump` twice. The new `ast_accept_evaluates_expression()` predicate exists so the void-`bussin` arm does not make this mistake; the declaration/assignment walk never consults it. The fixture tests `bussin a[bump(n)]`, which is the one array shape that path now gets right.

**Suggestion:** Add `rizz x = a[f()]` and `x = a[f()]` to the fixture. Skip the declaration/assignment RHS pre-walk when `ast_accept_evaluates_expression(right)` is true, or stop evaluating array access during a generic pre-visit.

semantic_analyzer.c
```diff
@@ -601,6 +601,21 @@ static StructDef *infer_struct_def_static(ASTNode *expr,
         return get_struct_def(sym->struct_name);
     }
 
+    if (expr->type == NODE_FUNC_CALL)
```
leo-aa88, line 604: **[suggestion]** `infer_struct_def_static()` now understands a `NODE_FUNC_CALL` base, so type-inference of `f().x` works. The main `NODE_STRUCT_ACCESS` walker in `semantic_analyze_with_scope_tracking` still only handles identifier / nested access / array-element objects and `break`s on a call, so it never validates the member, never rejects a non-struct return, and never writes `var_type`/`struct_name` unless parse-time `static_struct_field()` already did. That is why a non-struct `f().x` is reachable at runtime with no semantic error, and why a chain's types are pinned to the eager parse-time resolve.

**Suggestion:** Give the walker a `NODE_FUNC_CALL` arm using the same declared-return-type rules as this helper (by-value struct, `pointer_level == 0`), then populate the node the way the identifier arm does. Do not rely on parse-time `get_function()` having succeeded.

interpreter.c
```diff
@@ -313,22 +313,25 @@ static void interpreter_execute_call_statement(ASTNode *node)
    downstream evaluate_expression_*() call populate the memo cache itself,
    at the right time. (Bare statement-position and for-loop init/incr
    calls never reach here at all -- see interpreter_execute_call_statement()
-   above.) User-defined functions have no such cache and are still invoked
-   from both places -- a pre-existing gap, not introduced here, tracked
-   separately from native-call support. */
+   above.)
+
+   The same reasoning applies to a USER-DEFINED function, which is why this
```
leo-aa88: **[suggestion]** Several new comments narrate the bug's discovery, a discarded revision, and what a wrong implementation looked like, rather than stating the invariant the next editor must preserve. The load-bearing facts (pre-visit must not execute; `report_errors=false` must not execute; void `bussin` must not re-evaluate a node the visitor already computed; pending-return `pvalue` overlaps the union) are buried in the memoir.

**Suggestion:** Cut each of those to the invariant and the non-obvious reason it exists. Bug history belongs in the PR body, which already has it.

ast.c
```diff
@@ -1988,7 +2110,13 @@ int get_expression_pointer_level(ASTNode *node)
...
         if (!resolve_struct_access(node, &def, &base, &fld, false))
-            return 0;
+        {
+            /* Declared types answer a call base; nothing here may run it. */
+            fld = static_struct_field(node);
+            if (!fld)
+                return node->pointer_level;
+        }
+        return fld->desc.pointer_level;
         return fld->desc.pointer_level;
```
leo-aa88: **[nit]** After the `static_struct_field` fallback, `return fld->desc.pointer_level;` is written twice; the second is dead. Same leftover in `get_expression_type` at ast.c:2292 (`return fld->desc.type;` twice). Harmless under this GCC, but it is the tell that the hunk was patched around the old return instead of replacing it.

**Suggestion:** Delete the unreachable second return in both functions.

visitor.c
```diff
@@ -121,23 +121,16 @@ void ast_accept(ASTNode *node, Visitor *visitor)
...
-                op == OP_PRE_DEC)
-            {
-                skip_decl_right_visit = true;
-            }
-        }
-
-        // Visit the initializer expression first (unless it's a side-effect
-        // operation)
-        if (!skip_decl_right_visit && node->data.op.right)
+            !ast_accept_evaluates_expression(node->data.op.right))
```
leo-aa88: **[suggestion]** `ast_accept_evaluates_expression()` is applied only to the declaration/assignment RHS *root*. `interpreter_visit_array_access()` still evaluates during any pre-walk that reaches a nested `NODE_ARRAY_ACCESS`, and the statement visitor then evaluates the whole expression again.

Empirically `rizz x = arr[bump(&calls)-1] + 0;` prints `nested 2 8` (bump ran twice). Same for `bussin a[bump(n)] + 0;`. The fixture only pins the root form `rizz ai = arr[bump(&calls)-1]`, which this skip already handles. An implementation that still double-evals any wrapped array access still passes.

The last review said: do not merely report the missed node; report the broken invariant. Traversal ownership is still "the predicate, but only if the evaluating node is the root."

**Suggestion:** Either stop evaluating in `interpreter_visit_array_access()` and always evaluate from the statement/return arm, or skip the pre-walk whenever an evaluating node appears *anywhere* under the expression the later visitor will recompute. Pin `rizz x = a[f()] + 0` and `bussin a[f()] + 0`.

ast.c
```diff
@@ -902,6 +968,65 @@ bool resolve_struct_access(ASTNode *node, StructDef **def_out, void **base_out,
...
+                              obj->data.func_call.arguments);
+        if (current_return_value.desc.type != VAR_STRUCT ||
+            current_return_value.desc.pointer_level != 0 ||
+            !current_return_value.value.pvalue)
+        {
+            /* Every other failure path here reports, and no caller adds a
+               second diagnostic -- they all assume this one spoke. Staying
+               silent let `rizz x = not_a_struct().nope;` run the call,
+               yield 0 and exit 0. Semantic analysis does not catch it
+               either, so this is the only check. */
```
leo-aa88: **[suggestion]** Finding 5 is still present, and this comment is now false rather than merely historical. "Semantic analysis does not catch it either, so this is the only check." This same commit's `semantic_error_struct_access_call_base.brainrot` is rejected *before* interpret, with the analyzer's "Member access on non-struct/union value". The runtime `yyerror` here is defense-in-depth, not "the only check."

Two more claims this branch made false and did not touch (not in this hunk, but still HEAD):
- `visitor.c:193` — `interpreter_visit_function_call()` "executes a user-defined call and frees its result." That function is a no-op on this branch.
- `ast.c:6610` — a user-defined function "is genuinely invoked twice as a result" of the pre-visit. That is the bug this PR exists to kill.

Adjacent new comments still narrate the bug hunt (`take(make_outer().inner, g())` UAF, chad `%f` printing as int, "the sixth site") instead of stating the invariant.

**Suggestion:** State the invariant only: `report_errors=false` does not execute; a call-base field that is kept past the next `free_pending_return_value()` is copied and owned; the pre-visit must not run a call. Delete the discovery narrative. Correct the three claims that no longer match HEAD.

PR #304, opened 2026-08-29
fix(arena): preserve allocated regions and reset counts in arena_reset

### LINKED ISSUE #286: [Bug]: arena_reset() frees regions it should keep and leaves arena->start dangling (latent use-after-free)
### Description

`arena_reset()` (`lib/arena.c:126`) is supposed to make the arena reusable by resetting allocation counts while keeping the underlying regions. Instead it **frees the regions it is meant to keep** and leaves `arena->start` dangling:

```c
void arena_reset(Arena *arena)
{
    arena->end = arena->start;              // end -> first region (call it A)
    while (arena->end->next != NULL)
    {
        Region *next = arena->end->next;
        region_free(arena->end);            // frees A -- but arena->start STILL points at A
        arena->end = next;
    }
    arena->end->count = 0;                  // only the LAST region is reset
}
```

For an arena `A -> B -> C` this frees A and B, sets `C->count = 0`, and leaves `arena->start` pointing at the now-freed A. After the call:

- `arena->start` is a dangling pointer to freed memory.
- Only the last region survives (the rest are freed, not reset), defeating the purpose of "reset".
- A subsequent `arena_free(arena)` walks the list starting from `arena->start` → **use-after-free / double-free**.
- A subsequent `arena_reset(arena)` dereferences the dangling `arena->start` → **use-after-free**.

### Minimal reproduction code

```c
🚽 Not reachable from Brainrot source today — arena_reset() has no callers.
🚽 Equivalent C usage that triggers it:
🚽   Arena a = {0};
🚽   arena_alloc(&a, BIG); arena_alloc(&a, BIG); arena_alloc(&a, BIG); // forces multiple regions
🚽   arena_reset(&a);   // frees regions still referenced by a.start
🚽   arena_free(&a);    // walks from freed a.start -> use-after-free / double-free
```

### Expected behavior

A reset keeps all regions allocated, sets `count = 0` on each (or at least on the first, with `end` pointing back at `start`), and frees nothing — so the arena can be refilled without reallocating and remains safe to `arena_free()` afterward.

### Actual behavior

```shell
Regions are freed while still referenced by arena->start; a later arena_free()
or arena_reset() is a use-after-free / double-free (would be flagged by ASan).
```

### Brainrot version / commit

### Operating system

WSL2 (Ubuntu), x86-64

### Additional context

**This is currently latent**: `arena_reset()` has no callers (only its declaration in `lib/arena.h:27` and definition in `lib/arena.c:126`), so it does not affect any program today. But an arena "reset" is an obvious thing for future code to reach for, and the implementation is unsafe as written — worth fixing or removing before someone wires it up. A correct implementation would keep the region chain intact and just walk it setting `count = 0`, then set `arena->end = arena->start`.

Aditya-233:
## Description

Fixes an issue where `arena_reset()` freed intermediate regions in the linked list while leaving `arena->start` dangling. The implementation now safely walks the allocated region chain, resets `count = 0` on each region while preserving the allocated blocks, and resets `arena->end` back to `arena->start`.

Additionally adds a host C unit test (`tests/arena/arena_check.c`, wired as `arena-check` in `Makefile` and `make test`) verifying under ASan/UBSan:
1. Multi-region allocation (forcing $\ge 3$ regions), `arena_reset()`, followed by `arena_free()`, verifying no dangling pointer or use-after-free occurs.
2. Complete reuse of all regions across allocations after reset, asserting exact pointer equality with pre-reset chunk allocations and verifying no excess regions are created.
3. Reset on `{0}` / unallocated arena, `NULL` pointer safety, and repeated / consecutive resets without crashing.

## Related Issue

Fixes #286

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement
- [ ] Refactor

## Checklist

- [ ] I have documented my changes in the code or documentation

### DIFF
--- a/lib/arena.c
+++ b/lib/arena.c
@@ -119,22 +119,22 @@ String arena_strdup(Arena *arena, String str)
 }
 
 /*
- * @brief reset the arena.
- * @param arena The arena to reset.
+ * @brief Reset the arena for reuse.
+ * Frees nothing: every region stays allocated with count = 0, and end is
+ * rewound to start, so the arena refills without reallocating and remains
+ * safe to arena_free(). Safe on NULL and on a never-allocated arena.
+ * @param arena The arena to reset (may be NULL).
  */
-
 void arena_reset(Arena *arena)
 {
-    if (arena == NULL || arena->start == NULL)
+    if (arena == NULL)
     {
         return;
     }
 
-    Region *current = arena->start;
-    while (current != NULL)
+    for (Region *r = arena->start; r != NULL; r = r->next)
     {
-        current->count = 0;
-        current = current->next;
+        r->count = 0;
     }
     arena->end = arena->start;
 }
--- a/lib/arena.c
+++ b/lib/arena.c
@@ -125,14 +125,18 @@ String arena_strdup(Arena *arena, String str)
 
 void arena_reset(Arena *arena)
 {
-    arena->end = arena->start;
-    while (arena->end->next != NULL)
+    if (arena == NULL || arena->start == NULL)
     {
-        Region *next = arena->end->next;
-        region_free(arena->end);
-        arena->end = next;
+        return;
     }
-    arena->end->count = 0;
+
+    Region *current = arena->start;
+    while (current != NULL)
+    {
+        current->count = 0;
+        current = current->next;
+    }
+    arena->end = arena->start;
 }
 
 /*
--- a/tests/arena/arena_check.c
+++ b/tests/arena/arena_check.c
@@ -1,0 +1,156 @@
+/* tests/arena/arena_check.c -- unit tests for arena allocator and arena_reset
+ *
+ * Exercises the arena_reset contract (issue #286):
+ * 1. Force >= 3 regions, reset, then free (catches dangling start / UAF).
+ * 2. Reset, realloc, and assert all regions are reused, including
+ *    pointer equality with the first pre-reset allocation (catches partial
+ *    count reset).
+ * 3. Reset of {0} / unallocated arena and second/repeated resets (catches
+ *    crashes on empty/repeated reset).
+ */
+
+#include "lib/arena.h"
+#include <assert.h>
+#include <stdint.h>
+#include <stdio.h>
+#include <string.h>
+
+static void test_arena_multi_region_reset_and_free(void)
+{
+    Arena arena = {0};
+    const size_t chunk_size = DEFAULT_REGION_SIZE * sizeof(uintptr_t);
+
+    /* Allocate 3 large chunks to force at least 3 distinct regions. */
+    void *p1 = arena_alloc(&arena, chunk_size);
+    void *p2 = arena_alloc(&arena, chunk_size);
+    void *p3 = arena_alloc(&arena, chunk_size);
+
+    assert(p1 != NULL);
+    assert(p2 != NULL);
+    assert(p3 != NULL);
+    assert(p1 != p2);
+    assert(p2 != p3);
+
+    /* Verify we have at least 3 regions linked. */
+    int region_count = 0;
+    Region *r = arena.start;
+    while (r != NULL)
+    {
+        region_count++;
+        r = r->next;
+    }
+    assert(region_count >= 3);
+
+    /* Reset the arena: must preserve all allocated regions with count = 0. */
+    arena_reset(&arena);
+    assert(arena.start != NULL);
+    assert(arena.end == arena.start);
+
+    int reset_region_count = 0;
+    r = arena.start;
+    while (r != NULL)
+    {
+        assert(r->count == 0);
+        reset_region_count++;
+        r = r->next;
+    }
+    assert(reset_region_count == region_count);
+
+    /* Under the old bug, regions were freed during reset and arena.start
+     * was left dangling; arena_free would then trigger use-after-free. */
+    arena_free(&arena);
+    assert(arena.start == NULL);
+    assert(arena.end == NULL);
+}
+
+static void test_arena_reset_reuse_allocations(void)
+{
+    Arena arena = {0};
+    const size_t chunk_size = DEFAULT_REGION_SIZE * sizeof(uintptr_t);
+
+    /* Initial allocations forcing 3 regions. */
+    void *orig_p1 = arena_alloc(&arena, chunk_size);
+    void *orig_p2 = arena_alloc(&arena, chunk_size);
+    void *orig_p3 = arena_alloc(&arena, chunk_size);
+
+    assert(orig_p1 != NULL && orig_p2 != NULL && orig_p3 != NULL);
+
+    /* Reset. */
+    arena_reset(&arena);
+    assert(arena.end == arena.start);
+
+    /* Reallocate the same chunks: must reuse existing regions and match
+     * original pointers. If count was only reset on start, subsequent
+     * allocations would skip existing regions and allocate new ones. */
+    void *new_p1 = arena_alloc(&arena, chunk_size);
+    void *new_p2 = arena_alloc(&arena, chunk_size);
+    void *new_p3 = arena_alloc(&arena, chunk_size);
+
+    assert(new_p1 == orig_p1);
+    assert(new_p2 == orig_p2);
+    assert(new_p3 == orig_p3);
+
+    /* Confirm no new 4th region was added. */
+    int count = 0;
+    Region *r = arena.start;
+    while (r != NULL)
+    {
+        count++;
+        r = r->next;
+    }
+    assert(count >= 3);
+    assert(arena.end->next == NULL);
+
+    arena_free(&arena);
+}
+
+static void test_arena_reset_empty_and_repeated(void)
+{
+    /* 1. Reset on a zero-initialized / never-allocated arena. */
+    Arena empty = {0};
+    arena_reset(&empty);
+    assert(empty.start == NULL);
+    assert(empty.end == NULL);
+
+    /* Second reset must also be safe. */
+    arena_reset(&empty);
+    assert(empty.start == NULL);
+    assert(empty.end == NULL);
+    arena_free(&empty);
+
+    /* 2. NULL pointer safety. */
+    arena_reset(NULL);
+    arena_free(NULL);
+
+    /* 3. Consecutive resets after allocation. */
+    Arena arena = {0};
+    void *p = arena_alloc(&arena, 256);
+    assert(p != NULL);
+
+    arena_reset(&arena);
+    arena_reset(&arena);
+
+    void *p_after = arena_alloc(&arena, 256);
+    assert(p_after == p);
+
+    arena_free(&arena);
+
+    /* 4. Multiple alloc-reset cycles. */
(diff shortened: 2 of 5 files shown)


leo-aa88 (changes requested):
Always provide tests proving your fix landed when implementing a fix PR. I'll provide a more thorough review.

leo-aa88 (approved):
LGTM

lib/arena.c
```diff
@@ -125,14 +125,18 @@ String arena_strdup(Arena *arena, String str)
 
 void arena_reset(Arena *arena)
```
leo-aa88, line 128: **[bug]** This is the entire fix and nothing calls it. `make test` cannot see a revert of this function.

A `.brainrot` fixture cannot reach `arena_reset`. The issue’s own C reproduction (three oversized allocs → reset → free) is the test. Wire it into `make test` as a host binary, same pattern as `tests/abi/struct_layout_abi_check`. Also assert post-reset reuse of the first region — otherwise a “reset” that only zeros `start->count` still passes.

leo-aa88 (changes requested):
## Missing tests

The reset implementation matches #286. That is not the same as being done.

`arena_reset` still has zero callers. This PR adds no test. `make test` never enters this function. The old body — free the chain, leave `start` dangling, UAF in `arena_free` — would still be green CI. Issue #286 already included the reproduction. It is not in the tree.

The checklist item “I have added tests that prove my changes work (if applicable)” is checked. A latent memory-safety bug is the applicable case. `(if applicable)` is not an exemption because the existing pytest fixtures cannot reach C.

`make test` already runs a host C binary (`tests/abi/struct_layout_abi_check`). Add the same kind of thing for the arena. Minimum, under ASan/UBSan:

1. Force ≥3 regions, `arena_reset`, `arena_free` — old code dies here.
2. Reset, then realloc, and assert the first region is reused (pointer equality with the first pre-reset allocation). Zeroing only `start->count` and leaving later leftover counts would still pass (1).
3. Reset of `{0}` / never-allocated, and a second reset, must not crash.

CI is green. This is not a failing-test problem. The current tests do not exercise this contract.

Aditya-233:
Thanks

I have added the host C unit test suite in `tests/arena/arena_check.c` and wired it into `make test` via the `arena-check` target (as well as `make clean`, `.gitignore`, and the CI workflow under `.github/workflows/ci.yml`).

Verified locally:
- `make arena-check` passes cleanly under ASan/UBSan.
- `make test` (422 test cases + `abi-check` + `arena-check`) passes.
- `make valgrind` reports 0 errors and 0 leaks across all test cases.
- `make cppcheck` passes cleanly.

leo-aa88 (commented):
# COMMENT

The fix is correct and this time the tests are real. I reverted `lib/arena.c` to the old body and ran the new binary: it dies at `assert(arena.end == arena.start)` before it even reaches the use-after-free. Reset now frees nothing, zeroes every count, rewinds `end` — exactly #286's expected behavior — and all three tests the previous review asked for are present. Format-check is clean on both changed files, and the CI step plus the `make test` prerequisite mean this can't silently stop running.

Four non-blocking things, all cheap. None of them stop a merge.

# VERDICT

The implementation is right; what's left is documentation and test strength. The docblock still doesn't state the retention/ownership contract whose absence caused the bug in the first place, one test comment claims a stronger check than the assert performs, and the suite validates pointer identity after reset without ever writing a byte to the memory it claims is reusable. Fix those and this is done.

Merge after the docblock and the `count >= 3` assert are addressed; the other two are optional.

lib/arena.c
leo-aa88, line 128: The entire bug here was an unstated contract, and the docblock still says nothing about it. `@brief reset the arena` is what the old, region-freeing version could also have claimed. State the postcondition so the next person doesn't "optimize" the frees back in:

```c
/*
 * @brief Reset the arena for reuse.
 * Frees nothing: every region stays allocated with count = 0, and end is
 * rewound to start, so the arena refills without reallocating and remains
 * safe to arena_free(). Safe on NULL and on a never-allocated arena.
 * @param arena The arena to reset (may be NULL).
 */
```

The checklist box "I have documented my changes in the code or documentation" is unchecked, which is honest — this is the one function in the file where that box matters.

lib/arena.c
```diff
@@ -125,14 +125,18 @@ String arena_strdup(Arena *arena, String str)
...
 {
-    arena->end = arena->start;
-    while (arena->end->next != NULL)
+    if (arena == NULL || arena->start == NULL)
```
leo-aa88: `|| arena->start == NULL` is dead weight, and strictly weaker than not having it. With `start == NULL` the loop body never executes and `arena->end = arena->start` sets `end = NULL` — identical result for every reachable state.

The only case where the two differ is the malformed one (`start == NULL`, `end != NULL`), and there the early return *skips* re-establishing `end == start` — the single postcondition this function exists to guarantee. A guard that opts out of restoring the invariant precisely when the invariant is broken is backwards. ಠ_ಠ

```c
if (arena == NULL)
    return;
for (Region *r = arena->start; r != NULL; r = r->next)
    r->count = 0;
arena->end = arena->start;
```

tests/arena/arena_check.c
```diff
@@ -0,0 +1,156 @@
...
+
+    /* Confirm no new 4th region was added. */
+    int count = 0;
+    Region *r = arena.start;
+    while (r != NULL)
+    {
+        count++;
+        r = r->next;
+    }
+    assert(count >= 3);
```
leo-aa88: The comment says "no new 4th region was added." The assert says `>= 3`, which passes with 4, 40, or 400 regions. It confirms nothing the comment claims — the `new_pN == orig_pN` asserts above are what actually catch region growth.

In a file whose whole job is to be the contract oracle, a comment that overstates its assert is the same class of defect this PR is fixing. Capture the region count before the reset and assert equality, the way `test_arena_multi_region_reset_and_free` already does with `reset_region_count`.

tests/arena/arena_check.c
```diff
@@ -0,0 +1,156 @@
...
+
+static void test_arena_multi_region_reset_and_free(void)
+{
+    Arena arena = {0};
+    const size_t chunk_size = DEFAULT_REGION_SIZE * sizeof(uintptr_t);
+
+    /* Allocate 3 large chunks to force at least 3 distinct regions. */
+    void *p1 = arena_alloc(&arena, chunk_size);
+    void *p2 = arena_alloc(&arena, chunk_size);
+    void *p3 = arena_alloc(&arena, chunk_size);
```
leo-aa88, line 26: Nothing in this file ever writes a byte into an arena chunk. The suite proves pointer *identity* after reset; it never proves the returned memory is legal to use.

`arena_alloc` returns `&arena->end->data[arena->end->count]` inside a `malloc(sizeof(Region) + sizeof(uintptr_t) * capacity)` block. An off-by-one in the count/capacity arithmetic hands back a pointer past the end of that block and every assert here still passes — the addresses would still compare equal to the pre-reset ones. ASan is linked in and sees nothing, because nothing touches the memory.

One line per chunk turns ASan into a real oracle for the "refill without reallocating" claim:

```c
memset(p1, 0xAB, chunk_size);
```

(`<string.h>` is already included and currently unused.)

Aditya-233:
Thanks! I have addressed all 4 points in latest commit:

1. **Docblock (`lib/arena.c`)**: Updated the docblock on `arena_reset` to explicitly document the retention and ownership contract, postconditions, and `NULL` / unallocated safety.
2. **Simplified guard & loop (`lib/arena.c`)**: Removed the redundant `|| arena->start == NULL` check and switched to the clean `for (Region *r = arena->start; r != NULL; r = r->next)` loop, always ensuring `arena->end = arena->start`.
3. **Exact region count assertion (`tests/arena/arena_check.c`)**: Captured `orig_region_count` before reset and asserted exact count equality `assert(new_region_count == orig_region_count)` after reallocating.
4. **Buffer writes with `memset` (`tests/arena/arena_check.c`)**: Added `memset` across chunks upon allocation and re-allocation to ensure ASan actively validates chunk memory bounds and usability.

All tests (`make arena-check`, `make test`, `make cppcheck`, and `make valgrind`) pass cleanly locally.

lib/arena.c
```diff
@@ -119,20 +119,24 @@ String arena_strdup(Arena *arena, String str)
...
-    while (arena->end->next != NULL)
+    if (arena == NULL)
     {
-        Region *next = arena->end->next;
-        region_free(arena->end);
-        arena->end = next;
+        return;
     }
-    arena->end->count = 0;
+
```
leo-aa88, line 134: Correct fix. Verified: the previous body left `arena->start` pointing at a freed region (it advanced `end` but never `start`), so a following `arena_free()` was a use-after-free, and it also dereferenced `start->next` on a `{0}` arena. Zeroing every region's `count` and rewinding `end = start` without freeing is exactly what `arena_alloc`'s refill traversal needs, and it matches the new docblock. Compiling the added test against the old body reproduces the failure, so the regression is genuinely covered.

leo-aa88 (approved):
LGTM.

I checked this out and put it through the wringer rather than trusting the description. Summary of what I verified:

**The fix is correct and matches its docblock.** The old `arena_reset` set `arena->end = arena->start` and then freed regions starting from `arena->end` without ever advancing `arena->start`, so `arena->start` was left dangling and a later `arena_free()` was a use-after-free (and it dereferenced `arena->start->next` on a `{0}` arena). The new version frees nothing, zeroes `count` on every region in the chain, and rewinds `end` to `start`. That is exactly what the comment now promises, and it is what the bump allocator in `arena_alloc` needs to reuse regions on refill.

**The test actually tests the contract, not just the implementation.** I compiled the new `arena_check.c` against the OLD buggy `arena_reset` and it fails at `assert(arena.end == arena.start)` before it can even reach the UAF, so the regression is genuinely caught. The pointer-equality assertions (`new_p1 == orig_p1`, etc.) also catch the plausible half-fix of zeroing only `start->count`: I traced that a start-only reset sends the second reallocation into an unreset region and forces a fresh `region_new`, which the exact-region-count and pointer assertions reject. Empty/`{0}`, NULL, and repeated-reset paths are covered too.

**Sanitizer coverage is in the right place.** `arena-check` compiles `lib/arena.c` and `lib/mem.c` directly with the project's `-fsanitize=address,undefined -Werror` flags, so this exercises the allocator under sanitizers that the dlopen'd `libstdrot.so` does not get. It builds and passes clean here.

**CI wiring is right, not redundant.** The workflow runs `pytest` directly rather than `make test`, so the standalone `make arena-check` step is the only thing that runs this test on the merge gate. Adding it to the `test` target as well is fine for local runs. `.gitignore` and the `clean` target are both updated for the new binary.

Full `pytest` run is green (427 passed) with no regression. I could not run `make cppcheck` locally (this box has cppcheck 2.7, below the required 2.13); the CI static-analysis job will cover that.

Optional, non-blocking: the tests use a single uniform chunk size. A mixed-size alloc/reset/refill cycle would harden the reuse assertions against future changes to region-sizing, but the current contract (preserve regions, zero counts, rewind end) is fully covered as is. Ship it.

Issue #282, opened 2026-08-27
[Bug]: giga/thicc (long/long long) silently truncate to 32 bits — no 64-bit storage, literal, or marshalling


leo-aa88:
### Description

`giga` (`long`) and `thicc` (`long long`) are documented as ✅ implemented (README keyword table; feature issues #125/#126 are closed), but there is **no 64-bit integer storage anywhere in the pipeline**. A `giga`/`thicc` value that doesn't fit in 32-bit `int` is silently truncated. Three separate sites are all 32-bit:

1. **Storage** — the scalar value union has only `int ivalue;` (`ast.h:269`); there is no 64-bit member. A `giga` variable is stored in a 32-bit slot.
2. **Literal lexing** — every integer literal is parsed with `atoi()` into `yylval.ival` (an `int`), `lang.l:617/626/630`. A literal above `INT_MAX` overflows (`atoi` on overflow is undefined behavior) and truncates. There is no 64-bit literal token.
3. **Native-call marshalling** — a `VAR_INT` value marshals to `StdrotValue` as `STDROT_INT` with `out->val.i = var->value.ivalue` (`stdrot.c:1029`); `StdrotValue` has no 64-bit integer variant, so anything printed via `yapping`/`baka` is truncated to 32 bits.

Net effect: `giga`/`thicc` behave exactly like `rizz` (`int`), just with different type-checking names.

### Minimal reproduction code

```c
skibidi main {
    giga rizz x = 5000000000;   🚽 5e9, needs 64 bits
    yapping("%lld", x);         🚽 prints 705032704  (5000000000 mod 2^32)
}
```

### Expected behavior

`5000000000` — a `giga`/`long` holds at least 64 bits, so the value round-trips and prints unchanged.

### Actual behavior

```shell
705032704
```

Both `%d` and `%lld` print the truncated 32-bit value; the true value never survived storage. (UBSan additionally flags `atoi` overflow and `int` arithmetic overflow on any large-literal or large-arithmetic `giga` expression, e.g. `ast.c:2433` for `giga` addition — the arithmetic is done in `int` too.)

### Brainrot version / commit

dcddf726a3f818c9d4e1eadc040e7a91aa9ba99f

### Operating system

WSL2 (Ubuntu), x86-64

### Additional context

Distinct from #177 (which is only about `maxxing`/`sizeof(giga)` reporting 4 on wasm32). This is native, always-present, and about actual value storage/width rather than the reported size.

[...]

Issue #194, opened 2026-08-18
Support arrays as function parameters/return values by value


leo-aa88:
## Description

Arrays currently can only cross a function boundary via a pointer
parameter (`rizz *arr`), which aliases the caller's array — there's no
grammar support for declaring an array-typed parameter (`rizz arr[]`) or
returning an array by value.

This matches C's own restriction (C doesn't support by-value array
parameters/returns either — arrays decay to pointers), so it's arguably
not a gap so much as a design choice worth confirming intentional. Filing
as a tracking issue in case there's appetite for:

- Explicit array-parameter syntax (`rizz sum(rizz arr[], rizz len)`) that's
  still pointer-decay under the hood (matching C), for readability/
  self-documentation at call sites, even if the underlying semantics stay
  aliasing rather than a copy.
- A real by-value array copy semantic (non-C, more expensive, but
  consistent with how struct-by-value now works after #191) — would need
  its own design discussion since it's a bigger semantic commitment than
  the struct case (arrays can be large; copying implicitly on every call
  has real performance implications C explicitly avoids).

## Where this lives in the code

- `param_list` grammar (lang.y) — no `dimensions` alternative, so there's
  no syntax for an array-typed parameter today.
- `enter_function_scope` (ast.c) — argument binding has no array case at
  all (falls through to the pointer_level>0 path only if the caller
  explicitly passes a pointer).

## Related

#191 added struct-by-value parameters/returns using a similar
`ArrayDimensions`/`Variable` storage model; whichever direction is chosen
for arrays could reuse some of that runtime-allocation-at-declaration-time
machinery (`interpreter_visit_declaration`).

PR #309, opened 2026-08-30
fix: reject array arguments to user-defined calls (#308)

### LINKED ISSUE #308: [Bug]: array arguments are unchecked for user-defined calls (pointer read as a scalar; UB for `cap`)
### Description

Passing an array where a scalar parameter is declared is rejected for **native**
calls but silently accepted for **user-defined** calls, where the array's
backing pointer is read as if it were the scalar's own value.

This is the union-aliasing hazard `is_unmarshallable_array_arg()`'s own comment
describes:

> arrays report the same VarType as their element (`rizz arr[2]`'s var_type is
> VAR_INT, indistinguishable from a scalar `rizz x` by
> `infer_expression_type()` alone), but passing one through would reinterpret
> its backing pointer as a scalar value.

That helper exists and is correct — it is just only wired into
`semantic_check_native_call()`. Nothing on the user-defined call path
(`semantic_visit_function_call()` → `enter_function_scope()`) consults it, so
`Variable.value`'s union aliasing of `array_data` with the scalar members goes
unchecked.

For a `cap` parameter the result is not merely a wrong number but **undefined
behavior**, caught by UBSan on the standard sanitizer build.

Found while reviewing #307 (thanks @leo-aa88). Not introduced by it — verified
on `463958a`, the commit before it — but worth its own issue because that PR
closed the same hole for `gang` arguments only, which leaves structs as the
*best*-behaved case in this gap rather than the worst.

### Minimal reproduction code

```c
rizz twice(rizz n) {
    bussin n * 2;
}

skibidi main {
    rizz a[2];
    a[0] = 41;
    a[1] = 99;
    yapping("%d", twice(a));   🚽 `a` is an ARRAY, not a rizz
    bussin 0;
}
```

The same mistake against a native is already reported correctly:

```c
skibidi main {
    rizz a[2];
    a[0] = 41;
    yapping("%d", slorp(a));
    bussin 0;
}
```

### Expected behavior

The user-defined call should be rejected the way the native call already is:

```
Error: 'twice' argument 1: int arrays cannot be passed where a scalar/string is expected at line 9
```

(A `yap[N]` argument is the documented exception — it has a real
representation and is legitimately accepted where a string is expected. Any fix
should preserve that, as `is_unmarshallable_array_arg()` already does via its
`sym->type != VAR_CHAR` condition.)

### Actual behavior

```shell
$ ./brainrot arr.brainrot
1984
exit=0

# native path, for contrast -- correctly rejected
$ ./brainrot arr_native.brainrot
Error: 'slorp' argument 1: int arrays cannot be passed where a scalar/string is expected at line 4
exit=1
```

`1984` is the low bits of the array's heap address, not a value the program
computed. It is stable within a build but changes between builds — the same
program prints `2080` when built from `463958a` — which is the clearest
evidence it is an address being reinterpreted.

Every scalar parameter type is affected, and `cap` is undefined behavior:

```shell
$ ./brainrot types.brainrot
chad: 0.00
ast.c:5036:16: runtime error: load of value 144, which is not a valid value for type '_Bool'
cap: 0
exit=0

# rizz  -> 1984   (truncated address)
# yap   -> -16    (low byte of the address)
# chad  -> 0.00
# cap   -> UBSan: load of value 144 is not a valid _Bool
```

2-D arrays behave the same (`rizz g[2][2]` passed to a `rizz` parameter prints
`1888`).

### Brainrot version / commit

`2ac321dddab95701ba65d6d4cd308a8f18ac5ace` (also reproduced on `463958a`)

### Operating system

### Additional context

**Where the check is missing.** `is_unmarshallable_array_arg()`
(`semantic_analyzer.c`) already returns true for exactly this shape
(`sym->is_array && sym->type != VAR_CHAR`). It is called from
`semantic_check_native_call()`'s `STDROT_ANY` branch, its fixed-scalar branch,
and — since #307 — its `STDROT_STRUCT` branch. It is never called for a
user-defined callee, whose parameters are checked against
`check_type_compatibility_ex()`, which takes no `is_array` and cannot see the
difference.

**Why the value is garbage rather than element 0.** Unlike the struct case
fixed in #307 (where `value.array_data` legitimately points at the first
element's bytes, so the callee got a valid-looking `pts[0]`), a scalar
parameter reads the *union member* — `value.ivalue`/`fvalue`/`bvalue` — which
aliases the `array_data` pointer itself. So the callee sees the pointer's bits
interpreted as its declared type, which for `cap` is not necessarily a valid
`_Bool` at all.

**Suggested direction.** Call `is_unmarshallable_array_arg()` on the
user-defined argument path too, so the diagnostic is identical for both callee
kinds. If a static check is awkward there, `enter_function_scope()` (`ast.c`)
is the runtime choke point every user-defined call's arguments pass through —
the same place #307's struct fix landed via
`resolve_by_value_struct_source()`.

**Testing note.** Per CONTRIBUTING.md this wants `test_cases/*.brainrot`
coverage for each affected parameter type, and the fixture should print the
call's *result* rather than discarding it — a fixture that only pins the
diagnostic would still pass an implementation that reports the error and then
binds garbage anyway.

leo-aa88:
## Description

`rizz a[2]; twice(a)` printed `1984` — the array's own heap address
reinterpreted as an int — with no diagnostic anywhere. The identical mistake
against a native has always been reported cleanly.

`enter_function_scope()` (`ast.c`) binds a scalar parameter from
`arg_values[i].ivalue` / `.fvalue` / `.bvalue`, and `Variable`'s value union
aliases those members with `array_data` — so the callee received the backing
**pointer** as its declared type. The value changes between builds (`1984`
here, `2080` on `463958a`), which is what gives it away as an address rather
than anything the program computed.

For a `cap` parameter it is **undefined behavior**, not merely a wrong number:

```
ast.c:5036:16: runtime error: load of value 144, which is not a valid value for type '_Bool'
```

### The fix

`is_unmarshallable_array_arg()` already detected this exact shape and was
simply never wired into the user-defined call path — it is only consulted by
`semantic_check_native_call()`. This reports it at the same layer a native
call does, in `semantic_visit_function_call()`'s existing per-argument loop,
next to the struct-pointer tag check #248 added there for the same
"this path checks almost nothing" reason.

**It needs its own helper rather than reusing that one**, and that is the
subtle part. `is_unmarshallable_array_arg()` deliberately exempts `VAR_CHAR`
arrays, because a `yap buf[32]` genuinely *is* marshallable for a native —
`ast_expr_to_stdrot_value()` produces a `STDROT_STRING` and
`coerce_arg_to_param()` converts it to a `STDROT_CSTRING`. A Brainrot-defined
callee has no such conversion: `enter_function_scope()` rejects `VAR_STRING`
parameters outright, and its `VAR_CHAR` case reads `.ivalue` like every other
scalar. So `yap ca[4]` passed to a `yap` parameter was broken exactly like the
rest, and sharing the helper would have let precisely that one case through.

Before / after, all four scalar element types:

```
Error: 'twice' argument 1: int arrays cannot be passed to a function -- pass an element (`ia[0]`) or its address (`&ia[0]`) at line 49
Error: 'half'  argument 1: float arrays cannot be passed to a function -- ...
Error: 'flip'  argument 1: bool arrays cannot be passed to a function -- ...
Error: 'pick'  argument 1: char arrays cannot be passed to a function -- ...
```

### Nothing that used to work now fails

A parameter can never itself be an array — `rizz sum(rizz a[2])` is a syntax
error (#194) — and array-to-pointer decay is not implemented either
(`first(a)` for a `rizz *p` parameter already reports *"Expression is not a
pointer"*). So there is no shape where passing an array identifier was
meaningful. Verified explicitly that `yap[N]` still reaches natives (the
`STDROT_CSTRING` path), and that `arr[0]` and `&arr[0]` still work.

### Interaction with #307

Struct arrays are now caught **statically**, where #307 caught them at
runtime. Both layers stay, because neither subsumes the other: the runtime
check in `resolve_by_value_struct_source()` still covers the by-value struct
sources this analyzer check does not see — copy initialization
(`gang Point a = pts;`) and struct returns (`bussin pts;`).

So `struct_array_by_value_arg_fail.brainrot` now pins the static rejection,
and a new `struct_array_copy_init_fail.brainrot` pins the runtime one. Both
print the resulting value rather than discarding it, so a regression that
reported the error and then bound the array anyway would still fail them.

## Related Issue

Fixes #308

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement
- [ ] Refactor

## Checklist

**Test coverage.** Three `test_cases` fixtures, at the lowest layer that
reaches each behavior:

| Fixture | Pins |
| --- | --- |
| `array_arg_scalar_param_fail` (new) | the original bug, all four scalar element types including the `yap` case the native helper exempts |
| `struct_array_by_value_arg_fail` (rewritten) | the static rejection for struct arrays; exits 1 without running |
| `struct_array_copy_init_fail` (new) | the #307 runtime backstop, via the copy-init path this PR's check does not cover |

Adversarial cases covered: the `yap`/`VAR_CHAR` exemption that would have
slipped through a naive fix; `yap[N]` still reaching a native; `arr[0]` and
`&arr[0]` still accepted; and both struct fixtures printing the bound value so
a report-then-bind-anyway regression fails rather than passes.

**Verification.** 473 tests pass (471 → +2). Full valgrind sweep **412/412
clean**, zero errors, zero leaks — run against a non-sanitized build, since
`make valgrind` currently cannot run an ASan binary (that is #186/#203, open
PR #202, not introduced here). `make format-check` and `make tidy` clean.

## Note for reviewers

Two things adjacent to this that I did **not** fold in:

- The user-defined call path still does no general argument type-checking —
  `twice(3.7)` or `twice(someStruct)` are unchecked. That is the "separate,
  much larger, pre-existing gap" `semantic_visit_function_call()`'s own
  comment already names, and it deserves its own issue rather than being
  smuggled in here.
- `giga`/`thicc` truncating to 32 bits (#282) is a different bug in the same
  neighbourhood and is genuinely multi-part (value union, lexer, arithmetic,
  and an ABI bump), so it is not in scope here.

### DIFF
--- a/semantic_analyzer.c
+++ b/semantic_analyzer.c
@@ -1469,6 +1469,37 @@ static bool is_unmarshallable_array_arg(ASTNode *expr,
     return sym && sym->is_array && sym->type != VAR_CHAR;
 }
 
+/* The same "is this argument an array identifier" question as
+ * is_unmarshallable_array_arg() above, WITHOUT its VAR_CHAR exemption, for
+ * callers where that exemption does not apply (issue #308).
+ *
+ * That exemption is correct where it lives: a `yap buf[32]` handed to a
+ * NATIVE has a real representation -- ast_expr_to_stdrot_value()'s
+ * VAR_CHAR/is_array case produces a STDROT_STRING, and coerce_arg_to_param()
+ * can convert that to a STDROT_CSTRING -- so a char array is genuinely
+ * marshallable there and must not be rejected.
+ *
+ * A Brainrot-defined callee has no equivalent. enter_function_scope()
+ * (ast.c) rejects VAR_STRING parameters outright, and its VAR_CHAR case
+ * reads `arg_values[i].ivalue` like every other scalar -- which, for an
+ * array, is the union slot the backing pointer occupies. So `yap buf[4]`
+ * passed to a `yap` parameter yields the low byte of an address, exactly
+ * like `rizz a[2]` to a `rizz` parameter does. Sharing the native helper
+ * here would have let precisely that one case through.
+ *
+ * Returns the symbol (so the caller can name the element type in its
+ * diagnostic) or NULL.
+ */
+static SymbolEntry *array_identifier_symbol(ASTNode *expr,
+                                            SemanticAnalyzer *analyzer)
+{
+    if (!expr || expr->type != NODE_IDENTIFIER)
+        return NULL;
+
+    SymbolEntry *sym = find_symbol(analyzer, expr->data.name);
+    return (sym && sym->is_array) ? sym : NULL;
+}
+
 /* True when `expr` has NO valid StdrotValue representation ast_expr_to_
  * stdrot_value() (stdrot.c) can honestly construct for it -- generalizes
  * is_unmarshallable_array_arg() from "one specific representation bug"
@@ -2468,6 +2499,56 @@ void *semantic_visit_function_call(Visitor *self, ASTNode *node)
                 {
                     propagate_contextual_call_type(arg->expr, param->desc.type,
                                                    param->desc.pointer_level);
+
+                    /* An ARRAY argument (issue #308). A parameter can never
+                       itself be an array -- `rizz sum(rizz a[2])` is a
+                       syntax error (#194) -- and array-to-pointer decay
+                       isn't implemented either (`first(a)` for a `rizz *p`
+                       parameter already reports "Expression is not a
+                       pointer"), so there is no shape where passing an
+                       array identifier here is meaningful. What happened
+                       instead was silence: enter_function_scope() (ast.c)
+                       binds a scalar parameter from `arg_values[i].ivalue`
+                       / `.fvalue` / `.bvalue`, and Variable's value union
+                       aliases those with `array_data`, so the callee
+                       received the array's backing POINTER reinterpreted
+                       as its declared type. `twice(a)` printed 1984 -- the
+                       low bits of a heap address, changing between builds.
+
+                       For a `cap` parameter that is undefined behavior
+                       outright, not merely a wrong number: UBSan reports
+                       "load of value 144, which is not a valid value for
+                       type '_Bool'".
+
+                       Reported here rather than at the runtime binding site
+                       so the diagnostic matches what a native call has
+                       always produced for the identical mistake (see
+                       is_unmarshallable_array_arg()'s use in
+                       semantic_check_native_call()) -- that helper existed
+                       and was correct; it was simply never wired into this
+                       path. */
+                    SymbolEntry *arr_sym =
+                        array_identifier_symbol(arg->expr, analyzer);
+                    if (arr_sym)
+                    {
+                        int line =
+                            node->line_number > 0 ? node->line_number : 1;
+                        char error_msg[MAX_BUFFER_LEN];
+                        snprintf(error_msg, sizeof(error_msg),
+                                 "'%s' argument %d: %s arrays cannot be "
+                                 "passed to a function -- pass an element "
+                                 "(`%s[0]`) or its address (`&%s[0]`)",
+                                 func_name.data, arg_index,
+                                 vartype_to_string(arr_sym->type),
+                                 arg->expr->data.name.data,
+                                 arg->expr->data.name.data);
+                        add_semantic_error(analyzer,
+                                           SEMANTIC_ERROR_TYPE_MISMATCH,
+                                           STRING_LITERAL(error_msg), line);
+                        arg = arg->next;
+                        continue;
+                    }
+
                     /* Struct/union pointer parameters: this analyzer does
                        not type-check user-defined call arguments against
                        their parameters at all otherwise (a separate,
--- a/test_cases/array_arg_scalar_param_fail.brainrot
+++ b/test_cases/array_arg_scalar_param_fail.brainrot
@@ -1,0 +1,54 @@
+🚽 Test case: issue #308 -- an array argument to a user-defined function's
+🚽 SCALAR parameter. This is the shape the issue was filed for, and the
+🚽 most dangerous of the family, because the failure was not "the callee
+🚽 gets element 0" but "the callee gets the array's own address".
+🚽
+🚽 enter_function_scope() (ast.c) binds a scalar parameter from
+🚽 arg_values[i].ivalue / .fvalue / .bvalue, and Variable's value union
+🚽 aliases those members with array_data -- so the callee received the
+🚽 backing POINTER reinterpreted as its declared type. `twice(a)` printed
+🚽 1984 on one build and 2080 on another: the low bits of a heap address,
+🚽 not anything the program computed.
+🚽
+🚽 Every scalar element type is covered here, and `cap` is the reason this
+🚽 is a correctness bug rather than a cosmetic one -- reading an arbitrary
+🚽 byte as _Bool is undefined behavior, which UBSan reported as
+🚽 "load of value 144, which is not a valid value for type '_Bool'".
+🚽
+🚽 The equivalent mistake against a NATIVE has always been reported
+🚽 (is_unmarshallable_array_arg(), semantic_analyzer.c): `slorp(a)` gives
+🚽 "int arrays cannot be passed where a scalar/string is expected". That
+🚽 helper was correct and simply never wired into the user-defined call
+🚽 path. Note the `yap` case below specifically: it is why #308 needed its
+🚽 own helper rather than reusing that one, which deliberately exempts
+🚽 VAR_CHAR arrays because a `yap[N]` IS marshallable for a native (as a
+🚽 STDROT_STRING/STDROT_CSTRING) -- a Brainrot-defined callee has no such
+🚽 conversion, so the exemption would have let exactly that case through.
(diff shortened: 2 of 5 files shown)


semantic_analyzer.c
```diff
@@ -2468,6 +2499,56 @@ void *semantic_visit_function_call(Visitor *self, ASTNode *node)
...
+                        array_identifier_symbol(arg->expr, analyzer);
+                    if (arr_sym)
+                    {
+                        int line =
+                            node->line_number > 0 ? node->line_number : 1;
+                        char error_msg[MAX_BUFFER_LEN];
+                        snprintf(error_msg, sizeof(error_msg),
+                                 "'%s' argument %d: %s arrays cannot be "
+                                 "passed to a function -- pass an element "
+                                 "(`%s[0]`) or its address (`&%s[0]`)",
```
leo-aa88, line 2540: **MINOR.** The `&%s[0]` half of this hint is unconditional, but it is only correct advice when the parameter is a pointer — and `param` is in scope one line up.

For a by-value parameter, which is every case in your own fixture, following it goes nowhere good:

```
rizz twice(rizz x)          twice(&a[0])    → "Cannot use pointer in integer context", prints 0, exit 0
chad len2(gang Point p)     len2(&pts[0])   → "Expected a by-value struct/union value...", prints 0.0, exit 0
cap  flip(cap b)            flip(&ba[0])    → prints 0. No diagnostic at all.
```

That last one is the one worth caring about. A pointer bound to a `cap` parameter, silently, wrong answer — which is the exact failure mode this PR exists to eliminate, reached by following the message this PR just added. I checked `main`: identical behavior there, so the gap is pre-existing and sits squarely inside the "general argument type-checking for user-defined calls" caveat your reviewer note already carves out. Not asking you to fix that here. But the new diagnostic actively routes people into it, and that part is one line:

```c
param->desc.pointer_level > 0
    ? "... pass its address (`&%s[0]`)"
    : "... pass an element (`%s[0]`)"
```

Second, smaller thing on the other half of the hint. `%s[0]` is 1-D-shaped. For `rizz m[2][2]`, `twice(m[0])` is *accepted* and yields `m[0][0]` — partial indexing collapses to the first element rather than producing a row. Pre-existing semantics, not something this PR changed, and the value is at least defined rather than an address. But the hint teaches a C-shaped mental model (`m[0]` is a row) that this language doesn't have, and the fixture only covers 1-D so nothing pins the difference. Worth either saying "an element" without the subscript example, or adding a 2-D line to `array_arg_scalar_param_fail` so the behavior is at least written down somewhere.

leo-aa88 (commented):
# APPROVE

This is the follow-up to the out-of-scope note from #307's round-2 review, and
it's a better fix than the one I'd have sketched. Built it, ran it, and spent
most of my time trying to get an array past the new check rather than
confirming the happy path.

## The subtle part is right, and it's right in both directions

Not reusing `is_unmarshallable_array_arg()` is the whole load-bearing decision
here, and it holds under test:

```
yap name[32]; slorp(name);          → "got skibidi"    (native path intact)
yap buf[8];   yapping("%s", buf);   → "Hi"             (variadic native intact)
yap ca[4];    pick(ca);             → rejected         (user-defined path closed)
```

A naive shared helper would have let the third one through — silently, and
`enter_function_scope()`'s `VAR_CHAR` case reads `.ivalue`, so it's the same
address-as-a-value bug as the other three. Getting the exemption boundary
right in both directions is the part that would have been easy to miss.

## What I tried to break

| Probe | Result |
| --- | --- |
| all four scalar element types | rejected, correct lines |
| `add3(1, arr, 3)` | `argument 2` — index is correct, the `continue` doesn't desync it |
| `twice((a))` — parenthesized bypass | rejected; the grammar folds parens into the identifier node |
| `rizz m[2][2]` | rejected |
| array → pointer parameter | now static, was a runtime *"Expression is not a pointer"* |
| `arr[0]`, `&arr[0]` | still work |
| `chungus s` as a parameter | syntax error — so the "no valid shape exists" claim genuinely holds; strings can't be the loophole |
| callee declared after the call | syntax error — no forward-call shape either |

**Struct arrays are strictly better off than #307 left them:** static, exit 1,
and the rewritten `struct_array_by_value_arg_fail` reports at line 47 — the
actual call — instead of baking in the closing-brace misattribution the old
runtime version had to document as known-wrong. Splitting the value-pinning
form out into `struct_array_copy_init_fail` keeps the round-3 discipline
exactly where it still applies.

I mutation-tested that new fixture rather than trusting its own comment:
deleting the `is_array` branch from `resolve_by_value_struct_source()` makes
it fail. The comment's claim ("delete the runtime check and this fixture
starts printing 3.0") is accurate, and keeping both layers is justified —
copy-init and `bussin pts;` never reach the analyzer check.

473 pass (472 + 1 deselected for the stale system install on this box),
`format-check` and `tidy` clean. I did not re-run the valgrind sweep
independently; LSan was clean across every path I exercised, which is
evidence, not the same thing.

## Two MINOR notes, both about the message rather than the check

Inline. Neither blocks.

# VERDICT

The fix wires an existing, correct helper into the one path that never called
it, and the interesting engineering was recognizing that the helper's one
documented exemption doesn't transfer — then proving that both ways round.
The diagnostic is the only thing I'd change, and only its advice, not its
trigger.

Ship it.

PR #310, opened 2026-08-30
fix: guard the three trapping integer division/modulo cases (#271, #272, #273)

### LINKED ISSUE #271: [Bug]: smol (short) modulo by zero crashes with SIGFPE — missing zero-divisor guard in OP_MOD VAR_SHORT
### Description

In `handle_binary_operation()` (`ast.c`), the `OP_MOD` case guards against a zero divisor for `VAR_INT` (emits "Modulo by zero" and falls back to 0), and the `OP_DIVIDE` case guards the zero divisor for **both** `VAR_INT` and `VAR_SHORT`. But the `OP_MOD` branch for `VAR_SHORT` performs the modulo with **no zero check**:

```c
else if (promoted_type == VAR_SHORT)
{
    *(short *)result = *(short *)left_value % *(short *)right_value;   // ast.c:2536
}
```

When both operands are `smol` (short) and the divisor is 0, this is integer division by zero → **SIGFPE / crash** (and UBSan flags it). The equivalent int expression is handled gracefully, so the behavior is inconsistent and the short path is a hard crash.

### Minimal reproduction code

```c
skibidi main {
    smol a = 10;
    smol b = 0;
    smol c = a % b;
    yapping("%d", c);
    bussin 0;
}
```

### Expected behavior

Same graceful handling as the int case: emit "Modulo by zero" via `yyerror` and produce a defined fallback (0), rather than crashing the interpreter.

### Actual behavior

```shell
ast.c:2536:53: runtime error: division by zero
AddressSanitizer:DEADLYSIGNAL
==...==ERROR: AddressSanitizer: FPE on unknown address ...
    #0 ... in handle_binary_operation ast.c:2536
    #1 ... in evaluate_expression_short ast.c:3904
    #2 ... in interpreter_visit_declaration interpreter.c:652
    ...
SUMMARY: AddressSanitizer: FPE ast.c:2536 in handle_binary_operation
```

### Brainrot version / commit

### Operating system

WSL2 (Ubuntu), x86-64

### Additional context

Fix: mirror the `VAR_INT` `OP_MOD` guard (and the short `OP_DIVIDE` guard) in the `VAR_SHORT` `OP_MOD` branch:

```c
else if (promoted_type == VAR_SHORT)
{
    if (*(short *)right_value == 0)
    {
        yyerror("Modulo by zero");
        *(short *)result = 0;
    }
    else
    {
        *(short *)result = *(short *)left_value % *(short *)right_value;
    }
}
```

### LINKED ISSUE #272: [Bug]: INT_MIN / -1 and INT_MIN % -1 crash with SIGFPE — division/modulo overflow not guarded
### Description

`handle_binary_operation()` (`ast.c`) guards integer division/modulo against a **zero** divisor, but not against the other well-known overflow case: `INT_MIN / -1` and `INT_MIN % -1`. In two's-complement the mathematical result (`2147483648`) is not representable in `int`, and on x86-64 the `idiv` instruction raises `#DE` → **SIGFPE**:

```c
*(int *)result = *(int *)left_value / *(int *)right_value;   // ast.c:2476  (OP_DIVIDE)
...
*(int *)result = left % right;                               // ast.c:2521  (OP_MOD)
```

`right != 0` is checked, but `!(left == INT_MIN && right == -1)` is not, so the divide/modulo executes and crashes.

### Minimal reproduction code

```c
skibidi main {
    rizz a = -2147483648;
    rizz b = -1;
    rizz c = a / b;   🚽 also crashes with %  (a % b)
    yapping("%d", c);
    bussin 0;
}
```

### Expected behavior

A defined result / catchable error, consistent with how divide-by-zero is handled (`yyerror` + fallback), rather than a hard SIGFPE that takes down the interpreter.

### Actual behavior

```shell
ast.c:2476:53: runtime error: division of -2147483648 by -1 cannot be represented in type 'int'
AddressSanitizer:DEADLYSIGNAL
==...==ERROR: AddressSanitizer: FPE on unknown address ...
    #0 ... in handle_binary_operation ast.c:2476
    #1 ... in evaluate_expression_int ast.c:4078
    #2 ... in interpreter_visit_declaration interpreter.c:620
    ...
SUMMARY: AddressSanitizer: FPE ast.c:2476 in handle_binary_operation
```

### Brainrot version / commit

### Operating system

WSL2 (Ubuntu), x86-64

### Additional context

Extend the existing zero guard in both the `OP_DIVIDE` and `OP_MOD` `VAR_INT` branches, e.g.:

```c
if (*(int *)right_value == 0 ||
    (*(int *)left_value == INT_MIN && *(int *)right_value == -1))
{
    yyerror("Integer division overflow");
    *(int *)result = ...;   /* defined fallback */
}
```

Related but separate: negating the literal `-2147483648` also trips UBSan (`ast.c:3011: negation of -2147483648 cannot be represented in type 'int'`) in the unary-minus path. See also #271 (short modulo-by-zero missing guard).

### LINKED ISSUE #273: [Bug]: INT_MIN / -1 and INT_MIN % -1 crash with SIGFPE (division-overflow case unguarded)
### Description

`handle_binary_operation` guards integer division and modulo against a **zero** divisor, but not against the other well-known trapping case in C: `INT_MIN / -1` (and `INT_MIN % -1`). On x86-64 the `idiv` instruction raises `#DE` for this input just as it does for a zero divisor, so the interpreter crashes hard with `SIGFPE` instead of producing a value or a clean diagnostic.

The overflow also affects the value that produces `INT_MIN` in the first place (`0 - 2147483648`), which UBSan flags separately, but the fatal fault is the division/modulo itself at `ast.c:2476` (OP_DIVIDE) and `ast.c:2521` (OP_MOD).

The zero guards live at `ast.c:2468` / `ast.c:2507`; neither is extended to reject `left == INT_MIN && right == -1`.

### Minimal reproduction code

```c
skibidi main {
    rizz a = 0 - 2147483648;   🚽 INT_MIN
    rizz b = 0 - 1;            🚽 -1
    yapping("%d", a / b);      🚽 SIGFPE — also a % b
}
```

### Expected behavior

Either a defined, non-crashing result (like the existing division-by-zero path, which reports an error and yields a fallback of `0`), or a clean interpreter error. The interpreter should never take a `SIGFPE` on attacker/user-supplied arithmetic.

### Actual behavior

```shell
ast.c:2476:53: runtime error: division of -2147483648 by -1 cannot be represented in type 'int'
AddressSanitizer:DEADLYSIGNAL
==...==ERROR: AddressSanitizer: FPE on unknown address ...
    #0 ... in handle_binary_operation ast.c:2476
    #1 ... in evaluate_expression_int ast.c:4078
    #2 ... in ast_expr_to_stdrot_value stdrot.c:1130
```

`a % b` crashes identically at `ast.c:2521`. Process dies with `SIGFPE` (exit 136 / 141 under ASan).

### Brainrot version / commit

### Operating system

WSL2 (Ubuntu), x86-64

### Additional context

The same trap applies to `giga`/`thicc` (`long`/`long long`) division where the operands are the type minimum and `-1`. A fix should mirror the existing zero-divisor guard: add `else if (left == INT_MIN && right == -1)` handling to OP_DIVIDE and OP_MOD for each signed integer promoted type. Note `INT_MIN % -1` is mathematically `0`, so the modulo fallback can be exact.

leo-aa88:
## Description

`handle_binary_operation()` crashed the interpreter with **SIGFPE** on three
inputs. All three are integer division/modulo trapping on x86-64, all three
live in the same function, and none had test coverage.

```c
smol a = 7; smol b = 0;  a % b     🚽 SIGFPE -- no guard at all
INT_MIN / -1                       🚽 SIGFPE -- the zero-divisor guard misses it
INT_MIN % -1                       🚽 SIGFPE -- same, via the shared quotient
```

### 1. `smol` modulo by zero (#271)

`OP_MOD`'s `VAR_INT` arm guarded a zero divisor, and `OP_DIVIDE` guarded it
for **both** `VAR_INT` and `VAR_SHORT` — but `OP_MOD`'s `VAR_SHORT` arm went
straight to the modulo. So `smol / 0` reported *"Division by zero"* cleanly
while `smol % 0`, one operator away, took down the process:

```
ast.c:2940:53: runtime error: division by zero
AddressSanitizer:DEADLYSIGNAL
```

Now mirrors its siblings, message and fallback included.

### 2/3. `INT_MIN / -1` and `INT_MIN % -1` (#272 — and #273, which is a duplicate of it)

A zero check is not the only trapping case. The mathematical quotient
`2147483648` is not representable in `int`, C leaves it undefined, and `idiv`
raises `#DE` for it exactly as it does for a zero divisor:

```
ast.c:2880:53: runtime error: division of -2147483648 by -1 cannot be represented in type 'int'
AddressSanitizer:DEADLYSIGNAL
```

**The two are handled differently, on purpose:**

| | Behavior | Why |
| --- | --- | --- |
| `INT_MIN / -1` | error + fallback `0` | There is no correct `int` to return, so it matches how division by zero already behaves here. |
| `INT_MIN % -1` | returns `0`, **no error** | `0` is the mathematically correct remainder and is exactly representable. It only trapped because `idiv` computes quotient and remainder together, so the unrepresentable *quotient* faults even when only the remainder was asked for. Reporting an error would pin a worse implementation. |

`VAR_SHORT` deliberately gets **no** overflow guard: C's usual arithmetic
conversions promote both operands to `int` before dividing, so `-32768 / -1`
is computed where the result *is* representable and cannot trap. The fixture
pins that too, so nobody later "fixes" a case that isn't broken.

## Related Issue

Fixes #271
Fixes #272
Fixes #273

(#273 is a duplicate of #272 — same crash, same two lines. Closing both.)

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement
- [ ] Refactor

## Checklist

**Test coverage.** Two `test_cases` fixtures:

| Fixture | Pins |
| --- | --- |
| `modulo_by_zero_smol_fail` | #271, with the `rizz` and `smol` forms side by side — one guard existing without the other is exactly the shape of the bug |
| `division_overflow_int_min_fail` | #272/#273, both operators, plus the `smol` case that must *not* be guarded |

Adversarial cases covered: both fixtures **print the resulting value** rather
than discarding it, so an implementation that reported the error and then left
the result slot uninitialized would still fail; and the modulo fixture would
catch a "fix" that made `INT_MIN % -1` report an error instead of returning
its correct answer.

Verified normal division and modulo are unaffected (`17/5`, `17%5`, negative
operands, `smol`, float `fmod`).

**Verification.** 473 tests pass (471 → +2). Full valgrind sweep **412/412
clean** — run against a non-sanitized build, since `make valgrind` cannot
currently run an ASan binary (#186/#203, open PR #202). `make format-check`
and `make tidy` clean.

## Note for reviewers

`smol` division narrowing prints `32768` for `-32768 / -1`, which does not fit
in a `short`. I verified this is **identical on `main`** — a separate,
pre-existing narrowing quirk in `OP_DIVIDE`'s `VAR_SHORT` arm, untouched by
this PR and not part of the trapping bug. Happy to file it separately if you
want it tracked.


test_cases/modulo_by_zero_smol_fail.brainrot
```diff
@@ -0,0 +1,26 @@
...
+🚽 and crashed the interpreter with SIGFPE.
+🚽
+🚽 handle_binary_operation()'s OP_MOD branch guarded VAR_INT, and OP_DIVIDE
+🚽 guarded both VAR_INT and VAR_SHORT -- but OP_MOD's VAR_SHORT arm went
+🚽 straight to the modulo. So `smol / 0` reported "Division by zero"
+🚽 cleanly while `smol % 0`, one operator away, took down the process.
+🚽
+🚽 The result is printed rather than discarded so this pins the fallback
+🚽 value too, not just the message: an implementation that reported the
+🚽 error and then left the slot uninitialized would still fail here.
```
leo-aa88: **MINOR**, but it's a false claim living in the tree, and this PR series has been about exactly that.

> an implementation that reported the error and then left the slot uninitialized would still fail here

It wouldn't. I tried it:

```c
if (*(short *)right_value == 0)
{
    yyerror("Modulo by zero");
    /* *(short *)result = 0;  <- deleted */
}
```
```
1 passed, 472 deselected
```

Same for the division-overflow branch in `division_overflow_int_min_fail` — dropped `*(int *)result = 0;`, kept the `yyerror()`, fixture still green.

The reason is `result = SAFE_MALLOC(int)` (ast.c:2814). `safe_malloc()` is plain `malloc` behind a header — it does not zero. But `safe_free()` memsets each block before releasing it, so a recycled block comes back zeroed, and a fresh one from the OS is zeroed too. An uninitialized read therefore yields the same `0` the fixture expects. Printing the value cannot distinguish "explicitly set to 0" from "never written."

Nothing is broken in the shipped code — it does set the fallback. What's wrong is the sentence, and the identical claim in the PR description's adversarial-coverage list. Two honest options:

- Say what the fixture *does* pin: the diagnostic, and that both the `rizz` and `smol` forms produce the same message and the same value. That is already worth having.
- Or make the claim true. It would need a distinguishable sentinel — the fallback would have to be something other than the value uninitialized memory reliably reads as, which for a "we have no answer" case is not obviously an improvement over `0`.

I'd take the first. `0` is the right fallback and matches how division by zero has always behaved here; the fixture just can't see the difference, and should say so.

ast.c
```diff
@@ -2875,6 +2875,21 @@ void *handle_binary_operation(ASTNode *node)
...
+            /* The OTHER trapping case, and the one a zero check alone
+               misses: INT_MIN / -1. The mathematical result (2147483648)
+               is not representable in int, C leaves it undefined, and on
+               x86-64 `idiv` raises #DE for it exactly as it does for a
+               zero divisor -- so this crashed the interpreter with SIGFPE
+               rather than producing a value or a diagnostic (#272/#273).
+               Handled like division by zero, since the situation is the
+               same one: there is no correct int to return, so say so
+               instead of trapping. */
+            else if (*(int *)left_value == INT_MIN && *(int *)right_value == -1)
```
leo-aa88, line 2905: Observation, not a request — filing it because you clearly thought about this one branch away and the asymmetry will bite whoever fixes `nonut`.

`OP_MOD`'s new guard is deliberately placed *after* the `is_unsigned` branch, with a comment explaining why ("unsigned arithmetic has no such case"). This one has no such consideration, because `OP_DIVIDE`'s `VAR_INT` arm has no `is_unsigned` branch at all. So for two `nonut rizz` values holding `0x80000000` and `0xFFFFFFFF`:

```
Error: Division overflow: the most negative rizz divided by -1 has no representable result
udiv: 0
```

The unsigned answer really is `0`, so the value is right — but the user gets a diagnostic about "the most negative rizz" for variables they declared unsigned, describing an overflow that cannot happen in unsigned arithmetic.

I chased it further before writing this up, and it's currently unreachable as a *wrong result*: `node->modifiers.is_unsigned` never seems to be set on a binary-op node at all. `nonut rizz a = 0 - 1; a % 3` gives `-1`, not the unsigned `0` — so `OP_MOD`'s own unsigned branch is dead in that shape too. Unsigned semantics simply aren't propagating, which is a much larger pre-existing thing and firmly out of scope here.

So: nothing to fix today, and I'd rather you didn't add a dead `is_unsigned` check to match another dead one. But when someone does wire unsigned through, `OP_DIVIDE` is a place that will silently produce a signed-flavored diagnostic for a well-defined unsigned operation, and this branch is where it'll come from. Worth a sentence in the comment, or a line in whatever issue tracks `nonut`.

leo-aa88 (commented):
# COMMENT

The three guards are correct and I verified each of them. Two things stop this
from being a straight approve, neither of them a defect in the guards
themselves: **the PR does not currently merge**, and one of its new in-tree
comments claims test coverage it does not have — which I demonstrated by
mutation.

## The fix itself is right, including the part that's easy to get wrong

Handling `INT_MIN / -1` and `INT_MIN % -1` *differently* is the correct call,
and the reasoning in the comment is the actual reason: the remainder is
mathematically 0 and exactly representable, so the trap is purely an artifact
of `idiv` producing quotient and remainder together. Reporting an error there
would have pinned a worse implementation. Good.

What I probed looking for a fourth trapping case:

| Probe | Result |
| --- | --- |
| all three original crashes | fixed, no SIGFPE |
| `yap` arithmetic with a zero divisor (`'A' - 'A'`) | promotes to `VAR_INT`, gets **both** guards |
| `cap` arithmetic | rejected by the analyzer — no division surface at all |
| `/=`, `%=` | not in the grammar, so no second entry point |
| literal `7 / 0`, `7 % 0` | guarded |
| `17/5`, `-17/5`, `-17%5`, `smol`, float, `gigachad % 2.0` | all unaffected |

The `VAR_SHORT`-needs-no-overflow-guard reasoning holds: `-32768 % -1` is
computed after promotion to `int` and never traps. And your reviewer note
checks out — `smol div: 32768` prints identically on `main`, so the narrowing
quirk really is pre-existing and untouched. Worth its own issue; it is a
`short` holding a value no `short` can hold.

473 pass, `format-check` and `tidy` clean.

## 1. It doesn't merge

`mergeable: CONFLICTING`. Merge base is `2ac321dd`, so this branched before
#309 landed, and both PRs append to the tail of `tests/expected_results.json`
— on top of #309 having rewritten the `struct_array_by_value_arg_fail` entry
that sits immediately above.

Mechanical, but the resolution is not purely mechanical: keep **#309's**
version of `struct_array_by_value_arg_fail` (the static
`'len2' argument 1: struct arrays cannot be passed...` form), not the
runtime-form copy that appears as context in this diff. Taking "ours" reverts
#309's fixture rewrite.

## 2. A coverage claim in the tree is false — demonstrated

Inline on `modulo_by_zero_smol_fail.brainrot`. Short version: I removed the
`*(int *)result = 0;` fallback assignment, leaving only the `yyerror()`, and
both new fixtures still pass.

## 3. One observation for later, not now

Inline on the division guard: it doesn't consult `is_unsigned`, while the
modulo guard explicitly reasons about it. Currently invisible, because
`nonut` doesn't reach binary-op nodes at all.

# VERDICT

The engineering judgment here is good — particularly refusing to make
`INT_MIN % -1` an error just because its sibling is one, and pinning the
`smol` case that must *not* be guarded so nobody "fixes" it later. What needs
doing before merge is a rebase, and a one-line correction to a comment that
overstates what the fixture proves. Given this PR series' whole theme has been
claims matching checks, that second one is worth the two minutes.

Rebase and fix the comment and I'll approve.

leo-aa88 (commented):
# APPROVE

Both asks are done, and the second one came back with something I'd missed.

**Rebase.** `MERGEABLE` now. I checked the rebase didn't smuggle anything in:
`ast.c` is byte-identical between the old `93ca93b` and the rebased `6ca16a9`,
and the three crashes are still fixed on the new head with normal arithmetic
unaffected.

**The false coverage claim.** Gone, and replaced with the actual mechanism
rather than a hedge — `SAFE_MALLOC` doesn't zero, but `safe_free()` memsets
before releasing, so an uninitialized read yields the same `0` the fixture
expects. That's the right level of detail: it explains why this *style* of
check can't work here, not just that this instance didn't.

**The part I didn't have.** Separating the two printed values by strength is a
better answer than the one I asked for:

> `mod: 0` does real work, because what this fixture pins about the modulo
> case is the ABSENCE of a diagnostic.

I verified it — adding a `yyerror("Modulo overflow")` to the `INT_MIN % -1`
branch fails `division_overflow_int_min_fail`. And that *is* the regression
most likely to be attempted here, because making it "consistent with its
sibling" looks like a tidy-up. So one printed value is load-bearing and the
other isn't, and the fixture now says which is which. That distinction is
worth more than the claim I flagged.

**The `is_unsigned` note** is recorded at the guard with the dead-branch
reasoning and the `nonut rizz a = 0 - 1; a % 3` → `-1` evidence, and
explicitly tells whoever wires unsigned through what to do. Recorded, not
"fixed" — correct.

Verified: 475 pass, `format-check` and `tidy` clean, `6303423`'s `ast.c` hunk
is comment-only.

## One observation, no action needed

These two fixtures now carry ~40 lines of comment above ~10 lines of code, and
their expectations embed `at line N`. That's twice now that editing the prose
has shifted the reported lines and required a matching `expected_results.json`
update (26→42, 44→63). It fails loudly rather than silently, so it's
self-correcting — but it is a real tax on exactly the fixtures whose comments
you most want people to keep editing.

# VERDICT

The interesting outcome is that the fix for a weak test turned out to be
*classifying* the assertions by strength rather than strengthening them.
"Printing the value proves the fallback was written" was false; "printing the
value proves no diagnostic was emitted" is true, load-bearing, and catches the
likely regression. Writing down which is which is what a fixture comment is
for.

Ship it.

Issue #311, opened 2026-08-30
[Feature]: the three limitations that still force workarounds in a real Brainrot program — struct-array fields, struct pointer arithmetic, `rant` parameters


leo-aa88:
## Problem / motivation

These three are the *complete* remaining set of language limitations that force structural workarounds in [tung-tung-sahur](https://github.com/Brainrotlang/tung-tung-sahur), the Phase 5 game. Every other workaround that game carried has been fixed upstream — #294, #296, #299, #303 — and each of those was found the same way: by trying to write a real program.

All four reproductions below were re-verified against **v0.2.0** (`2ac321d`), so none of them is stale.

Filing as one issue because they were assessed together and two of them compose into a single architectural cost; happy to split into three if that tracks better.

---

### 1. A struct field cannot be an array of structs

The boundary is specific — a *scalar* array field is fine, a *struct* array field is a parse error:

```c
gang S { rizz nums[4]; };          🚽 works: prints 7
gang E { chad x; };
gang S2 { gang E es[4]; };         🚽 Error: syntax error, unexpected LBRACKET, expecting SEMICOLON
```

`docs/ROADMAP.md` lists "fixed arrays in structs" as delivered, which is true for the scalar case only — worth a note either way.

**Cost:** there is no "God struct" to hand to a helper. Every entity pool has to be a `skibidi main` local.

### 2. No pointer arithmetic on struct pointers

```c
gang E { chad x; };
gang E pool[2];
gang E *p = &pool[0];
p = p + 1;
🚽 Error: Cannot perform pointer arithmetic on a type-erased pointer -- its pointee size is unknown
```

The indexed form fails differently:

```c
skibidi walk(gang E *p, rizz n) { yapping("%.1f", p[1].x); }
🚽 Error: Member access on a non-struct/union array element
```

**Cost:** a helper cannot walk a pool. It can only be handed one already-resolved `&pool[i]`.

### 1 + 2 together

These compose into the one structural workaround the game can't design around. #294 made it possible to extract per-entity work into `sim.brainrot`, but the *loops* stay inlined in `main` — roughly 150 lines of scroll and collision iteration that would otherwise be four functions:

```c
🚽 what has to be written, once per pool, in main
flex (rizz i = 0; i < MAX_OBS; i = i + 1) {
    edgy (obstacles[i].alive) { ent_scroll(&obstacles[i], &wd, 0.0); }
    ...
}

🚽 what it wants to be
scroll_all(&wd);
```

[...]

Issue #277, opened 2026-08-27
[Bug]: Reassigning a rant (string) variable leaks the previous string buffer


leo-aa88:
### Description

Assigning a new value to an existing `rant` (string) variable overwrites the variable's `strvalue` slot **without freeing the heap buffer it already owns**, leaking the previous string on every reassignment.

Each VAR_STRING variable owns its `.data` buffer: declaration allocates it via `safe_strdup`/`evaluate_expression_string`, and `hm_free` frees `strvalue.data` at scope teardown. But the assignment write path does a bare overwrite:

- `write_value_to_address()` — `ast.c:2945`: `*(String *)address = evaluate_expression_string(expr);`
- the sibling scalar path — `ast.c:2991`: `var->value.strvalue = evaluate_expression_string(expr);`

Neither frees the old `strvalue.data` first, so the buffer from the prior value is orphaned. In a loop this leaks once per iteration.

### Minimal reproduction code

```c
skibidi main {
    rant s = "hello";
    s = "world";          🚽 "hello" buffer is leaked here
    yapping("%s", s);
}
```

### Expected behavior

No leak. Reassigning a string frees the buffer the variable previously owned before taking ownership of the new one. Clean exit under LeakSanitizer.

### Actual behavior

```shell
world

=================================================================
==...==ERROR: LeakSanitizer: detected memory leaks

Direct leak of 24 byte(s) in 1 object(s) allocated from:
    #1 safe_malloc lib/mem.c:174
    #2 safe_strdup lib/mem.c:367
    #3 evaluate_expression_string ast.c:3775
    #4 interpreter_visit_declaration interpreter.c:665   🚽 the "hello" from the declaration
    ...
SUMMARY: AddressSanitizer: 24 byte(s) leaked in 1 allocation(s).
```

A loop that reassigns the string N times leaks N buffers.

### Brainrot version / commit

dcddf726a3f818c9d4e1eadc040e7a91aa9ba99f

### Operating system

WSL2 (Ubuntu), x86-64

### Additional context

A plain single declaration (`rant s = "hello";` with no reassignment) does **not** leak — `hm_free` cleans it up — which confirms the leak is specific to the overwrite-on-assignment path, not declaration.

[...]

PR #314, opened 2026-08-30
feat: rant (string) parameters on user-defined functions (#311)

leo-aa88:
## Description

First of the three limitations in #311, taken in the order the issue proposes
(smallest and most mechanical first). **`rant` parameters only** — struct-array
fields and struct pointer arithmetic will follow as separate PRs.

`skibidi show(rant s)` died on *"String parameters are not supported"*, raised
from two sites in `enter_function_scope()`: the argument-evaluation pass and
the parameter-binding pass. Natives have always taken strings
(`rl_draw_text("...", ...)` works), so this was a user-function gap rather than
an ABI one. Its cost in the Phase 5 game was fifteen near-identical loader
functions differing only in one string literal, because the path could not be a
parameter:

```c
rizz load_sfx(rant path) {
    yapping("loading %s", path);
    bussin 1;
}

load_sfx("assets/sfx/jump.ogg");
load_sfx("assets/sfx/bruh.ogg");
```

### Semantics: a copy, not an alias

The issue proposed aliasing the caller's buffer as the conservative C
`const char *` reading. I went with a copy instead, because aliasing is unsafe
here for a reason C does not have: `exit_scope()` tears down the callee's
scope, so a parameter pointing into the caller's storage would make teardown's
ownership depend on where the argument came from. A copy makes the parameter an
ordinary local `rant` — assigning to it does not disturb the caller.

### The binding transfers rather than re-copies, and that's correctness

The frame's already-evaluated `safe_strdup`'d buffer is handed straight to the
parameter. That mirrors how `initialize_variable_from_expr()` binds a `rant`
**local** (`var->value.strvalue = evaluate_expression_string(expr)`), so a
parameter ends up the same shape as one the callee declared itself.

Routing it through `set_string_variable()` instead does *not* work, and this is
worth recording: `hm_free()` releases a `VAR_STRING` variable with `SAFE_FREE`,
which is right for a `safe_malloc`'d buffer and wrong for `set_variable()`'s
`ARENA_STRDUP` — it printed *"Attempt to free invalid/corrupted pointer"* at
every scope exit. That allocator mismatch is **pre-existing** in
`set_variable()`'s own `VAR_STRING` case; this path simply does not go through
it, and I left it alone rather than widen the PR.

Ownership is tracked with a second array rather than folded into
`arg_owns_blob`, because the two are released with different allocators —
plain `free()` for `calloc`'d struct temporaries, `SAFE_FREE` for string
buffers — and one flag could not say which. Every existing early-exit path that
frees blobs now frees strings too.

## Related Issue

Part of #311 (the `rant` parameters item; the other two follow separately).

## Type of Change

- [x] Documentation update
- [ ] Performance improvement
- [ ] Refactor

## Checklist

**Test coverage.** `test_cases/rant_parameter.brainrot` covers literal
arguments, variable arguments, several `rant` parameters in one call alongside
a scalar, an empty string, a returned `rant`, and a parameter passed straight
through to a recursive call.

**One case is deliberately absent, and the fixture says so rather than leaving
it implicit:** a callee that *assigns* to its parameter. That is the single
case distinguishing copy from alias — but reassigning **any** `rant` leaks its
previous buffer (#277), and a plain local does it too
(`rant s = "hi"; s = "bye";` leaks 24 bytes on `main` today), so a fixture
containing one would fail `make valgrind` for a pre-existing reason unrelated
to this change. I verified it by hand instead: with
`mutate(rant s) { s = "x"; }` the caller's variable reads back unchanged and
the only leak is #277's. The fixture records that the case should be added once
#277 is fixed.

Also checked by hand and leak-clean: an `int` passed where a `rant` is declared
(reports *"Invalid string expression"* and continues, matching this family's
convention), and a `rant` passed where a `rizz` is declared (unchecked — the
known general gap, noted in the docs).

**Verification.** 476 tests pass (475 → +1). Full valgrind sweep **415/415
clean**, run against a non-sanitized build since `make valgrind` cannot
currently run an ASan binary (#186/#203, open PR #202). `make format-check` and
`make tidy` clean.

**Docs.** `docs/the-brainrot-programming-language.md` §7.7 gains a *Parameter
types* subsection and a *Current Limitations* list — the new `rant` capability
and copy semantics, plus two limitations that were previously undocumented: no
array arguments (per #308/#309) and #277's reassignment leak.

### DIFF
--- a/ast.c
+++ b/ast.c
@@ -7600,17 +7600,27 @@ bool enter_function_scope(Function *func, ArgumentList *args)
                family's looser report-and-continue convention.
                The distinction is worth the extra check: a NULL-buffered
                `rant` is a REPRESENTABLE INVALID STATE, not merely a
-               wrong value. No other path in the language can produce
-               one -- `rant s = 42;` is a static type error -- so binding
-               it would make a `rant` parameter the one `rant` that can
-               be something an ordinary local cannot, contradicting this
-               feature's own documented "behaves as an ordinary local
-               `rant`" contract. It also propagates: through a nested
-               call, and out through `bussin s` into a caller's `rant`.
-               And printing it is undefined behavior (C11 7.21.6.1p8) --
-               glibc and BSD libc print "(null)", musl does not, which is
+               wrong value. It propagates -- through a nested call, and
+               out through `bussin s` into a caller's `rant` -- and
+               printing it is undefined behavior (C11 7.21.6.1p8): glibc
+               and BSD libc print "(null)", musl does not, which is
                exactly the kind of thing that looks benign in CI and
-               isn't (PR #314 review).
+               isn't.
+               What this check does NOT do is make that state
+               unreachable, and an earlier version of this comment
+               wrongly claimed it did (PR #314 review). A NULL-buffered
+               `rant` is reachable on main today with no `rant` parameter
+               involved at all -- any call REFUSED mid-argument leaves a
+               `rant` declaration target NULL, e.g. `rant r = f(1, 2);`
+               (arity mismatch) or `rant r = g(notastruct);` (the
+               VAR_STRUCT arm below refusing). Both verified. That is a
+               pre-existing property of what a refused call leaves
+               behind, and closing it means changing every failed call,
+               not this one. The correct, narrower justification stands
+               on its own: this feature opened a NEW route to that state
+               -- a bound parameter, which unlike a refused call's
+               leftovers is then USED -- and this closes the route it
+               opened.
                evaluate_expression_string() has already emitted exactly
                one diagnostic on both of its NULL paths, so this adds no
                second error -- the same discipline as the
--- a/ast.c
+++ b/ast.c
@@ -7487,11 +7487,32 @@ static void free_owned_struct_arg_blobs(const Value *arg_values,
             free((void *)arg_values[k].pvalue);
 }
 
+/* The same job for `rant` arguments (#311). evaluate_expression_string()
+ * always hands back a safe_strdup'd buffer the caller owns -- a literal, a
+ * variable's contents, a call result, all of them -- and the binding loop
+ * then ARENA_STRDUPs it into the parameter, so the intermediate has to be
+ * released or every string argument leaks one buffer per call.
+ *
+ * Kept as a separate array from arg_owns_blob rather than folded into it,
+ * even though a slot can only ever be one or the other (the two live in
+ * the same Value union, so a slot holding a blob pointer cannot also hold
+ * a String): the two are freed with different allocators -- plain free()
+ * for the calloc'd struct temporaries, SAFE_FREE for safe_malloc'd string
+ * buffers -- and a single flag could not say which. */
+static void free_owned_string_args(Value *arg_values, const bool *owns,
+                                   int count)
+{
+    for (int k = 0; k < count; k++)
+        if (owns[k])
+            SAFE_FREE(arg_values[k].strvalue.data);
+}
+
 bool enter_function_scope(Function *func, ArgumentList *args)
 {
     ArgumentList *curr_arg = args;
     Value arg_values[MAX_ARGUMENTS];
     bool arg_owns_blob[MAX_ARGUMENTS] = {false};
+    bool arg_owns_string[MAX_ARGUMENTS] = {false};
     int arg_count = 0;
 
     /* Snapshot the parameters in call order into a local array. The parser
@@ -7555,9 +7576,24 @@ bool enter_function_scope(Function *func, ArgumentList *args)
                 evaluate_expression_short(curr_arg->expr);
             break;
         case VAR_STRING:
-            yyerror("String parameters are not supported");
-            free_owned_struct_arg_blobs(arg_values, arg_owns_blob, arg_count);
-            return false;
+            /* `rant s` parameters (#311). Natives have always taken
+               strings; this was the user-function gap.
+
+               evaluate_expression_string() returns a safe_strdup'd buffer
+               this frame owns, so it is recorded for release below --
+               NOT aliased into the parameter. Aliasing the caller's
+               buffer would be the C `const char *` semantic the issue
+               proposed, but it is unsafe here for a reason C does not
+               have: exit_scope() tears down the callee's scope, and a
+               parameter pointing into the caller's storage would make
+               that teardown's ownership depend on where the argument
+               came from. Copying makes the parameter an ordinary local
+               `rant` -- indistinguishable from one the callee declared
+               itself -- which is also what makes assigning to it safe. */
+            arg_values[arg_count].strvalue =
+                evaluate_expression_string(curr_arg->expr);
+            arg_owns_string[arg_count] = true;
+            break;
         case VAR_STRUCT:
         {
             /* By-value struct argument. Resolve the source blob now, while
@@ -7597,6 +7633,8 @@ bool enter_function_scope(Function *func, ArgumentList *args)
                     free_pending_return_value();
                     free_owned_struct_arg_blobs(arg_values, arg_owns_blob,
                                                 arg_count);
+                    free_owned_string_args(arg_values, arg_owns_string,
+                                           arg_count);
                     return false;
                 }
                 if (!current_return_value.desc.struct_name.data ||
@@ -7609,6 +7647,8 @@ bool enter_function_scope(Function *func, ArgumentList *args)
                     free_pending_return_value();
                     free_owned_struct_arg_blobs(arg_values, arg_owns_blob,
                                                 arg_count);
+                    free_owned_string_args(arg_values, arg_owns_string,
+                                           arg_count);
                     return false;
                 }
                 StructDef *cdef = get_struct_def(curr_param->desc.struct_name);
@@ -7629,6 +7669,7 @@ bool enter_function_scope(Function *func, ArgumentList *args)
             {
                 free_owned_struct_arg_blobs(arg_values, arg_owns_blob,
                                             arg_count);
+                free_owned_string_args(arg_values, arg_owns_string, arg_count);
                 return false;
             }
             if (!src_tag.data || !curr_param->desc.struct_name.data ||
@@ -7639,6 +7680,7 @@ bool enter_function_scope(Function *func, ArgumentList *args)
                     free(src_blob);
                 free_owned_struct_arg_blobs(arg_values, arg_owns_blob,
                                             arg_count);
+                free_owned_string_args(arg_values, arg_owns_string, arg_count);
                 return false;
             }
             /* `take(make_outer().inner, ...)` resolves to an OWNED copy for
@@ -7670,6 +7712,7 @@ bool enter_function_scope(Function *func, ArgumentList *args)
     {
         yyerror("Mismatched number of arguments and parameters");
         free_owned_struct_arg_blobs(arg_values, arg_owns_blob, arg_count);
+        free_owned_string_args(arg_values, arg_owns_string, arg_count);
         return false;
     }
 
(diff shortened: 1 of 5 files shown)


ast.c
```diff
@@ -7395,9 +7416,24 @@ bool enter_function_scope(Function *func, ArgumentList *args)
...
+               proposed, but it is unsafe here for a reason C does not
+               have: exit_scope() tears down the callee's scope, and a
+               parameter pointing into the caller's storage would make
+               that teardown's ownership depend on where the argument
+               came from. Copying makes the parameter an ordinary local
+               `rant` -- indistinguishable from one the callee declared
+               itself -- which is also what makes assigning to it safe. */
+            arg_values[arg_count].strvalue =
+                evaluate_expression_string(curr_arg->expr);
+            arg_owns_string[arg_count] = true;
```
leo-aa88, line 7635: **MAJOR.** This binds the parameter unconditionally, including when `evaluate_expression_string()` returned `{NULL, 0}`.

```
skibidi show(rant s) { yapping("got [%s]", s); }
skibidi main { show(42); yapping("survived"); bussin 0; }
```
```
Error: Invalid string expression at line 8
got [(null)]
survived
exit=0
```

The callee gets a `rant` with a NULL buffer, and it propagates — through a nested call to another `rant` parameter, and out through `bussin s` into a caller's `rant r`. All of it prints `(null)` and none of it crashes, so this is not a segfault report. It's an invariant report.

**No other path in the language can produce that value.** The analogous declaration is rejected statically:

```
rant s = 42;
→ Error: Type mismatch in initialization of 's': expected string, got int
```

Same on `main`. So a `rant` local always has a real buffer; after this PR a `rant` *parameter* doesn't have to. That directly contradicts the doc paragraph this PR adds:

> A `rant` parameter takes its **own copy** of the string, so it behaves as an ordinary local `rant`.

It behaves as one except in the case where it can be something an ordinary local cannot.

You already identified this case — the description lists it under "checked by hand" as *"reports 'Invalid string expression' and continues, matching this family's convention."* The convention is real, but it isn't uniform, and the counter-example is in the same switch statement: `VAR_STRUCT`, a few lines down, does **not** bind-and-continue when its source won't resolve. It frees what it owns and `return false`s, precisely so a malformed value never reaches the callee. That is the sibling worth matching here:

```c
arg_values[arg_count].strvalue =
    evaluate_expression_string(curr_arg->expr);
if (!arg_values[arg_count].strvalue.data)
{
    free_owned_struct_arg_blobs(arg_values, arg_owns_blob, arg_count);
    free_owned_string_args(arg_values, arg_owns_string, arg_count);
    return false;
}
arg_owns_string[arg_count] = true;
```

`evaluate_expression_string()` has already emitted exactly one diagnostic on both of its NULL paths, so this adds no second error — same discipline as the `resolve_by_value_struct_source()` call below it.

Two smaller notes while you're here:

- `printf("%s", NULL)` is undefined per C11 7.21.6.1p8. glibc and BSD libc both print `(null)`, which is why this looks benign here and in CI — musl does not. `docs/brainray.md` advertises macOS support, so the tolerant-libc assumption is at least worth not relying on.
- The "Current Limitations" list you added is honest about the general argument-type gap, and I'd keep it. This isn't a request to close that gap — it's that the NULL `rant` is a *representable invalid state* rather than merely an unchecked wrong value, and those are worth refusing even inside a gap you're not otherwise closing.

leo-aa88 (commented):
# COMMENT

Built it, ran it, and went after the two things that could actually hurt here:
whether the "copy, not alias" claim survives contact, and whether the ownership
transfer is real. Both hold. One finding, on a case the PR already anticipated
but resolved the wrong way round.

## The ownership story checks out

I traced `evaluate_expression_string()` before trusting the claim that it
always hands back an owned buffer — literal, identifier, and call-result
branches all `safe_strdup`, and the two failure branches return `{NULL, 0}`.
So the frame really does own every buffer it records, and the transfer to the
parameter is a transfer rather than an alias.

Verified behaviourally:

```
rant path = "caller-owned";
show(path);        → callee: caller-owned
                     caller after: caller-owned    (survives, ASan clean)
```

Parameter named identically to a caller variable doesn't clobber it — the
parameter is `add_variable_to_scope`'d into the fresh scope before the switch
runs, so `get_variable()` can only find the new one. Recursion with a `rant`
carried through, several `rant`s alongside a scalar, an empty string, and a
returned `rant` are all clean.

**The leak claim is accurate.** `rant s = "hi"; s = "bye";` leaks 24 bytes
identically on `main` and on this branch, and the parameter-mutation case leaks
by the same mechanism — the buffer this PR allocates simply becomes the
"previous" one. That's #277, not this. Choosing to document the absent fixture
rather than ship one that fails `make valgrind` for an unrelated reason is the
right call.

Choosing a copy over the alias the issue proposed is also right, and for the
reason given: `exit_scope()` makes ownership depend on argument provenance
otherwise. 476 pass, `format-check` and `tidy` clean.

## The finding

Inline. `show(42)` binds a `rant` parameter whose `.data` is `NULL` —
a value no other construction path in the language can produce, and one the
docs added by this same PR say cannot exist.

# VERDICT

The mechanism is sound and the ownership reasoning is the part I'd have been
most worried about, so it's good that it's the part that's most carefully
argued. What's off is one failure path that reports and then binds a malformed
value, where the sibling case three lines down in the same switch already
knows to refuse instead. One line.

ast.c
```diff
@@ -7395,9 +7416,54 @@ bool enter_function_scope(Function *func, ArgumentList *args)
...
+                evaluate_expression_string(curr_arg->expr);
+            /* A NULL buffer means the argument was not a string at all
+               (`show(42)`) or the copy failed. Refuse the call rather
+               than bind it, matching the VAR_STRUCT arm below -- which
+               also frees what it owns and returns false rather than
+               handing the callee something malformed -- instead of this
+               family's looser report-and-continue convention.
+               The distinction is worth the extra check: a NULL-buffered
+               `rant` is a REPRESENTABLE INVALID STATE, not merely a
+               wrong value. No other path in the language can produce
```
leo-aa88: **MAJOR** — the change is correct, this justification for it is not.

> a NULL-buffered `rant` is a REPRESENTABLE INVALID STATE, not merely a wrong value. No other path in the language can produce one -- `rant s = 42;` is a static type error -- so binding it would make a `rant` parameter the one `rant` that can be something an ordinary local cannot

The first sentence is right and is a good reason to refuse. The second is false. `rant s = 42;` is indeed rejected statically — but that isn't the only way to initialize a `rant`, and **any refused call already produces exactly this state on `main`**, with no `rant` parameter anywhere:

```c
rant f() { bussin "x"; }
skibidi main {
    rant r = f(1, 2);          🚽 arity mismatch
    yapping("r=[%s]", r);
}
```
```
Error: Mismatched number of arguments and parameters at line 6
r=[(null)]
```

Same on `main` via a struct-argument refusal:

```c
gang P { rizz v; };
rant g(gang P p) { bussin "x"; }
...
rant r = g(notastruct);        → r=[(null)]
```

And it's still reachable *on this branch*, one hop past the fix:

```c
rant give(rant s) { bussin s; }
rant r = give(42);             → r=[(null)]
```

So an ordinary local `rant` can be something this comment says only a parameter could be, and the invariant isn't established by this change.

That doesn't make the fix wrong — this PR opened a route to a malformed value and closes it, and matching the `VAR_STRUCT` arm is the right precedent. It makes the *stated reason* wrong, which in this file is the part people read.

One more thing worth separating, because the commit message leans on it: the `VAR_STRUCT` comparison isn't equivalent.

> `gang P out = mk(notastruct);` prints `out.v = 0` and continues on main today, so leaving the declaration target zeroed is consistent rather than novel

A zeroed struct blob is a *valid* struct value — every field reads as 0, and nothing downstream can tell it apart from one you wrote yourself. A zeroed `String` is precisely the invalid state this comment is about. "Zeroed" means something different for the two, so the sibling's behaviour doesn't license this one's.

Suggested rewording: say that this closes the route the feature opens, that `evaluate_expression_string()`'s NULL result has no valid binding, and that a refused call in a `rant` initializer is a separate pre-existing hole (worth its own issue, since it's a wider question about what every failed call leaves behind — which the commit message already identifies).

docs/the-brainrot-programming-language.md
```diff
@@ -342,6 +342,42 @@ cap isPrime = is_prime(11)
...
+  array type, and array-to-pointer decay is not implemented. Pass an element
+  (`f(arr[0])`) or its address (`f(&arr[0])`).
+- Reassigning a `rant` — parameter or local — leaks its previous buffer
+  ([#277](https://github.com/Brainrotlang/brainrot/issues/277)).
+- Argument types are only checked against parameter types in specific cases
+  (arrays, struct/union tags); a general type check for user-defined calls
+  does not exist yet, so passing a `rizz` where a `chad` is declared is not
+  diagnosed. A non-string argument to a `rant` parameter *is* refused, though
+  — the call does not run — because binding one would give the parameter a
+  state no ordinary `rant` can have.
```
leo-aa88: Same correction as my `ast.c` comment, and it matters more here because this is the user-facing reference.

> because binding one would give the parameter a state no ordinary `rant` can have

An ordinary `rant` can have it:

```c
rant f() { bussin "x"; }
rant r = f(1, 2);      🚽 → r=[(null)], on main today
```

The behaviour this documents is correct — a non-string argument to a `rant` parameter *is* refused and the call doesn't run. It's the reason given that overstates.

Since this section is a "Current Limitations" list, the honest version is arguably stronger than what's there: the refusal is worth documenting, *and* the fact that a refused call leaves a `rant` target with no buffer belongs in the same list as a known limitation. A reader who hits `r=[(null)]` after an arity mistake will go looking for exactly that entry.

leo-aa88 (commented):
# COMMENT

The fix is the right one and I verified it does what it says. One load-bearing
claim underneath it is false, in both a code comment and the user-facing docs.

## Verified

```
show(42)                      → error, callee does NOT run, "survived"
outer(42) → inner(s)          → propagation gone
two("first-is-valid", 42)     → refused, no leak (arg 1's buffer released)
```

That last one is the case the new early `return false` actually exercises —
a slot that already owns a `safe_strdup`'d buffer when a later argument fails
— and `free_owned_string_args()` handles it. ASan/LSan clean.

Legitimate shapes all unaffected: literal and variable arguments, several
`rant`s alongside a scalar, empty string, returned `rant`, recursion carrying
one through.

**Mutation-verified the new fixture.** Removed the guard, kept the diagnostic:

```
FAILED test_brainrot.py::test_brainrot_examples[rant_parameter_invalid_arg_fail-...]
```

So it genuinely pins the refusal by asserting the callee's output is absent,
rather than pinning the message. 477 pass, `format-check` and `tidy` clean.

Matching the `VAR_STRUCT` arm rather than the looser report-and-continue
convention is the right call, and adding no second diagnostic is right too.

## The finding

Inline, on `ast.c:7443` and the doc line it mirrors. Short version: a
NULL-buffered `rant` is reachable on `main` today without any `rant` parameter
at all —

```
rant f() { bussin "x"; }
rant r = f(1, 2);        → r=[(null)]        (arity mismatch)
rant r = g(notastruct);  → r=[(null)]        (struct-arg refusal)
```

— so "No other path in the language can produce one" isn't true, and the fix
does not establish the invariant it claims to. It closes the route this PR
opened, which is the correct scope. The sentence just needs to say that.

# VERDICT

Right change, wrong reason written next to it. The refusal is justified on its
own terms — this PR opened a route to a malformed value and closes it, and the
`VAR_STRUCT` sibling is the correct precedent. What it can't claim is that the
state is now unreachable, because it was reachable before this PR existed and
still is. Correct the two sentences, file the pre-existing route, and this is
good to merge.

leo-aa88 (commented):
# APPROVE

The correction is accurate, and it went further than I asked in the right
direction.

**Verified the claim rather than the wording.** `c99adc7`'s `ast.c` change is
comment-only — zero code lines touched — and both counter-examples still
reproduce on this head:

```
show(42)             → refused, callee does not run
rant r = f(1, 2);    → r=[(null)]        (the pre-existing route, now documented)
```

The new comment says the right narrower thing: this feature opened a route to
a NULL-buffered `rant` that is *then used* by the callee — as distinct from a
refused call's leftovers, which sit in a declaration target — and it closes
the route it opened. "Both verified" in that comment is a claim I can confirm.

**The docs change is better than the one I suggested.** I proposed adding the
pre-existing hole to the limitations list; putting it inline as a parenthetical
next to the refusal is stronger, because a reader who hits `r=[(null)]` after
an arity mistake finds the explanation at the exact place they'd look for it,
rather than three bullets away.

479 pass, `format-check` and `tidy` clean. Already rebased onto merged `main`
and `MERGEABLE` — the CONFLICTING flag I saw a few minutes ago was GitHub
mid-update, not a real state.

# VERDICT

The pattern across this PR is worth naming: the fix was right on the first
pass and the *reason* attached to it was wrong, and the second pass changed
only the reason. Resisting the temptation to widen the code to match an
overclaimed comment — and instead narrowing the comment to match the code — is
the correct direction, and it left the pre-existing hole visible for its own
issue instead of half-closed here.

Ship it.

PR #315, opened 2026-08-30
feat: struct/union-typed array fields (#311)

leo-aa88:
## Description

Second of the three limitations in #311. **Struct-array fields only** —
`rant` parameters are #314; struct pointer arithmetic follows separately.

`gang Pool { gang Entity es[8]; };` was a parse error — *"unexpected LBRACKET,
expecting SEMICOLON"*. A scalar array field (`rizz nums[4];`) and an array of
structs as a variable (`gang Point pts[3];`) each already worked; their
intersection had nowhere to go.

### Grammar

A `struct_or_union name_token declarator dimensions SEMICOLON` production,
mirroring the by-value nested-field rule beside it. Both of that rule's checks
still apply and are kept — a struct containing an **array** of itself by value
is exactly as infinitely-sized as one containing a single self-embedded field,
and an array of an undefined tag has no element size. An array of pointers to
itself (`gang Node *kids[2];`) stays legal.

**No new layout code was needed**, which is the nice part.
`get_struct_field_size()` already takes a `VAR_STRUCT` field's element size
from the nested definition's `total_size` and multiplies by the dimension
product when `is_array` is set, and `get_struct_field_alignment()` already
returns the nested alignment while deliberately ignoring `is_array`. Both were
written for the two separate cases; their intersection was already correct.

### Access

`pool.es[i].x`. `resolve_struct_access()` handled a member access on a
*name-based* array element (`pts[i].x`) but not on a struct-typed array
**field** element — its own comment recorded that shape as unsupported and
falling through to the error. This adds that branch.

It deliberately does **not** call `evaluate_struct_field_array_access()`, which
computes the same address for a scalar element. That function reports and
`exit()`s on failure, which is right for an evaluation context and wrong here:
`resolve_struct_access()` is also called speculatively with
`report_errors == false` (`create_struct_field_array_access_node()`,
`infer_expression_struct_name()`), and a speculative probe of a malformed
access must return false rather than terminate the interpreter. The arithmetic
goes through the same two helpers, so bounds checking
(`calculate_array_offset`) and struct-aware striding (`array_element_address`)
are inherited rather than reimplemented. The rank check is duplicated for the
reason that function documents: an array field is inline in the enclosing
blob, so a dimension mismatch resolves into a *neighbouring* field's bytes.

### The motivating case

```c
skibidi scroll_all(gang Pool *p) {
    flex (rizz i = 0; i < 3; i = i + 1) {
        edgy (p.es[i].alive) { p.es[i].x = p.es[i].x - p.speed; }
    }
}
```

That loop previously had to be inlined at every call site, because there was
no struct to hand over.

## Related Issue

Part of #311 (the struct-array-fields item).

## Type of Change

- [x] Documentation update
- [ ] Performance improvement
- [ ] Refactor

## Checklist

**Test coverage.** `test_cases/struct_array_field_of_structs.brainrot` — a pool
walked by a helper through a struct pointer, a 2-D struct array field, and a
struct-array field nested two levels down beside a by-value field of the *same*
element type (so confusing the two changes the total size).

Every struct in it has a matching C declaration in
`tests/abi/struct_layout_abi_check.c` with `_Static_assert`/`offsetof`, per
that file's own contract — so 44/28/16 are a real compiler's numbers, not this
implementation agreeing with itself. Each struct also keeps a scalar field
**after** its array and checks it, because a wrong stride corrupts the
following field rather than merely misindexing the array.

Rejections verified by hand: self-embedded array by value, unknown element tag,
out-of-bounds index, wrong rank, and a self-referential *pointer* array still
accepted.

**A note on the fixture, because it is worth flagging rather than burying:** an
earlier revision of this branch created the new fixture at
`struct_array_field.brainrot`, which **already existed** (#206 3e, the
scalar-element case). That silently replaced it, and the suite stayed green
because I regenerated its expected entry to match — the old coverage vanished
without a single test failing. It is restored byte-identically here (`git diff`
against `main` for that path is empty) and the new case lives in its own file.

**Verification.** 476 tests pass (475 → +1). Full valgrind sweep clean.
`make format-check` and `make tidy` clean.

**Docs.** ROADMAP's "fixed arrays in structs" claim was true for *scalar*
element types only — the issue pointed this out — and now says so. The language
reference documents the field form and its restrictions.


ast.c
```diff
@@ -906,6 +906,93 @@ bool resolve_struct_access(ASTNode *node, StructDef **def_out, void **base_out,
             parent_base = outer_field_addr;
         }
     }
+    else if (obj && obj->type == NODE_ARRAY_ACCESS && obj->data.array.base)
+    {
+        /* Member access on an element of a struct/union-typed ARRAY FIELD
```
leo-aa88, line 911: **BLOCKING.** This branch teaches `resolve_struct_access()` about a struct-array field, which is correct and needed. Its sibling `resolve_by_value_struct_source()` — same file, same two node types, the function every by-value struct *source* goes through — didn't get the same treatment, and the new syntax makes both of its gaps reachable for the first time.

**Half one: the whole array is accepted.**

```c
gang Bag { gang Vec2 pts[2]; rizz n; };
...
len2(b.pts)     → 25.0        (silently pts[0], no diagnostic)
vec2_len2(b.pts) → 25.0       (same, through STDROT_STRUCT into a C library)
```

`resolve_by_value_struct_source()`'s `NODE_STRUCT_ACCESS` arm checks `fld->desc.type != VAR_STRUCT` and `fld->desc.pointer_level > 0` — but not `fld->desc.is_array`. #307 added that guard, with a comment explaining this exact failure, to the `NODE_IDENTIFIER` arm **only**, because a struct-array field could not be declared at the time. Now it can. #309's static check doesn't help either: `array_identifier_symbol()` is `NODE_IDENTIFIER`-only, and `b.pts` is a `NODE_STRUCT_ACCESS`.

```c
if (fld->desc.type != VAR_STRUCT || fld->desc.pointer_level > 0)
    ...
if (fld->desc.is_array)          /* <- missing */
{
    if (report_errors)
        yyerror("Expected a by-value struct/union value, got an "
                "array (index it, e.g. `arr[0]`)");
    if (struct_access_executes_call(expr))
        free_pending_return_value();
    return false;
}
```

**Half two: the element is rejected.**

```c
len2(b.pts[0])
→ Error: Expected a by-value struct/union value (a struct variable or a
         struct member access) at line 10
  0.0
```

`b.pts[0]` is a `NODE_ARRAY_ACCESS` whose `data.array.base` is set and whose `data.array.name` is not — the field form. That arm requires `expr->data.array.name.data`, so the field form falls straight through to the catch-all. It needs the same `base`/`name` split you just added above: resolve the base through `resolve_struct_access()`, take `arr_fld->offset`, and go through `calculate_array_offset()` / `array_element_address()` exactly as this branch does.

That second half is what makes this blocking rather than merely a missing guard. `f(pool.es[i])` is the natural thing to write once `pool.es[i].x` works — the motivating example in your own description is a loop over exactly that array — and right now it silently yields zeros in a Brainrot callee and `exit(1)`s against a native.

One structural suggestion, since this is the third time this shape has come back (#307, #309, now): `resolve_by_value_struct_source()` and `resolve_struct_access()` both dispatch on `NODE_IDENTIFIER` / `NODE_STRUCT_ACCESS` / `NODE_ARRAY_ACCESS`, and it's the *divergence* between their two dispatches that has produced all three bugs. Whatever the fix looks like, it's worth a comment at the top of one pointing at the other, so the next person adding an expression shape knows there are two lists to update.

test_cases/struct_array_field_of_structs.brainrot
```diff
@@ -0,0 +1,118 @@
...
+}
+
+skibidi main {
+    gang Pool pool;
+    yapping("sizeof Pool: %lu", maxxing(pool));
+
+    pool.speed = 1.5;
+    pool.count = 3;
+    flex (rizz i = 0; i < 3; i = i + 1) {
+        pool.es[i].x = 10.0;
```
leo-aa88, line 132: The fixture reads through the new syntax thoroughly — `pool.es[i].x` under a pointer, 2-D, nested two levels, a scalar after the array so a bad stride shows up. That part is good, and pairing each struct with a `_Static_assert`/`offsetof` declaration in `tests/abi/struct_layout_abi_check.c` is the right standard.

What it never does is *pass* one of these things. There's no `f(pool.es[i])` and no `f(pool.es)` anywhere in it, which is why the suite is green while both are broken (see my comment on `ast.c`).

Two lines would have caught it:

```c
    yapping("%.1f", len2(pool.es[0]));   🚽 must work
    yapping("%.1f", len2(pool.es));      🚽 must be rejected
```

Worth adding both once the `ast.c` fix lands — the second one in particular, since a rejection is the sort of thing that quietly stops being enforced.

Unrelated, and credit where it's due: flagging in the description that an earlier revision of this branch silently overwrote the existing `struct_array_field.brainrot` and that the suite stayed green because the expected entry was regenerated to match — that's the failure mode nobody reports on themselves. Restoring it byte-identically and verifying with an empty `git diff` against `main` is the right resolution.

leo-aa88 (commented):
# REQUEST CHANGES

The grammar work is right and the layout reasoning is right — `get_struct_field_size()` and `get_struct_field_alignment()` really did already compose, `maxxing` agrees (`{Vec2[2]; rizz;}` → 20), and `make abi-check` compiles, so 44/28/16 are a real compiler's numbers. Copy independence and by-value passing of the *enclosing* struct both work and are ASan-clean. 476 pass, `format-check` and `tidy` clean.

Then I tried to pass one of these things somewhere, and the polarity is inverted.

## The blocker

```c
gang Vec2 { chad x; chad y; };
gang Bag  { gang Vec2 pts[2]; rizz n; };
chad len2(gang Vec2 v) { bussin v.x * v.x + v.y * v.y; }
```

| Call | Should | Actually |
| --- | --- | --- |
| `len2(b.pts[0])` — one element, by value | accept | **rejected**, prints `0.0`, exit 0 |
| `len2(b.pts)` — the whole array | reject | **accepted**, silently passes `pts[0]` |
| `vec2_len2(b.pts[0])` — native `STDROT_STRUCT` | accept | **rejected**, exit 1 |
| `vec2_len2(b.pts)` — native `STDROT_STRUCT` | reject | **accepted**, `25.0` — array bytes into a C library |

The legitimate form is refused and the illegitimate one goes through. That is
exactly backwards from what #307 and #309 established for every other array
shape, and this PR is what makes both reachable — before it, neither could be
spelled.

Details and the fix inline. Both halves live in
`resolve_by_value_struct_source()`, one arm apart.

## The fixture cannot see it

`struct_array_field_of_structs.brainrot` only ever *reads through* the new
syntax — `pool.es[i].x`, `p.es[i].alive`. It never passes `pool.es[i]` or
`pool.es` anywhere. So the honest answer to "what incorrect implementation
would still pass this fixture" is: this one. That's the gap to close, not just
the code.

# VERDICT

This isn't a local slip; it's the same class of thing #307 and #309 each
closed, resurfacing because a new syntax was added without walking it through
the layer that consumes by-value struct sources. The feature makes a new
*expression shape* legal, and every consumer of that shape had to be checked,
not just the one that reads a field out of it. `resolve_struct_access()` was
updated; `resolve_by_value_struct_source()`, sitting right beside it and
handling the same two node types, wasn't.

Fix both arms, add the two argument cases to the fixture, and this is good
work.

ast.c
```diff
@@ -1197,6 +1324,82 @@ bool resolve_by_value_struct_source(ASTNode *expr, void **blob_out,
         return true;
     }
 
+    if (expr->type == NODE_ARRAY_ACCESS && expr->data.array.base)
```
leo-aa88, line 1456: **BLOCKING — heap-use-after-free.**

This arm returns a borrowed pointer and never considers that its base may have come out of the pending-return slot. When the base chain bottoms out in a call, `resolve_struct_access()` (line ~1085) runs that call and hands back a base pointing *inside* `current_return_value`'s blob. This arm forwards that as `*blob_out` with `*owned_out` still false.

```c
gang Bag make_bag() { ... bussin t; }
chad dot(gang Vec2 a, gang Vec2 b) { ... }

dot(make_bag().pts[0], mk())
```
```
ERROR: AddressSanitizer: heap-use-after-free on address 0x503000001150
READ of size 8
    #2 enter_function_scope ast.c:7819        ← memcpy from freed region
freed by thread T0 here:
    #1 free_pending_return_value ast.c:6934
    #2 execute_function_call ast.c:6856       ← mk() runs, clears the slot
```

`mk()` frees the `Bag` while argument 1 still points into it, then the bind loop copies from it. Single-argument forms leak instead of corrupting — `len2(make_bag().pts[0])` leaks 20 bytes; `gang Vec2 c = make_bag().pts[0];` and `bussin make_bag().pts[0];` leak 44 across two allocations.

The `NODE_STRUCT_ACCESS` arm forty lines up already solved this, and its comment describes this exact pair of symptoms — a UAF on the two-argument form, a leak on the return form. It needs the same treatment here: copy the element out, release the slot immediately, and report ownership.

**One wrinkle worth flagging**, because the obvious edit doesn't work: `struct_access_executes_call()` walks `NODE_STRUCT_ACCESS` links only —

```c
while (node && node->type == NODE_STRUCT_ACCESS)
    node = node->data.struct_access.object;
return node && node->type == NODE_FUNC_CALL;
```

`make_bag().pts[0]` is a `NODE_ARRAY_ACCESS` whose `base` is the `NODE_STRUCT_ACCESS`, so calling it on `expr` returns false. Either call it on `expr->data.array.base`, or extend the loop to step through `NODE_ARRAY_ACCESS` via `data.array.base` — I'd prefer extending it, since it's the shared "does resolving this run a call" predicate and every future shape will want the same answer from it.

Then the standard sequence: `calloc(1, elem_def->total_size)` (matching the sibling arm's allocator, which is what `enter_function_scope()`'s `free_owned_struct_arg_blobs()` releases with plain `free()`), `memcpy` from the computed element address, `free_pending_return_value()`, `*owned_out = true`. `elem_def` is already fetched and currently used only as a validity check — it's exactly the `total_size` you need.

Also: on the two early-return paths below (`!arr_fld->desc.is_array ...` and the rank check), a call base leaves the slot populated and nothing frees it. The sibling arm handles the equivalent with a `struct_access_executes_call()` guard before each `return false`.

test_cases/struct_array_field_of_structs.brainrot
```diff
@@ -0,0 +1,141 @@
...
+    🚽 two-dimensional struct array field
+    gang Grid g;
+    g.tag = 42;
+    g.cells[0][0].v = 1;
+    g.cells[1][2].v = 6;
+    yapping("sizeof Grid: %lu", maxxing(g));
+    yapping("grid %d %d tag=%d", g.cells[0][0].v, g.cells[1][2].v, g.tag);
+
+    🚽 An ELEMENT of a struct-array field is a by-value struct value
+    🚽 anywhere one is accepted: as an argument, and by extension as a
```
leo-aa88: The two added lines are exactly right — `len2(pool.es[0])` must work, `len2(pool.es)` must be rejected — and having both in one fixture is what makes the polarity itself the thing under test rather than one side of it.

This comment claims more than the lines check, though:

> as an argument, and **by extension** as a return value and a copy-initializer (all three go through `resolve_by_value_struct_source()`)

They do all go through that function, but they arrive with different lifetimes on the way out, and that's where the UAF I flagged on `ast.c` lives. Only the argument form is exercised here; the return and copy-init forms are asserted by "by extension." Both leak today:

```
gang Vec2 c = make_bag().pts[0];   → 44 bytes leaked
bussin make_bag().pts[0];          → same
```

"by extension" is the same shape of over-claim as `modulo_by_zero_smol_fail`'s uninitialized-slot sentence from #310 — a true statement about the code path standing in for a check that isn't there. Either exercise the other two forms, or say that only the argument form is pinned and the others share its resolver.

The case that would have caught the blocker is a call base, since that's the one where the three forms genuinely diverge:

```c
    yapping("%.1f", len2(make_pool().es[0]));
```

Worth adding alongside a plain-variable base once the ownership fix lands.

leo-aa88 (commented):
# REQUEST CHANGES

The polarity is fixed — both halves, both paths. I re-ran the exact cases:

```
len2(b.pts[0])      → 25.0 / 100.0     (was rejected)
len2(b.pts)         → rejected          (was silently pts[0])
vec2_len2(b.pts[0]) → 25.0              (was exit 1)
vec2_len2(b.pts)    → rejected, exit 1  (was 25.0 into a C library)
```

And the "TWO DISPATCHES" comment above `resolve_by_value_struct_source()` is
exactly the artifact this needed. It names the three bugs, says there are two
lists, and tells you to pass one somewhere rather than only read a field out of
it.

Which is why it stings that the new arm reproduces, one arm over, the other
half of what #307 already fixed for its sibling.

## BLOCKING — heap-use-after-free

```c
dot(make_bag().pts[0], mk())
```
```
ERROR: AddressSanitizer: heap-use-after-free on address 0x503000001150
READ of size 8
    #2 enter_function_scope ast.c:7819           ← memcpy from freed region
freed by thread T0 here:
    #1 free_pending_return_value ast.c:6934
    #2 execute_function_call ast.c:6856          ← mk()'s cleanup
    #3 enter_function_scope ast.c:7632
```

Argument 1 resolves through the new arm and hands back a pointer *into the
pending-return `Bag` blob*, borrowed, `*owned_out` left false. Argument 2's own
call frees that slot. The bind loop then copies from freed memory.

That is verbatim the scenario the `NODE_STRUCT_ACCESS` arm's own comment
records:

> Leaving it borrowed made `take(make_outer().inner, g())` a heap-use-after-free
> — g()'s cleanup freed the Outer, then the bind loop copied from it — and made
> `bussin make_outer().inner;` leak the Outer

Single-argument and non-argument forms leak rather than corrupt:

```
len2(make_bag().pts[0])              → 20 bytes leaked
gang Vec2 c = make_bag().pts[0];     → 44 bytes leaked across 2 allocations
bussin make_bag().pts[0];              (same)
```

`AGENTS.md` is explicit that a new leak or UB is a blocker rather than a
follow-up, and this is both.

Fix and the `struct_access_executes_call()` wrinkle inline.

## MINOR — the fixture's "by extension"

Also inline. The new fixture lines are the two I asked for and they do their
job. The comment beside them claims more than they check, in the specific
direction that's now broken.

# VERDICT

The fix is right about *what* the new arm has to compute and wrong about *who
owns it*. The address arithmetic was correctly mirrored from
`resolve_struct_access()` — but that function only ever hands its base to an
immediate field read, so it has no ownership story to copy, and the arm that
does have one is the `NODE_STRUCT_ACCESS` arm forty lines up.

Worth noting what this says about the comment added in the same commit: it
warned that adding an expression shape means updating two dispatches, and it
was right. What it didn't say is that the sibling arm's *pending-return
handling* is part of what has to be copied across, not just its address
arithmetic. Once fixed, that's the sentence worth adding to it.

Fix the ownership, add a call-base case to the fixture, and this is done.

leo-aa88 (commented):
# REQUEST CHANGES

No new commits since my last review — head is still `272a0391`. Rather than
repeat myself I rebuilt it, re-verified the finding, and spent the time
widening the pass into shapes I hadn't covered. One of those turned up a
correction I owe you, and one turned up more evidence.

## The blocker is unchanged

```
dot(make_bag().pts[0], mk())
→ ERROR: AddressSanitizer: heap-use-after-free
  SUMMARY: heap-use-after-free ... in __interceptor_memcpy
```

Still reproduces on this head.

**New evidence: the native path leaks too.** I'd only shown the Brainrot-side
leak before.

```
vec2_len2(make_bag().pts[0])      🚽 through STDROT_STRUCT
→ 25.0
  SUMMARY: AddressSanitizer: 44 bytes leaked in 2 allocations
```

So all three consumers of `resolve_by_value_struct_source()` are affected by
the same borrowed pointer, not just the two I'd tested.

## Correction — I was wrong about the early returns

My previous inline said:

> on the two early-return paths below (`!arr_fld->desc.is_array ...` and the
> rank check), a call base leaves the slot populated and nothing frees it

I tested both. Neither leaks:

```
len2(make_bag().pts[0][1])    → rank error, no leak
len2(make_bag().solo[0])      → not-an-array error, no leak
```

The enclosing call's own cleanup reaches the slot on those paths. The
`struct_access_executes_call()` guard is only needed on the **success** path,
where the pointer escapes this function — which is the actual bug. Disregard
that paragraph; the rest of the comment stands.

## What I newly verified as correct

All clean, all ASan-clean:

| Shape | Result |
| --- | --- |
| 2-D field element `g.rows[1][1]` by value | correct, neighbouring `tag` intact |
| element of a struct-array field of an element `p.bags[1].pts[1]` | correct, `k` intact |
| `chungus` (union) array field, element and whole-array | element works, whole array correctly rejected |
| OOB `b.pts[5]` used by value | bounds-checked — the inherited `calculate_array_offset()` claim holds |
| copy-init `gang Vec2 c = b.pts[0];` | works, and the copy is independent |

The nesting and union cases are the ones I'd have most expected to expose a
stride or tag error, and they don't. The address arithmetic in the new arm is
right; it's only the lifetime that isn't.

## Minor

The commit message lists the consumers of `resolve_by_value_struct_source()`
as "call arguments, copy initialization, struct returns, and struct
copy-assignment." There are three, not four — the callers are
`enter_function_scope()`, `marshal_struct_argument()`,
`interpreter_visit_declaration()` and `handle_return_statement()`, and none is
an assignment. Struct copy-assignment doesn't work anywhere in the language:
`e = s;` between two plain `gang Vec2`s reports *"Unsupported assignment
type"* on `main` today. Worth getting right because that enumeration is the
basis for the "single choke point" argument — which is otherwise correct, and
is why fixing this one arm covers everything.

# VERDICT

Unchanged: the arithmetic is right, the ownership isn't, and
`dot(make_bag().pts[0], mk())` corrupts the heap. Everything I threw at the
new arm this round came back correct except that. Fix the success path's
lifetime and add a call-base case to the fixture.

ast.c
leo-aa88, line 1456: Follow-up on my previous comment, after re-testing this arm at the current head.

**Still reproduces**, and the native path leaks as well as the Brainrot one:

```
vec2_len2(make_bag().pts[0])   → 25.0, 44 bytes leaked in 2 allocations
dot(make_bag().pts[0], mk())   → heap-use-after-free
```

**One correction to what I wrote before.** I said the two early returns below — the `!arr_fld->desc.is_array ...` check and the rank check — also leak when the base is a call. They don't:

```
len2(make_bag().pts[0][1])   → rank error, no leak
len2(make_bag().solo[0])     → not-an-array error, no leak
```

The enclosing call's cleanup reaches the slot on those paths. Only the **success** path needs the `struct_access_executes_call()` treatment, because that's the only one where the pointer outlives this function. Simpler fix than I implied — ignore that paragraph.

The rest holds: copy the element out, `free_pending_return_value()`, `*owned_out = true`, and remember that `struct_access_executes_call()` walks `NODE_STRUCT_ACCESS` links only, so it needs to see `expr->data.array.base` (or learn to step through `NODE_ARRAY_ACCESS`).

Everything else I tried against this arm came back correct — 2-D field elements, an element of a struct-array field reached through another element (`p.bags[1].pts[1]`), union array fields in both directions, OOB indices still bounds-checked, and copy-init producing an independent copy. The address arithmetic is sound; it's the lifetime that isn't.

ast.c
```diff
@@ -1277,6 +1400,82 @@ bool resolve_by_value_struct_source(ASTNode *expr, void **blob_out,
...
```
leo-aa88, line 1456: Still unfixed at this head — third review, same arm.

```
dot(make_bag().pts[0], mk())   → heap-use-after-free
len2(make_bag().pts[0])        → 20 bytes leaked
```

`grep -n struct_access_executes_call ast.c` on `7324b866` returns three hits, all in the `NODE_STRUCT_ACCESS` arm above. This arm still returns `*blob_out` pointing into the pending-return blob with `*owned_out` left false, so a sibling argument's call frees it out from under the bind loop.

The shape of the fix, restated compactly since it's now spread across three reviews:

- **success path only** — I withdrew the early-return part last time; those don't leak, the enclosing call's cleanup reaches them
- `struct_access_executes_call()` walks `NODE_STRUCT_ACCESS` links only, so it needs `expr->data.array.base` rather than `expr` (or teach it to step through `NODE_ARRAY_ACCESS` bases, which is what I'd do — it's the shared "does resolving this run a call" predicate and every future shape wants the same answer)
- then `calloc(1, elem_def->total_size)` — matching the sibling arm's allocator, which is what `free_owned_struct_arg_blobs()` releases with plain `free()` — `memcpy` from the computed element address, `free_pending_return_value()`, `*owned_out = true`
- `elem_def` is already fetched here and currently used only as a validity check; its `total_size` is exactly what the copy needs

Everything else I've thrown at this arm across three passes comes back correct — 2-D field elements, an element reached through another element, union array fields both directions, OOB still bounds-checked, copy-init independent, and now a pointer taken into a field and walked. It really is only the lifetime.

leo-aa88 (commented):
# REQUEST CHANGES

The blocker is untouched. I checked before reading the commit message, and
again after.

```
dot(make_bag().pts[0], mk())
→ ERROR: AddressSanitizer: heap-use-after-free
  SUMMARY: heap-use-after-free ... in __interceptor_memcpy

len2(make_bag().pts[0])
→ 20 bytes leaked in 1 allocation
```

Both still reproduce on `7324b866`. `struct_access_executes_call()` appears
three times in `ast.c` on this head, all of them in the `NODE_STRUCT_ACCESS`
arm; the `NODE_ARRAY_ACCESS`-with-base arm still hands back a borrowed pointer
with `*owned_out` false.

The new commit is good work — it just isn't this work, and its message doesn't
claim otherwise. Worth saying what it did land, because I'd want it kept:

**The combined-feature test is the right instinct and I'd not thought to ask
for it.** Taking a pointer *into* a struct-array field and walking it is what
a real pool helper writes, and it's precisely the shape that falls between two
PRs' fixtures — each feature tested alone, their intersection tested by
neither. Verified:

```
gang E *ep = &pool.es[0];
ep = ep + 2;      → x=20.0 id=2, ep[0] agrees, pool.k intact
```

**The comment reconciliation is correct.** #316 wrote "unsupported here, #315
adds a branch for it"; now that this *is* #315, saying the branch exists and
why the earlier wording was right at the time is the honest version — and it's
the second time in this pair of PRs that a comment spanning two branches has
been handled by dating it rather than by guessing.

478 pass.

# VERDICT

Unchanged from my last two reviews: the address arithmetic in the new arm is
right — I've now thrown 2-D fields, nested field-of-element, unions, OOB
indices, copy-init and a pointer-into-a-field at it, and all of them come back
correct — and the lifetime is wrong. `dot(make_bag().pts[0], mk())` corrupts
the heap, which `AGENTS.md` makes a blocker rather than a follow-up.

Rebasing first was reasonable with #316 landing underneath you. The fix is
small: `struct_access_executes_call()` on the success path only (my last
review withdrew the claim that the early returns need it too).

ast.c
```diff
@@ -906,21 +906,104 @@ bool resolve_struct_access(ASTNode *node, StructDef **def_out, void **base_out,
             parent_base = outer_field_addr;
         }
     }
+    else if (obj && obj->type == NODE_ARRAY_ACCESS && obj->data.array.base)
+    {
+        /* Member access on an element of a struct/union-typed ARRAY FIELD
```
leo-aa88, line 911: **MAJOR** — this branch resolves a member access *on* a struct-array-field element, but only one hop. A second hop is rejected before it ever gets here.

```c
gang Inner { chad v; chad w; };
gang Cell  { gang Inner core; chad y; };
gang Bag   { gang Cell pts[2]; rizz n; };

b.pts[0].y        🚽 works
b.pts[0].core.v   🚽 Error: Member access on non-struct/union value  (exit 1)
```

The shape it mirrors has no such limit:

```c
gang Cell arr[2];
arr[0].y        → 9.0
arr[0].core.v   → 3.0        🚽 both fine
```

So the new field form is strictly less capable than the name-based form it was
modelled on, and `pool.es[i].nested.field` — an immediately natural thing to
write once `pool.es[i].x` works — is a clean rejection with no explanation.

The rejection is not in this file. It comes from `semantic_analyzer.c:4023`,
whose member-access dispatch has an arm keyed on `obj->data.array.name` and
none on `obj->data.array.base`, so it can't decide that `b.pts[0]` is
struct-typed and falls through to *"Member access on non-struct/union value"*.

Two ways to close it, and I'd take either:

- teach that dispatch the `base` form, so it infers the element's tag from the
  field descriptor the way this branch does; or
- document it, the way #316 documented `(p + 1).x` as a known limitation of
  `resolve_struct_access()`. What's not fine is leaving it undiagnosed-in-
  substance — the message says the value isn't a struct, and it is one.

Worth noting the fixture gets close and stops one hop short: `Mid` has
`gang Inner many[2]`, and `o.m.many[0].a` reaches a *scalar* on the element.
Give `Inner` a nested struct member and the same fixture would have caught
this.

leo-aa88 (commented):
# COMMENT

**The blocker is fixed.** All four shapes that were broken now run clean, no
ASan or LSan output:

```
len2(make_bag().pts[0])              → 25.0
dot(make_bag().pts[0], mk())         → 7.0      (was heap-use-after-free)
gang Vec2 c = make_bag().pts[1];     → 6.0 8.0
gang Vec2 r = first();               → 3.0 4.0
vec2_len2(make_bag().pts[0])         → 25.0     (native path)
vec2_dot(make_bag().pts[0], make_bag().pts[0])  → 25.0
```

Non-call bases still work and don't double-free. Both early returns stay
clean — you were right to confirm that rather than take my withdrawal on
faith.

Extending `struct_access_executes_call()` to walk array links rather than
having one caller pass `expr->data.array.base` is the better of the two
options I offered, for the reason given: it's the shared predicate and the
next shape wants the same answer from the same place.

**Mutation-verified the fixture.** Disabling the ownership branch makes
`struct_array_field_of_structs` fail. Exercising all four forms against a
*call* base specifically — the only base where they diverge — is the right
correction to the "by extension" over-claim.

478 pass, `format-check` and `tidy` clean.

## One new finding

Inline. Short version: `pool.es[i].nested.field` is rejected at analysis time,
while the shape it mirrors — `arr[i].nested.field` on a plain array of
structs — works. There is a **third** dispatch on these node types, in
`semantic_analyzer.c`, and it wasn't updated.

## Two housekeeping notes

- **Conflicting again** since #314 merged. `tests/expected_results.json` only;
  `ast.c` merges clean.
- Unrelated pre-existing bug I tripped over while writing probes, worth its own
  issue: a struct parameter whose name collides with a caller variable of a
  *different* struct type resolves members against the caller's type.
  `chad getx(gang Vec2 b)` called with a `gang Bag b` in scope gives
  *"Struct 'Bag' has no member 'x'"* — identical on merged `main`, so not
  yours.

# VERDICT

The ownership fix is right and the reasoning recorded next to it is the most
useful thing this PR leaves behind. The remaining gap is the same lesson one
layer further out: the comment says two dispatches, and there are three.

ast.c
```diff
@@ -1151,13 +1234,66 @@ bool resolve_struct_access(ASTNode *node, StructDef **def_out, void **base_out,
...
+ * Every divergence between those two lists has been a bug. #307: the
+ * array-variable guard was added to one arm and not its sibling. #309:
+ * the static array check covered NODE_IDENTIFIER only. #315: a new
+ * expression shape (struct-typed array fields) was taught to
+ * resolve_struct_access() and not to this function, which inverted the
+ * polarity outright -- the legitimate `f(b.pts[0])` was refused while the
+ * illegitimate `f(b.pts)` silently passed element 0.
+ *
+ * So: if you are adding an expression shape that can denote a struct,
+ * there are TWO lists to update, not one -- and a fixture that only reads
```
leo-aa88: This comment is the best thing in the PR, and it's now one short of the truth.

> there are TWO lists to update, not one

There are three. `semantic_analyzer.c`'s member-access dispatch (around line
4023) keys on the same node types for the same expression shapes and answers a
third question — *is this expression struct-typed at all* — and it was not
updated for the `base` form. That's why `b.pts[0].core.v` is rejected while
`arr[0].core.v` works; see my other comment.

It fits the pattern the paragraph already describes exactly: two of the three
lists were updated, the third diverged, and the symptom is a shape that reads
fine one hop and dies on the second.

The addition about ownership crossing between dispatches is well judged, and
the sentence I'd keep verbatim if this comment ever gets trimmed:

> Getting that wrong is a heap-use-after-free with two arguments and a leak
> with one -- not a wrong value, so no amount of checking printed output finds
> it.

That's the actual reason the three prior fixtures missed it, stated in a way
that generalises past this bug. Worth adding the analyzer to the list and
naming what each of the three answers, so the next person can check all three
rather than the two that happen to live in `ast.c`.

leo-aa88 (commented):
# APPROVE

The gap is closed at the right layer, and I verified the rejections it had to
preserve as well as the shape it had to accept.

## Verified

```
b.pts[0].y              → 9.0     one hop
b.pts[0].core.v         → 3.0     two hops   (was a false diagnostic)
b.pts[0].core.deep.z    → 7.0     three hops
g.cs[1][1].core.v       → 4.0     2-D field, second hop
make_bag().pts[1].core.v → 5.0    second hop off a call base
```

Neighbouring scalars intact in every case, ASan/LSan silent — including the
call-base ones, which run through the ownership path from the previous commit.
That still holds: `dot(make_bag().pts[0], mk())` → `7.0`, clean.

The rejections that mattered are all still rejections, which is the half a
looser fix would have lost:

```
b.solo[0].core.v              → rejected   (non-array struct field, indexed)
make_bag().solo[0].core.v     → rejected   (same, off a call base)
b.pts[0].core.nope            → "Struct 'Inner' has no member 'nope'"
b.pts[0].y.z                  → rejected   (scalar on the second hop)
```

That last pair matters: the `is_array` guard is load-bearing exactly as the
comment says, and `infer_struct_def_static()` resolves a call base well enough
that the guard runs there too rather than failing open.

**Mutation-verified.** Disabling the new analyzer arm fails
`struct_array_field_of_structs` on its `second hop` line. And growing `Inner`
to carry a nested `gang Deep` is the right way to get there — the fixture
reached one hop before because `Inner` was flat, so the fix and the coverage
land together rather than the coverage being bolted on. `make abi-check`
re-confirms `Outer` at 28 with a real compiler.

480 pass, `format-check` and `tidy` clean, rebased and mergeable.

## The comment

`ast.c:1254` now reads **THREE DISPATCHES**, names what each one answers, and
adds the sentence that actually generalises:

> Two of the three live in ast.c, which is exactly why the third gets missed.

Plus what kind of fixture notices each — one hop misses the analyzer,
read-only misses the ownership dispatch. That is the part worth having: it
turns three separate bugs into a checklist someone can run before adding the
fourth expression shape.

# VERDICT

Worth naming what this PR converged on, because it took four rounds and the
shape of each round was the same: a new expression shape was added, and each
consumer of that shape had to be found independently — the runtime resolver,
the ownership resolver, then the static analyzer. Each miss had a different
signature (a false rejection, a use-after-free, a false diagnostic), and none
of them was findable by reading the feature's own code. The comment left
behind is the only artifact that makes the next one cheaper, and it now says
the right number.

Ship it.

PR #316, opened 2026-08-30
feat: struct pointer arithmetic and indexing (#311)

### LINKED ISSUE #311: [Feature]: the three limitations that still force workarounds in a real Brainrot program — struct-array fields, struct pointer arithmetic, `rant` parameters

leo-aa88:
## Description

Third and last of the three limitations in #311. **Struct pointer arithmetic
and indexing only** — `rant` parameters are #314, struct-array fields are #315.

Both forms failed:

```c
p = p + 1;   🚽 Cannot perform pointer arithmetic on a type-erased pointer
p[1].x       🚽 Member access on a non-struct/union array element
```

so a helper could not walk a pool — it could only be handed one
already-resolved `&pool[i]`, leaving the loops inlined at every call site:

```c
skibidi bump(gang Entity *p, rizz n) {
    flex (rizz i = 0; i < n; i = i + 1) { p[i].x = p[i].x + 100.0; }
}
```

### No new descriptor field was needed

The issue proposed carrying a struct tag on the pointer's descriptor. That
turned out to be unnecessary — **the tag is already on the `Variable`**.
`get_type_size_for_descriptor()` takes only `(VarType, pointer_level,
modifiers)`, so it genuinely cannot tell `gang Vec2 *` from `gang Matrix *`
(both `VAR_STRUCT` at pointer level 1) and correctly returns 0 rather than
guessing. `pointer_arith_scale()` does the second lookup, via
`get_struct_def_for_expression()`, for exactly the case that signature
structurally cannot serve. Opaque `VAR_PTR`/`VAR_VOID` pointers still get 0
and are still rejected.

### Indexing had to land in the shared address routine

This was the whole difficulty, and it is worth recording. Fixing
`resolve_struct_access()` alone left `yapping("%f", p[1].x)` working while
`chad t = p[1].x;` still died — those arrive by different routes.
`evaluate_multi_array_access()` is the shared element-address routine all of
them funnel through: the interpreter's generic traversal
(`interpreter_visit_array_access`), `evaluate_lvalue_address()` for an
assignment target, and the scalar evaluators for a value load. I found that by
taking a backtrace rather than guessing at which path was involved.

`calculate_array_offset()` is deliberately **not** reused for the pointer case:
it returns 0 whenever `is_array` is false, which would silently collapse
`p[5]` to `p[0]`.

### Rejections, and one deliberate non-rejection

Null pointer, multi-level pointer (`gang E **pp` needs an explicit
dereference — the same single-level rule the member-access branches already
enforce), and more than one index (the second would be indexing into a struct).

An **out-of-bounds index is not rejected**, and cannot be: a pointer carries no
extent, exactly as in C. That asymmetry with the bounds-checked array form
(`array_of_structs_oob_fail`) is a real property of the feature, and the
fixture says so rather than leaving it to be discovered.

### Not implemented

`(p + 1).x` — member access directly on a parenthesized expression. That is a
**pre-existing general limitation** of `resolve_struct_access()` (it resolves
identifier, member-access, array-access and call bases, not arbitrary
operations), not anything specific to struct pointers. Both forms the issue
asked for work without it, and it is documented.

## Related Issue

Closes #311 (with #314 and #315 covering the other two items).

## Type of Change

- [x] Documentation update
- [ ] Performance improvement
- [ ] Refactor

## Checklist

**Test coverage.** `struct_pointer_arithmetic.brainrot` walks a pool through a
bare pointer parameter, **writes** through it and confirms the caller's own
array changed (so the element addresses are lvalues, not readable copies), does
`p + 2` and `p - 1`, indexes relative to a moved pointer, and checks an ordinary
scalar pointer is unaffected.

`Entity` is deliberately two fields rather than one: a stride bug using
`sizeof(chad)` would still read element 0 correctly and only diverge from
element 1 onward, so a single-field struct would not catch it.

`struct_pointer_index_fail.brainrot` pins the three rejections **and the values
printed after them**, so an implementation that reported the error and then
indexed anyway would fail.

**Verification.** 477 tests pass (475 → +2). Full valgrind sweep **416/416
clean**, run against a non-sanitized build since `make valgrind` cannot
currently run an ASan binary (#186/#203, open PR #202). `make format-check` and
`make tidy` clean.

## Note for reviewers

This branches from `main`, as do #314 and #315 — the three are independent and
touch different code (`enter_function_scope` / `lang.y` + `resolve_struct_access`
/ `evaluate_multi_array_access` + pointer arithmetic). #315 and this one both
add to `docs/the-brainrot-programming-language.md`'s struct *Current
Limitations* list and to `tests/expected_results.json`, so whichever merges
second will want a trivial rebase there.


test_cases/struct_pointer_arithmetic.brainrot
```diff
@@ -0,0 +1,93 @@
...
+🚽 The issue proposed carrying a struct tag on the pointer's descriptor.
+🚽 That turned out to be unnecessary: the tag is already on the Variable.
+🚽 get_type_size_for_descriptor() takes only (VarType, pointer_level,
+🚽 modifiers), so it genuinely cannot tell `gang Vec2 *` from
+🚽 `gang Matrix *` -- both are VAR_STRUCT at pointer_level 1 -- and
+🚽 correctly returns 0 rather than guessing. pointer_arith_scale() (ast.c)
+🚽 does the second lookup that recovers the tag for exactly that case.
+🚽
+🚽 ── Why the sizes below are what they are ─────────────────────────────
+🚽 Entity is {chad x; rizz id;} = 8 bytes, so `p + 1` advances 8 and
```
leo-aa88: **MAJOR.** This comment notices that `Entity` is 8 bytes. It doesn't notice that `sizeof(void *)` is also 8 — which makes the fixture blind to the exact failure this PR exists to prevent.

`pointer_arith_scale()`'s new struct branch is the substance of the change: `get_type_size_for_descriptor()` returns 0 for a by-value struct pointee, and the fallback lookup through `get_struct_def_for_expression()` is what supplies the real stride. The most likely way to get that wrong is for the fallback not to fire and a pointer-sized stride to be used instead. At 8 bytes those are the same number.

Mutation, replacing the entire struct-tag lookup:

```c
if (ptr_level == 1 && get_expression_type(expr) == VAR_STRUCT)
{
    return sizeof(void *);      /* was: get_struct_def_for_expression(...) */
}
```
```
1 passed, 476 deselected
```

The fixture is green with the feature's core removed.

For what it's worth, the code is right — I checked separately with a 24-byte struct:

```c
gang Big { gigachad a; gigachad b; gigachad c; };   🚽 maxxing = 24
p = p + 1;   → a=20.0 c=2.0     (element 1)
p = p + 1;   → a=30.0 c=3.0     (element 2)
q[2].a = 99.0;  → arr[2].a == 99.0
```

So this is a test-strength problem, not a correctness one. But the description argues at length that no descriptor field was needed because the tag is reachable on the `Variable` — that argument is the PR's central claim, and nothing in the suite currently tests it.

Cheapest fix: make `Entity` not 8 bytes (add a `gigachad`, or a third field), or keep `Entity` and add one `Big`-shaped struct alongside it. A struct whose size is neither `sizeof(void *)` nor `sizeof(int)` is the one that makes this assertion mean something. Worth saying so in the comment too, since "8 bytes" reads as a deliberate choice right now and the next person will preserve it.

ast.c
```diff
@@ -909,63 +909,135 @@ bool resolve_struct_access(ASTNode *node, StructDef **def_out, void **base_out,
...
     {
         /* Member access on an element of an array of struct/union values or
-           pointers (`pts[i].x`, `ptrs[i].x`). Only the name-based array form
-           is an lvalue-addressable struct here; an array-typed struct FIELD
-           element (`foo.arr[i].x`, obj->data.array.base set) is not
-           supported and falls through to the error below. */
+           pointers (`pts[i].x`, `ptrs[i].x`), or on an INDEXED STRUCT
+           POINTER (`p[i].x`, #311 -- see the pointer branch below). The
+           struct-field array form (`foo.arr[i].x`, obj->data.array.base
+           set) is handled by its own branch above. */
```
leo-aa88: **MINOR**, and it's a merge-order artifact rather than a mistake in the logic.

> The struct-field array form (`foo.arr[i].x`, `obj->data.array.base` set) is handled by its own branch above.

There is no such branch on this PR. `grep -c "ARRAY FIELD" ast.c` → `0`. That branch is #315's, and this PR is based on `main`, independent of it.

The comment this replaced was accurate for this branch:

> an array-typed struct FIELD element (`foo.arr[i].x`, obj->data.array.base set) is not supported and falls through to the error below.

So as it stands, an accurate comment was swapped for one that describes a different branch's code. If #316 lands first, or if #315 gets reworked, it's a reader pointing at a branch that isn't there.

Good news on the mechanics: I test-merged the two and `ast.c` auto-merges cleanly — #315 inserts its branch above, this rewrites the one below, and git handles it. Only `tests/expected_results.json` conflicts, the usual tail-append. So **merge #315 first** and this comment becomes true on arrival; otherwise reword it to say the field form is unsupported here and will be handled by #315.

Either way it's worth one line in the PR description saying the two touch the same function, since both are open and neither mentions the other in that respect.

leo-aa88 (commented):
# COMMENT

The implementation is correct — I verified it independently, including the
part the description spends its longest section on. The test suite does not
establish that, and I can show you the mutation that proves it doesn't.

## Verified working

The `evaluate_multi_array_access()` insight is real and is the thing that
would have been easiest to get wrong. `chad t = p[1].x;` and
`yapping("%f", p[1].x)` do arrive by different routes, and putting the pointer
case in the shared address routine is what makes both work. Finding that by
backtrace rather than guessing was the right move.

Also checked, all ASan-clean:

- `p[i].x` read **and** write through a parameter (`bump(gang E *p, rizz n)`),
  with the caller's array observably updated
- `p = p + 1`, `p = p - 1`, and `q[2]` all landing on the right element
- **the stride is genuinely `sizeof(struct)`** — 24 for a
  `{gigachad a,b,c;}` — not a pointer size
- array-of-struct-**pointers** (`ptrs[i].x`) still works after the branch was
  restructured; that was the regression risk in splitting that `if`
- member access on a non-struct array still errors
- null pointer, multi-level pointer, and two-index forms all rejected cleanly,
  and the fail fixture prints the resulting values rather than discarding them

Not reusing `calculate_array_offset()` is right for the stated reason — it
returns 0 when `is_array` is false, which would collapse `p[5]` to `p[0]`.
Declining to add a descriptor field because the tag is already on the
`Variable` is also right. 477 pass, `format-check` and `tidy` clean.

## The finding

`struct_pointer_arithmetic.brainrot` cannot tell your implementation from the
most likely wrong one, because `Entity` is 8 bytes and so is a pointer.
Inline, with the mutation.

## Merge coordination

`ast.c` auto-merges cleanly between this and #315 — only
`tests/expected_results.json` conflicts, the usual tail-append. But the comment
at `ast.c:915` is written for the post-#315 world; see the inline note. Merging
#315 first makes it true. (My blocking finding on #315 is in
`resolve_by_value_struct_source()`, which this PR doesn't touch, so the two are
independent on that axis.)

# VERDICT

Good feature work with a test that agrees with the implementation for the
wrong reason. The struct-tag lookup in `pointer_arith_scale()` is the whole
substance of this PR and it is the one thing currently unverified — one
differently-sized struct in the fixture fixes that.

leo-aa88 (commented):
# APPROVE

Both notes addressed, and I re-ran the mutations myself rather than taking the
commit message's word for it.

**The stride is now actually tested.** The original mutation — replacing the
whole struct-tag lookup in `pointer_arith_scale()` with `return sizeof(void *)`
— went from passing to failing:

```
FAILED test_brainrot.py::test_brainrot_examples[struct_pointer_arithmetic-...]
```

I also checked the second claim the new fixture comment makes, that an
int-sized stride would be caught. Mutated it to `sizeof(int)`: also fails. So
both numbers named in the comment are real, not asserted.

Picking 24 via `{gigachad a, b, c;}` is the right shape — it collides with
neither `sizeof(void *)` nor `sizeof(rizz)` nor `sizeof(gigachad)`, and
exercising it three ways (`bp[0]`/`bp[2]`, repeated `+ 1`, and a write through
the far element read back via the array) means a stride error shows up as a
wrong value rather than an off-by-one that happens to land in bounds. Rewriting
the header comment so "8 bytes" no longer reads as a sufficient deliberate
choice matters as much as the fixture change — that's the line that would have
invited someone to simplify it back.

**The `ast.c` comment is correct now**, and correctly hedged. `grep -c
"ARRAY FIELD" ast.c` is still 0 on this head, so "NOT handled here and falls
through to the error" is accurate, and naming #315 as the change in flight
tells a reader where to look without asserting code that isn't present.
Adding that it will be "stale in the harmless direction" once #315 lands is
the right way to write a comment that spans two open branches — it can't
become *wrong*, only redundant.

477 pass, `format-check` and `tidy` clean.

# VERDICT

The implementation was already right; what changed is that the suite now says
so. The mutation you can't survive is the one worth writing, and this fixture
now fails two of them instead of passing both.

Ship it.

PR #320, opened 2026-08-30
fix(grammar): allow bare bussin; void return statements (#283)

### LINKED ISSUE #283: [Bug]: bare bussin; (void return) is a parse error — skibidi functions can't return early
### Description

A bare `bussin;` (return with no value) is a **parse error**, so a `skibidi` (void) function has no way to return early. This contradicts the interpreter, which was clearly built to support a value-less return: `interpreter_visit_return_statement` explicitly handles the `node->data.op.left == NULL` case and calls `handle_return_statement(NULL)` (`interpreter.c:824-835`).

The gap is purely in the grammar. `return_statement` has a single production that *requires* an expression:

```
return_statement:
    BUSSIN expression
        { $$ = create_return_node($2); }
    ;
```
(`lang.y:1907`)

There is no `BUSSIN SEMICOLON` alternative, so `bussin;` can never be produced.

### Minimal reproduction code

```c
skibidi noop() {
    bussin;              🚽 early return from a void function
}
skibidi main {
    noop();
    yapping("ok");
}
```

### Expected behavior

Parses and runs, printing `ok`. A `skibidi` (void) function can `bussin;` to return early, mirroring C's `return;`.

### Actual behavior

```shell
Error: syntax error, unexpected SEMICOLON at line 1
Parsing failed
```

### Brainrot version / commit

### Operating system

WSL2 (Ubuntu), x86-64

### Additional context

Fix is a one-line grammar addition — a `BUSSIN SEMICOLON` alternative in `return_statement` that builds a return node with a NULL expression (`create_return_node(NULL)`), which the interpreter already handles. Distinct from #246, which is about `bussin <expr>` in `main` not halting execution.

Aditya-233:
## Description

Fixes an issue where a bare `bussin;` (void return with no expression) failed to parse because the Bison grammar only accepted `BUSSIN expression` under `return_statement`.

The interpreter was already implemented to support value-less return (`interpreter_visit_return_statement` in `interpreter.c` explicitly handles `node->data.op.left == NULL` by calling `handle_return_statement(NULL)`).

This change adds the `BUSSIN` alternative to `return_statement` in `lang.y` to build a return node with a `NULL` expression (`create_return_node(NULL)`).

### Tests and Stress Testing
Added regression test fixture `test_cases/bare_bussin_void_return.brainrot` (and matching expected output in `tests/expected_results.json`) covering:
- Bare `bussin;` in a noop void function
- Conditional early return from a void function (`edgy (...) { bussin; }`)
- Recursive void function using `bussin;` for base case termination
- Scope and local variable cleanup when `bussin;` exits early
- Bare `bussin;` at the end of `skibidi main`

Additionally, stress tested across:
- Nested conditional branches (`edgy` / `sus`)
- Deeply nested scopes and structs
- Multiple sequential void calls with early exits
- Full valgrind memory leak suite (0 leaks, 0 errors)
- Full compiler test harness (477 passed, 5 skipped for optional raylib)

## Related Issue

Fixes #283

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement
- [ ] Refactor

## Checklist

### DIFF
--- a/lang.y
+++ b/lang.y
@@ -1991,8 +1991,10 @@ error_statement:
     ;
 
 return_statement:
-    BUSSIN expression
+      BUSSIN expression
         { $$ = create_return_node($2); }
+    | BUSSIN
+        { $$ = create_return_node(NULL); }
     ;
 
 expression:
--- a/test_cases/bare_bussin_void_return.brainrot
+++ b/test_cases/bare_bussin_void_return.brainrot
@@ -1,0 +1,64 @@
+🚽 Test case: issue #283 -- bare `bussin;` (void return with no value)
+🚽 was a parse error in the grammar.
+🚽
+🚽 return_statement in lang.y only accepted `BUSSIN expression`, preventing
+🚽 skibidi (void) functions from returning early with `bussin;`.
+🚽
+🚽 This fixture tests:
+🚽 1. Bare `bussin;` in a noop void function.
+🚽 2. Conditional early return from a void function (`edgy (...) { bussin; }`).
+🚽 3. Recursive void function using `bussin;` for base case termination.
+🚽 4. Scope and local variable cleanup when `bussin;` exits early.
+🚽 5. Bare `bussin;` at the end of `skibidi main`.
+
+gang Point {
+    rizz x;
+    rizz y;
+};
+
+skibidi noop() {
+    bussin;
+}
+
+skibidi check_positive(rizz n) {
+    edgy (n <= 0) {
+        bussin;
+    }
+    yapping("positive: %d", n);
+}
+
+skibidi countdown(rizz n) {
+    edgy (n <= 0) {
+        yapping("liftoff");
+        bussin;
+    }
+    yapping("count: %d", n);
+    countdown(n - 1);
+    bussin;
+}
+
+skibidi early_exit_locals() {
+    gang Point p;
+    p.x = 42;
+    p.y = 100;
+    rizz arr[4];
+    arr[0] = 7;
+    bussin;
+    yapping("unreachable locals");
+}
+
+skibidi main {
+    noop();
+    yapping("noop passed");
+
+    check_positive(-5);
+    check_positive(10);
+    check_positive(0);
+
+    countdown(3);
+
+    early_exit_locals();
+    yapping("locals exit passed");
+
+    bussin;
+}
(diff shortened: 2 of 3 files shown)


lang.y
```diff
@@ -1991,8 +1991,10 @@ error_statement:
     ;
 
 return_statement:
-    BUSSIN expression
+      BUSSIN expression
         { $$ = create_return_node($2); }
+    | BUSSIN
```
leo-aa88, line 1996: Correct and safe. `create_return_node(NULL)` sets `data.op.left = NULL`, which `interpreter_visit_return_statement` and `handle_return_statement`'s `!expr` arm already handle, and the semantic analyzer's return visitor is NULL so nothing dereferences the missing expression. Verified: `bison -Wcounterexamples` reports zero conflicts before and after (SEMICOLON is not in FIRST(expression), so the optional-expression form is unambiguous), and the checked-in fixture that passes here fails to parse on main's old grammar (`unexpected SEMICOLON`). Non-blocking: this also makes bare `bussin;` legal in a value-returning function, where the `!expr` arm returns a stale `current_return_value` payload, but that is the same pre-existing gap as falling off the end of such a function, not something this PR introduces.

leo-aa88 (approved):
LGTM.

I built this and verified it end to end rather than trusting the description.

**The fix is correct and minimal.** `return_statement` only accepted `BUSSIN expression`, so a bare `bussin;` was a hard parse error even though the interpreter already handled a value-less return (`interpreter_visit_return_statement` branches on `node->data.op.left == NULL`, and `handle_return_statement` has an `!expr` arm). Adding `| BUSSIN { create_return_node(NULL); }` activates that existing path. `create_return_node(NULL)` sets `data.op.left = NULL` safely, and the semantic analyzer registers a NULL return visitor, so the value-less node is never dereferenced during analysis.

**No new grammar conflicts.** I regenerated the parser with `bison -Wcounterexamples` on both main and this branch: zero conflicts before, zero after. The optional-expression alternative is unambiguous because SEMICOLON (the FOLLOW of `return_statement`) is not in the FIRST set of `expression`.

**The fixture is a real regression test.** On this branch the checked-in `test_cases/bare_bussin_void_return.brainrot` prints exactly the expected `noop passed / positive: 10 / count: 3 / count: 2 / count: 1 / liftoff / locals exit passed`. I then rebuilt the interpreter with main's old grammar and the same fixture fails to parse (`syntax error, unexpected SEMICOLON at line 19`), so it genuinely pins the bug. It covers a noop void function, conditional early return, a recursive base case, early return with live locals/struct/array in scope, and bare `bussin;` at the end of `skibidi main`. Expected-output entry matches byte for byte, and the ASan/UBSan build runs it leak-clean.

Full `pytest` run is green (483 passed) with no regression.

**One non-blocking observation, not introduced by this PR.** Allowing bare `bussin;` everywhere also makes it legal inside a value-returning function (e.g. `rizz f() { bussin; }`), where `handle_return_statement`'s `!expr` arm sets `has_value = true` and the declared type but never writes `value.*`, so the caller reads whatever was last left in the global `current_return_value` slot. I confirmed `rizz bad(...) { rizz t = side(999); bussin; }` returns a stale `999`. This is NOT a regression: falling off the end of a value-returning function with no `bussin` at all already returns the same stale value today, so this is the pre-existing "no missing-return diagnostic" gap, reached by one more syntactic path. Worth a follow-up if the project ever adds return-value checking; not a reason to hold this fix.

Ship it.

Issue #319, opened 2026-08-30
[Bug]: `bussin` inside a loop is caught by the loop's jump buffer — the program dies with "No scope to exit" and prints nothing


leo-aa88:
### What happens

Returning a value from inside a `flex` or `goon` loop kills the whole program. No output is produced at all, and the exit code is 1.

### Repro

```c
rizz first_match(rizz n) {
    flex (rizz i = 0; i < n; i = i + 1) {
        edgy (i == 2) { bussin i; }
    }
    bussin 0 - 1;
}

skibidi main {
    rizz a = first_match(5);
    yapping("a %d", a);
    yapping("done");
    bussin 0;
}
```

```
$ ./brainrot repro.brainrot
Error: No scope to exit at line 12
$ echo $?
1
```

**stdout is completely empty** — neither `yapping` runs. `goon` behaves identically. Reproduces on `v0.2.0` and on `main` (`f16c9e2`), so it is not a regression from the recent feature work.

### Mechanism

`interpreter_visit_for_statement()` (`interpreter.c:747`) pushes its **own** jump buffer so that `break` can escape the loop:

```c
    PUSH_JUMP_BUFFER();
    if (setjmp(CURRENT_JUMP_BUFFER()) == 0)
    {
        enter_scope();
        ...
        exit_scope();
    }
    POP_JUMP_BUFFER();
```

`handle_return_statement()` (`ast.c`) unwinds to the function scope and then longjmps:

```c
    while (current_scope && !current_scope->is_function_scope) { exit_scope(); }
    if (jump_buffer) { exit_scope(); LONGJMP(); }
```

`LONGJMP()` goes to the **innermost** buffer, which is the loop's, not the function's. So:

1. The return unwinds the loop's scopes and the function scope.
2. The longjmp is caught by the loop, whose `setjmp` now returns non-zero — the `if` body is skipped, `POP_JUMP_BUFFER()` runs, and control falls through to the statement *after* the loop, still inside a function that was supposed to have returned.
3. That function runs on with no scope, and the next `exit_scope()` finds `current_scope == NULL`.
4. `exit_scope()` calls `exit(1)` directly (`ast.c:6949`), so the process dies before anything is flushed.

In short: **`bussin` inside a loop is silently executed as `break`**, and then the mismatched scope stack takes the process down.

### Why it matters

Search-and-return is one of the most ordinary shapes there is:

```c
rizz free_slot(gang Ent *pool, rizz n) {
[...]

Issue #312, opened 2026-08-30
[Bug]: a caller's local shadows a callee's parameter name during semantic analysis, so a correct call is rejected


leo-aa88:
### What happens

If a function's parameter shares its name with a variable in the **caller's** scope, the semantic analyser types the parameter from the caller's variable instead of from the declaration. A correct call is then rejected.

### Repro

```c
gang P { chad x; };

chad wid(gang P *bm) { bussin 7.0; }
cap  use(gang P *bm) { bussin wid(bm) > 0.0; }

skibidi main {
    gang P bm;          🚽 a NON-pointer local with the same name as the param
    bm.x = 1.0;
    yapping("%b", use(&bm));
    bussin 0;
}
```

```
Error: 'wid' argument 1: expected pointer to struct/union 'P' (level 1), got struct pointer level 0 at line 3
```

Line 3 is `use`'s body. Inside `use`, `bm` **is** `gang P *` — declared one line above. The analyser resolves it against `main`'s `gang P bm` (pointer level 0) instead.

Rename the caller's local and nothing else changes:

```c
    gang P other;
    other.x = 1.0;
    yapping("%b", use(&other));
```

```
W
```

Verified on `2ac321d` (v0.2.0).

### Why it matters

It is not a shadowing *warning* — a valid program is rejected, and the error points at the callee while naming a type that is correct there. The message says "got struct pointer level 0" about a variable whose declaration two lines up says `*`, which sends you looking in the wrong function.

It also makes parameter names part of a library's public surface. In [tung-tung-sahur](https://github.com/Brainrotlang/tung-tung-sahur), `src/sim.brainrot` names its parameters `w`, `p`, `b`, `e`, `bm` and `skibidi main` names its locals `wd`, `pl`, `bo` — that non-collision is load-bearing and nothing enforces it. A caller that picks a natural local name for a struct it owns gets a compile error inside a file it did not write, and no naming discipline on either side is actually safe.

This is the "separate scoping bug" already noted in passing in `test_cases/func_call_evaluated_once.brainrot`:

> And the parameter is `flag`, not `done`: if it shares a name with the CALLER's local, `*flag` inside here resolves against the caller's non-pointer variable and the call is rejected -- a separate scoping bug, avoided here rather than exercised.

Filing it so it is exercised rather than avoided.

### Expected

A parameter is in scope for its own function body and shadows anything of the same name in the caller. The repro should print `W`.

### Notes

[...]

PR #324, opened 2026-08-30
fix: allow a cap-returning call in an integer context (#313)

### LINKED ISSUE #313: [Bug]: a `cap`-returning call cannot be used as a condition — `edgy (f())` errors and the branch silently never runs
### What happens

A function returning `cap` cannot be used directly where an integer is wanted — including an `edgy` / `goon` condition. The call errors, evaluates to 0, and the branch never runs.

### Repro

```c
cap yes() { bussin W; }

skibidi main {
    edgy (yes()) { yapping("direct"); }        🚽 errors; never prints

    cap v = yes();
    edgy (v) { yapping("via var"); }           🚽 fine

    bussin 0;
}
```

```
Error: Native call result (bool) cannot be used in an integer context at line 3
via var
```

No `#cooked` modules, no natives involved. Verified on `2ac321d` (v0.2.0).

### Root cause

`unbox_native_numeric_result()` (`ast.c:3779`) has no `VAR_BOOL` case:

```c
static bool unbox_native_numeric_result(void *raw, VarType actual, double *out)
{
    switch (actual)
    {
    case VAR_INT:
    case VAR_ENUM:   ...
    case VAR_SHORT:  ...
    case VAR_FLOAT:  ...
    case VAR_DOUBLE: ...
    case VAR_CHAR:   ...
    default:
        return false;     🚽 VAR_BOOL lands here
    }
}
```

Both integer-context callers (`ast.c:4579` and `ast.c:4799`) then emit the diagnostic and `return 0`. `cap` is the one scalar type the unboxer does not know.

### Two things that make this worse than the error suggests

1. **It fails open.** The condition evaluates to 0, so `edgy (yes())` takes the false branch. A program that ignores stderr just quietly does the wrong thing — the same failure mode as #303 and #296.
2. **The message names the wrong culprit.** It says *"Native call result"* for a user-defined function, because the diagnostic is shared with the native path. That sends you looking at your `#cooked` modules for a function that is twelve lines up in the same file.

### Impact

This silently shapes how code gets written. In [tung-tung-sahur](https://github.com/Brainrotlang/tung-tung-sahur), every one of `player_visible`, `player_swinging`, `player_vulnerable`, `ent_hits_body`, `ent_in_bat`, `plat_jump_down`, `plat_should_close` returns `cap`, and **every single call site** assigns to a `cap` local first:

```c
cap show = player_visible(&pl);
edgy (show) { ... }
```

That reads like style. It is not — the direct form does not work, and nothing says so.

### Expected

`edgy (yes())` prints `direct`. A `cap` is 0 or 1 in an integer context, the same as `!0` already produces.

### Suggested fix

```c
    case VAR_BOOL:
        *out = (double)*(bool *)raw;
        return true;
```

alongside a test that pins `edgy (f())`, `goon (f())` and `rizz n = f();` for a `cap`-returning user function — the declaration path already works, so a fixture covering only that would pass without the fix.

leo-aa88:
## Description

`edgy (yes())` for a `cap`-returning function reported *"cannot be used in an
integer context"* and took the **false** branch.

```c
cap yes() { bussin W; }

skibidi main {
    edgy (yes()) { yapping("direct"); }   🚽 errored; never printed
    cap v = yes();
    edgy (v) { yapping("via var"); }      🚽 fine
    bussin 0;
}
```

`unbox_native_numeric_result()` (`ast.c`) handled `VAR_INT`/`ENUM`/`SHORT`/
`FLOAT`/`DOUBLE`/`CHAR` and sent everything else to `default: return false`.
`cap` was the one scalar type it did not know. `handle_function_call()` boxes a
bool result as `SAFE_MALLOC(bool)` and writes through a `bool *`, so reading it
back the same way is all that was missing — the fix the issue suggests.

**It failed open**, which is the part that matters: the condition evaluated to
`0`, so a program ignoring stderr quietly took the wrong branch — the same
failure mode as #296 and #303.

### The diagnostic named the wrong culprit

These messages are shared between the native and user-defined call paths and
said *"Native call result"* unconditionally, sending a reader to their `#cooked`
modules for a function twelve lines up in the same file. The distinction is
knowable at the site — the node *is* the call — so it now says `User-defined`
or `Native` accordingly:

```
Error: User-defined call result (string) cannot be used in an integer context
```

**Deliberately not made generic.** For a real native, "Native" is the
informative word, and 12 existing fixtures pin it — including
`native_zero_arg_string_ownership`, which pins this exact integer-context
message, still reads "Native", and still passes. A blanket rewording to
something like "Call result" would have lost real information in the common
case to fix the uncommon one.

## Related Issue

Fixes #313

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement
- [ ] Refactor

## Checklist

**Test coverage.** `test_cases/cap_returning_call_in_condition.brainrot` covers
the call directly as an `edgy` condition, the **false** case (so a fix that
hardcoded "true" would fail), a `goon` condition mixed into a larger boolean
expression, and a non-condition integer context (`!yes()`).

The adversarial part is *what* it asserts: because the bug failed open, the
fixture pins that the **true branch ran**, not merely that no diagnostic
appeared. An implementation that reports the error and evaluates `0` prints
nothing there and fails. **Mutation-verified** — removing the `VAR_BOOL` case
fails the fixture.

**Deliberately not covered, because it is not this bug:** `rizz n = yes();` is
rejected with *"expected int, got bool"* — but `rizz n = b;` for a plain
`cap b` is rejected identically, with no call involved. That is a declaration
type rule, consistent either way, and pinning it here would mislabel it as part
of this fix. The fixture says so.

**Verification.** 483 tests pass (482 → +1). Full valgrind sweep **422/422

## Note

This is the first of #312, #313 and #319 — the other two are separate PRs, as
they touch different subsystems (semantic scoping and loop/return control flow
respectively).


test_cases/cap_returning_call_in_condition.brainrot
```diff
@@ -0,0 +1,54 @@
...
+    rizz n = 0;
+    goon (n < 2 && yes()) {
+        n = n + 1;
+    }
+    yapping("goon ran %d", n);
+
+    🚽 an integer context that is not a condition
+    yapping("negated %d", !yes());
+    bussin 0;
+}
```
leo-aa88, line 54: **MAJOR.** This fixture is never executed.

`test_brainrot_examples` is parametrized over `tests/expected_results.json`, and there's no entry for `cap_returning_call_in_condition` — the PR doesn't touch that file at all.

```
$ pytest -q -k cap_returning_call_in_condition
481 deselected in 0.03s
```

Nothing selected. And the suite total on this branch is `480 passed, 1 deselected` — byte-identical to `main`'s, which is the giveaway: a PR that adds a test file should move that number.

`AGENTS.md` states the requirement directly:

> `test_cases/*.brainrot` — fixtures; each needs a matching entry in `tests/expected_results.json`

Running it by hand, it produces:

```
edgy true
edgy false
goon ran 2
negated 0
```

so the entry is just:

```json
"cap_returning_call_in_condition": "edgy true\nedgy false\ngoon ran 2\nnegated 0"
```

Worth confirming afterwards that it actually guards the fix — revert the `VAR_BOOL` case in `unbox_native_numeric_result()` and check this fixture goes red. Given the bug was fail-open, a fixture that only checks stderr would have passed against the broken build; this one checks which branch was taken, which is the right thing to pin.

leo-aa88 (commented):
# REQUEST CHANGES

The code fix is correct and I verified both halves of it. Two things around it
are not.

**The fix works.** Reproduced the bug on `main` first:

```
edgy (yes())   → Error: Native call result (bool) cannot be used in an
                 integer context   ... and the branch is NOT taken
```

Fail-open, exactly as described — a program ignoring stderr silently does the
wrong thing. On this branch `edgy (yes())` takes the branch and `edgy (no())`
doesn't. Both arms of the new qualifier are right too: a `rant`-returning
user function gives *"User-defined call result (string)..."*, `slorp` gives
*"Native call result (string)..."*.

## 1. A compiled `.so` is committed

`brainray/raylib.so`, 49,712 bytes of build output, in a directory that no
longer exists — #318 renamed `brainray/` to `rayrot/` and merged before this.
`AGENTS.md` is unambiguous:

> **Never**: commit generated files (`lang.tab.*`, `lex.yy.c`, `brainrot`,
> `libstdrot.so`)

I can see how it happened, and it's worth writing down because it will recur:
#318 changed the ignore entry from `brainray/raylib.so` to `rayrot/raylib.so`.
Anyone with a pre-rename working tree was left holding a stale
`brainray/raylib.so` that was suddenly **unignored**, and `git add` swept it
in. Worth deleting from the branch, and possibly worth a defensive
`*.so` line in `.gitignore` so a directory rename can't un-ignore build output
again.

## 2. The new fixture never runs

Inline. `pytest -k cap_returning_call_in_condition` → `481 deselected`, and
the suite total is identical to `main`'s.

# VERDICT

The change itself is a genuine correctness fix on a fail-open path, which is
the worst kind to leave in. It just currently ships with a binary it
shouldn't and a test that doesn't execute — so nothing stops the next person
from reverting the `VAR_BOOL` case and seeing green.

leo-aa88 (commented):
# APPROVE

Both findings closed, and the `.gitignore` fix is better than removing the file.

## Verified

**No `.so` is tracked anywhere** — `git ls-tree -r HEAD | grep -c '\.so$'` → 0.

**The fixture now runs, and it guards the fix.** I mutation-tested the claim
rather than taking it: deleted the `VAR_BOOL` arm from
`unbox_native_numeric_result()` and the fixture goes red.

```
$ pytest -q -k cap_returning_call_in_condition
1 passed, 481 deselected          # before the mutation
1 failed, 481 deselected          # after
```

That's the property that matters here. Because the original bug failed *open* —
error on stderr *and* the branch silently not taken — a fixture asserting only
the diagnostic would have passed against the broken build. This one pins which
branch ran, so it can't. 482 collected, `format-check` clean.

**Blanket `*.so` is the right call.** The reasoning generalises past this
incident: a path-specific ignore cannot survive a directory rename, and #318
proved it by moving `brainray/raylib.so` → `rayrot/raylib.so` and silently
un-ignoring everyone's pre-rename artifact. Folding `libstdrot.so` and
`rayrot/raylib.so` into one pattern removes the class rather than the instance,
and it matches what `tests/.gitignore` already did. Recording the incident in
the comment is what stops someone "tidying" it back into per-path entries.

## One housekeeping note

`CONFLICTING` — `tests/expected_results.json` only, the usual tail-append.
Needs a rebase before merge; nothing to re-review after it.

# VERDICT

Neither finding was in the code under test, and the response treated them that
way: the artifact removed at the class level rather than the file level, and the
fixture wired up *and* checked for whether it actually fails. Rebase and merge.

leo-aa88 (commented):
# APPROVE (unchanged)

Rebase only — verified rather than assumed. `ast.c`, `.gitignore` and the
fixture are all byte-identical to the revision I approved, no `.so` is tracked,
and the `expected_results.json` entry survived.

I checked the conflict resolution specifically, since that file was the
conflict and a tail-append merge is the easy place to lose a line:

```
valid JSON, 408 entries
lost from main:   []
changed vs main:  []
added:            ['cap_returning_call_in_condition']
```

Nothing dropped, nothing altered, exactly one addition. Worth doing explicitly
here of all PRs — silently losing an entry during resolution would make some
other fixture stop running with nothing going red, which is the precise failure
mode this PR was fixing.

483 collected. Good to merge.

leo-aa88 (commented):
# APPROVE (unchanged)

```
valid JSON, 412 entries
lost from main:   []
changed vs main:  []
added:            ['cap_returning_call_in_condition']
```

with nothing going red.

Re-confirmed after the rebase: `pytest -k cap_returning_call_in_condition` still
selects and passes, 487 collected, `format-check` clean.

Approval stands.

Issue #274, opened 2026-08-27
[Bug]: grind (continue) is documented as implemented but is a parse error in every loop


leo-aa88:
### Description

The keyword table in `README.md` lists `grind` → `continue` as **✅ Implemented**, but `continue` is not actually implemented in the grammar. `lang.l` maps `grind` to a `CONTINUE` token (`lang.l:499`) and `lang.y` declares that token (`lang.y:392`), but **no grammar production ever uses `CONTINUE`**. As a result any use of `grind` — in a `flex` (for) or `goon`/`mewing` (while/do) loop — is a hard parse error, and the program fails before it runs.

By contrast `bruh` (break) works correctly, so this is specific to `grind`.

### Minimal reproduction code

```c
skibidi main {
    flex (rizz i = 0; i < 5; i = i + 1) {
        edgy (i == 2) { grind; }   🚽 skip 2
        yapping("%d", i);
    }
}
```

### Expected behavior

Loop prints `0 1 3 4` — the `grind` skips the rest of the iteration when `i == 2`, matching C's `continue` and the README's "✅ Implemented" claim.

### Actual behavior

```shell
Error: syntax error, unexpected CONTINUE at line 2
Parsing failed
```

Same result in a `goon` (while) loop. `grind` is unusable in every context.

### Brainrot version / commit

dcddf726a3f818c9d4e1eadc040e7a91aa9ba99f

### Operating system

WSL2 (Ubuntu), x86-64

### Additional context

Two ways to resolve, and they should not diverge:
1. Implement `continue` — add a `CONTINUE SEMICOLON` statement production and the corresponding loop-continuation logic (analogous to how `bruh`/break uses the loop jump buffer), or
2. If it is intentionally unimplemented, correct the `README.md` keyword table to mark `grind` as ❌ (like `cringe`/goto) so the documented surface matches reality.

The lexer already tokenizes `grind`, so today it is the worst of both worlds: reserved (can't be used as an identifier) yet non-functional.

PR #325, opened 2026-08-30
fix: bussin from inside a loop or switch reaches its own function (#319)

### LINKED ISSUE #319: [Bug]: `bussin` inside a loop is caught by the loop's jump buffer — the program dies with "No scope to exit" and prints nothing

leo-aa88:
## Description

`bussin` inside a `flex`/`goon`/`mewing`/`ohio` killed the process — **empty
stdout**, exit 1, and *"Error: No scope to exit"* pointing at a line that was
not the return.

```c
rizz first_match(rizz n) {
    flex (rizz i = 0; i < n; i = i + 1) {
        edgy (i == 2) { bussin i; }
    }
    bussin 0 - 1;
}
```

### Mechanism

`LONGJMP()` jumps to the **innermost** jump buffer. Seven of the eight push
sites are **break** targets — `for`/`while`/`do-while`/`switch`, duplicated
across the legacy `ast.c` path and the `interpreter.c` visitor path — and
exactly one, `execute_function_call()`, is the **return** target.

So a return from inside a loop landed in the *loop's* `setjmp`, which skipped
its body, popped, and fell through to the statement **after** the loop, inside
a function that was supposed to have returned. `bussin` silently behaved as
`break`. The mismatched scope stack then reached `exit_scope()` with a NULL
scope, which `exit(1)`s directly — so nothing was flushed and the program
printed nothing at all.

### Fix

`JumpBuffer` now records whether it is a function frame, and
`handle_return_statement()` discards intervening break targets before jumping.

Popping there rather than teaching each of the seven loop sites to re-throw
keeps the knowledge in one place — and it is **required, not merely tidier**:
the longjmp abandons those `setjmp` frames, so their own `POP_JUMP_BUFFER()`
never runs and their buffers would leak. The scopes those frames entered are
already gone; the existing loop just above the fix unwinds every non-function
scope first.

## Related Issue

Fixes #319

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement
- [ ] Refactor

## Checklist

**Test coverage.** `test_cases/return_from_inside_loop.brainrot` covers `flex`,
`goon`, **two nested loops** (the return has to discard *both* buffers), and
`ohio` in its matching and default arms.

The adversarial part is *what* it asserts. It pins the returned **values**, not
just survival — and that distinction is the whole point: `bussin`-as-`break`
fell through to each function's trailing `bussin 0 - 1;`, so once the scope
stack stopped dying, every one of these would quietly return `-1`. A fixture
checking only for a clean exit would have missed the actual bug entirely.

`bruh` is exercised alongside, since the fix works by *discarding break
targets* — `break` is the thing most at risk of regressing. **Mutation-verified:**
removing the pop fails the fixture.

**Verification.** 482 tests pass (481 → +1). Full valgrind sweep **421/421

## Notes

- **`grind` (continue) is not covered** — it is a parse error (#274), so that
  case could not be written.
- This is the second of #312, #313 and #319. #313 is PR #324; #312 is separate,
  as it touches semantic scoping rather than control flow.

### DIFF
--- a/ast.c
+++ b/ast.c
@@ -7159,7 +7159,7 @@ void execute_function_call(const String name, ArgumentList *args)
         return;
     }
 
-    PUSH_JUMP_BUFFER();
+    PUSH_FUNCTION_JUMP_BUFFER();
     if (setjmp(CURRENT_JUMP_BUFFER()) == 0)
     {
         /* Use visitor pattern instead of old AST execution for function bodies
@@ -7609,6 +7609,28 @@ void handle_return_statement(ASTNode *expr)
     while (current_scope && !current_scope->is_function_scope)
     {
         exit_scope();
+    }
+
+    /* Discard any BREAK targets between here and this function's own
+       frame (#319). LONGJMP() goes to the innermost buffer; if a `bussin`
+       fires inside a loop or switch, that is the loop's, not the
+       function's. Landing there made the loop's setjmp return non-zero,
+       so it skipped its body, popped, and fell through to the statement
+       AFTER the loop -- inside a function that was supposed to have
+       returned. `bussin` silently behaved as `break`, and the next
+       exit_scope() then found a NULL scope and exit(1)'d the process
+       before anything was flushed, so a search-and-return loop produced
+       no output at all.
+
+       Popping here rather than teaching each loop to re-throw keeps the
+       knowledge in one place: the loops' setjmp frames are abandoned by
+       the longjmp below and their own POP_JUMP_BUFFER() never runs, which
+       is exactly why their buffers have to be released now instead. The
+       scopes those frames entered are already gone -- the loop above
+       unwound every non-function scope before this point. */
+    while (jump_buffer && !jump_buffer->is_function)
+    {
+        POP_JUMP_BUFFER();
     }
 
     // skibidi main function do not have jump buffer
--- a/ast.c
+++ b/ast.c
@@ -7627,10 +7627,44 @@ void handle_return_statement(ASTNode *expr)
        the longjmp below and their own POP_JUMP_BUFFER() never runs, which
        is exactly why their buffers have to be released now instead. The
        scopes those frames entered are already gone -- the loop above
-       unwound every non-function scope before this point. */
-    while (jump_buffer && !jump_buffer->is_function)
+       unwound every non-function scope before this point.
+
+       ── Only drain when there IS a function frame to drain to ──────────
+       "abandoned by the longjmp below" assumes the longjmp happens, and
+       in `main` it does not. execute_function_call() is the only site
+       that pushes an is_function buffer, and `main`'s body is not run
+       through it (skibidi_function reduces to a bare statement list --
+       see semantic_analyzer.c's find_symbol() comment), so nothing below
+       a `bussin` in `main` is ever is_function.
+
+       Draining unconditionally there empties the whole stack, leaves
+       jump_buffer NULL, and skips the LONGJMP() entirely -- so execution
+       falls back into the loop body with its scopes already unwound and
+       every subsequent statement reports against a dead scope. That
+       turned `main`'s single "No scope to exit" into a four-error cascade
+       (PR #325 review). Checking first is strictly non-regressive: every
+       real function still gets the fix, and `main` keeps the one
+       diagnostic it had before.
+
+       `bussin` in `main` remains wrong either way -- it should end the
+       program, and even a plain `bussin 3;` there exits 0 rather than 3,
+       so the value is not plumbed. That is a separate, larger change than
+       this one and is deliberately not folded in here. */
+    bool has_function_frame = false;
+    for (JumpBuffer *jb = jump_buffer; jb; jb = jb->next)
     {
-        POP_JUMP_BUFFER();
+        if (jb->is_function)
+        {
+            has_function_frame = true;
+            break;
+        }
+    }
+    if (has_function_frame)
+    {
+        while (jump_buffer && !jump_buffer->is_function)
+        {
+            POP_JUMP_BUFFER();
+        }
     }
 
     // skibidi main function do not have jump buffer
--- a/ast.h
+++ b/ast.h
@@ -56,6 +56,18 @@ typedef struct
 typedef struct JumpBuffer
 {
     jmp_buf data;
+    /* True only for the buffer execute_function_call() pushes around a
+       function body. Every other push -- for/while/do-while/switch -- is a
+       BREAK target.
+       LONGJMP() always jumps to the innermost buffer, which is right for
+       `bruh` and wrong for `bussin`: a return from inside a loop landed in
+       the LOOP's setjmp, which then skipped its body, popped, and fell
+       through to the statement after the loop inside a function that was
+       supposed to have returned -- so `bussin` silently behaved as `break`
+       and the mismatched scope stack killed the process on the next
+       exit_scope() (#319). This flag is what lets handle_return_statement()
+       discard the intervening break targets and reach its own frame. */
+    bool is_function;
     struct JumpBuffer *next;
 } JumpBuffer;
 
@@ -948,10 +960,23 @@ extern Arena arena;
     } while (0)
 
 /* Macros for handling jump buffer */
+/* A BREAK target: for/while/do-while/switch. See JumpBuffer.is_function. */
 #define PUSH_JUMP_BUFFER()                                                     \
     do                                                                         \
     {                                                                          \
         JumpBuffer *jb = SAFE_MALLOC(JumpBuffer);                              \
+        jb->is_function = false;                                               \
+        jb->next = jump_buffer;                                                \
+        jump_buffer = jb;                                                      \
+    } while (0)
+
+/* A RETURN target: the frame execute_function_call() sets up around a
+   function body. Exactly one push site uses this. */
+#define PUSH_FUNCTION_JUMP_BUFFER()                                            \
+    do                                                                         \
+    {                                                                          \
+        JumpBuffer *jb = SAFE_MALLOC(JumpBuffer);                              \
+        jb->is_function = true;                                                \
         jb->next = jump_buffer;                                                \
         jump_buffer = jb;                                                      \
     } while (0)
--- a/test_cases/return_from_inside_loop.brainrot
+++ b/test_cases/return_from_inside_loop.brainrot
@@ -67,6 +67,21 @@ rizz fromSwitch(rizz k) {
     bussin 0 - 1;
 }
 
+🚽 A BARE `bussin;` (void return, #320) from inside a loop -- the
+🚽 intersection of that change and this one, which neither fixture covered
+🚽 on its own. It must leave the function, not the loop: "NOT REACHED"
+🚽 below is what a `bussin`-as-`break` would print.
+skibidi findAndStop(rizz n) {
+    flex (rizz i = 0; i < n; i = i + 1) {
+        edgy (i == 2) {
+            yapping("stopping at %d", i);
+            bussin;
+        }
+        yapping("visiting %d", i);
+    }
+    yapping("NOT REACHED");
+}
+
 🚽 The control case: `bruh` must still escape exactly one loop.
 rizz stillBreaks(rizz n) {
     rizz total = 0;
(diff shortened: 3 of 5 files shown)


ast.c
```diff
@@ -7579,6 +7579,28 @@ void handle_return_statement(ASTNode *expr)
...
+       before anything was flushed, so a search-and-return loop produced
+       no output at all.
+
+       Popping here rather than teaching each loop to re-throw keeps the
+       knowledge in one place: the loops' setjmp frames are abandoned by
+       the longjmp below and their own POP_JUMP_BUFFER() never runs, which
+       is exactly why their buffers have to be released now instead. The
+       scopes those frames entered are already gone -- the loop above
+       unwound every non-function scope before this point. */
+    while (jump_buffer && !jump_buffer->is_function)
```
leo-aa88: **MAJOR.** This loop assumes a function frame exists somewhere below. In `main` it doesn't — `execute_function_call()` is the only site that pushes one, and `main` isn't called through it. So a `bussin` inside a loop in `main` pops *every* buffer, `jump_buffer` ends up NULL, no `LONGJMP()` happens, and execution falls through into the loop body with its scopes already unwound.

```c
skibidi main {
    yapping("before");
    flex (rizz i = 0; i < 5; i = i + 1) {
        edgy (i == 2) { bussin 0; }
    }
    yapping("AFTER LOOP");
    bussin 0;
}
```

`main`:
```
before
Error: No scope to exit at line 8
exit=1
```

this branch:
```
before
Error: Undefined variable at line 8
Error: Undefined variable in get_expression_type at line 8
Error: Assignment to undefined variable at line 8
Error: No scope to exit at line 8
exit=1
```

Both are wrong — `bussin 0` in `main` should just end the program — but this trades one error for a cascade, because the loop keeps running against a torn-down scope before hitting the same terminal failure. That's the "kept going after the failure and let later passes see the wreckage" shape rather than a new crash, so it's not a blocker; it is a strictly worse diagnostic than what's on `main` today.

The guard is small — only drain when there's something to drain to:

```c
bool has_function_frame = false;
for (JumpBuffer *jb = jump_buffer; jb; jb = jb->next)
    if (jb->is_function) { has_function_frame = true; break; }
if (has_function_frame)
    while (jump_buffer && !jump_buffer->is_function)
        POP_JUMP_BUFFER();
```

That restores `main`'s previous single-error behaviour exactly, and leaves the fix untouched for every function. If you'd rather make `bussin`-in-`main` actually work, that's a bigger question (a plain `bussin 3;` in `main` already exits 0 rather than 3, so the return value isn't plumbed either) and belongs in its own issue — but then it's worth saying so in the comment, since "the loops' setjmp frames are abandoned by the longjmp below" silently assumes the longjmp happens.

Either way the fixture should carry the `main` case; right now every scenario in it is inside a helper function, which is why this didn't show up.

leo-aa88 (commented):
# COMMENT

The fix is real and the diagnosis in the commit message is accurate. One shape
it doesn't cover gets *noisier* than before.

**Verified.** Your own fixture on `main`:

```
Error: No scope to exit at line 92
(no other output at all)
```

and on this branch:

```
for      2
while    30
nested   102
switch   11
switch d 99
break    1
done
```

Zero output to six correct lines — the "killed before anything was flushed"
description is exactly right. `bruh` still breaks (the fixture's `break 1`
covers it), and a function called from inside a loop still finds its own frame
rather than the caller's, which is the case the `is_function` flag exists for.

Splitting `PUSH_FUNCTION_JUMP_BUFFER()` out rather than adding a parameter to
the existing macro is the right call — there is exactly one function push site,
and a flag argument at ~15 call sites would be noise.

## The finding

Inline. `bussin` inside a loop in `main` has no function frame to stop at, so
the new loop drains the whole stack and execution falls through with dead
scopes. One error becomes four.

# VERDICT

Good fix on a genuinely nasty bug — `bussin` silently behaving as `bruh` and
then killing the process before flushing is about as bad as a control-flow bug
gets. The remaining case is the one where the invariant the loop assumes ("a
function frame exists below me") doesn't hold, and it's a one-line guard.

ast.c
leo-aa88: Still unguarded at this head, and the rebase onto #320 gave it a second entry point.

`main` is the one function whose body isn't run through `execute_function_call()`, so nothing below this loop is ever `is_function` — it drains the whole stack, `jump_buffer` ends NULL, no `LONGJMP()` fires, and the loop body keeps executing against scopes that were already unwound.

Both spellings now:

```
bussin 0;   inside a loop in main → before / Undefined variable ×3 / No scope to exit
bussin;     inside a loop in main → before / Undefined variable ×3 / No scope to exit
```

`main` branch gives one error for the first and rejects the second at parse time. So the net effect of this PR plus #320 on that shape is: one diagnostic became four, and the bare form joined it.

Same suggestion as before — only drain when there's something to drain to:

```c
bool has_function_frame = false;
for (JumpBuffer *jb = jump_buffer; jb; jb = jb->next)
    if (jb->is_function) { has_function_frame = true; break; }
if (has_function_frame)
    while (jump_buffer && !jump_buffer->is_function)
        POP_JUMP_BUFFER();
```

Strictly non-regressive: every function keeps the fix, `main` keeps its old single-error behaviour. If you'd rather make `bussin` in `main` actually terminate the program, that's the better outcome but a bigger change — a plain `bussin 3;` in `main` already exits 0 rather than 3, so the value isn't plumbed either — and it deserves its own issue rather than being folded in here.

Either way the comment above is worth a sentence: *"the loops' setjmp frames are abandoned by the longjmp below"* assumes the longjmp happens, and in `main` it doesn't. That assumption being implicit is what made this easy to miss.

leo-aa88 (commented):
# REQUEST CHANGES

The new commit is good and orthogonal — it doesn't touch the finding, and the
finding got slightly wider.

**What landed.** Composing with #320 is worth having, and the fixture now
covers a bare `bussin;` from inside a loop in a helper function:

```
visiting 0
visiting 1
stopping at 2
done
```

Noticing that #320 landed underneath and that neither PR's fixture covered the
intersection is exactly the right instinct — it's the same "each feature tested
alone, their intersection tested by neither" gap that #315 hit.

**What didn't.** The pop loop at `ast.c:7599` is unchanged, and `main` still
has no function frame to stop at. Both forms now fail the same way — the new
`bussin;` shape inherits it:

```
skibidi main { flex (...) { edgy (i == 2) { bussin 0; } } ... }
→ before / Undefined variable ×3 / No scope to exit

skibidi main { flex (...) { edgy (i == 2) { bussin; } } ... }
→ before / Undefined variable ×3 / No scope to exit
```

versus one error on `main`. So #320's shape widened the surface of the
unguarded case rather than the fix covering it.

483 pass, `format-check` clean.

# VERDICT

Unchanged from my last review: correct fix for functions, and one shape where
the invariant it assumes — that a function frame exists below — doesn't hold.
The guard is a few lines and the `main` case belongs in the fixture, which
still has every scenario inside a helper.

leo-aa88 (commented):
# APPROVE

The guard is in and I verified both halves of "strictly non-regressive."

**`main` is back to one error.** Was four:

```
bussin 0;  in a loop in main → before / No scope to exit          (was 4 errors)
bussin;    in a loop in main → before / No scope to exit
```

Identical to what `main` branch produces, which is the bar I asked for.

**The fix is untouched for real functions** — the full fixture still runs
end to end:

```
for 2 / while 30 / nested 102 / switch 11 / switch d 99 / break 1
visiting 0 / visiting 1 / stopping at 2 / done
```

484 collected, `format-check` clean.

The comment addition is the part worth keeping. It names the assumption that
was implicit before — *"abandoned by the longjmp below" assumes the longjmp
happens, and in `main` it does not* — and then says why `main` isn't run
through `execute_function_call()`, with a pointer to `find_symbol()`'s comment
for the grammar reason. That's the second time this session that
`skibidi_function` reducing to a bare statement list has been the root of a
bug; having both comments cross-reference it is how the next person finds it in
one hop instead of three.

Scoping `bussin`-in-`main` out explicitly, with the observation that a plain
`bussin 3;` there already exits 0 rather than 3, is the right call — that's a
value-plumbing question, not a jump-buffer one.

# VERDICT

Correct fix, correct scope, and the comment now carries the assumption that
made the gap invisible. Ship it.

leo-aa88 (commented):
# APPROVE (unchanged)

```
valid JSON, 413 entries
lost from main:   []
changed vs main:  []
added:            ['return_from_inside_loop', 'return_from_inside_loop_in_main_fail']
```

with nothing going red.

Re-confirmed after the rebase: `main`'s `bussin`-in-a-loop still produces the
single `No scope to exit`, and the function fixture still runs to `done`. 488
collected, `format-check` clean.

One correction to my own last review: I asked for the `main` case to be carried
in a fixture, and `return_from_inside_loop_in_main_fail` was already there —
pinning `before` / `No scope to exit`. I'd verified that behaviour by hand and
missed the fixture because I only diffed `ast.c`, `ast.h` and
`return_from_inside_loop.brainrot`. The ask was already satisfied; disregard
that part.

Approval stands.

PR #326, opened 2026-08-30
fix: a function's parameter shadows a same-named main local (#312, #321)

### LINKED ISSUE #312: [Bug]: a caller's local shadows a callee's parameter name during semantic analysis, so a correct call is rejected

### LINKED ISSUE #321: [Bug]: a struct parameter resolves members against a same-named caller variable's type
### Description

A struct parameter whose **name collides** with a variable in the *caller's*
scope resolves its members against the **caller's** struct type, not its own.
When the two types differ, a valid member access fails:

```
Error: Struct 'Bag' has no member 'x'
```

`Bag` is the caller's variable's type. `getx`'s parameter is a `gang Vec2`, and
`Vec2` does have `x`. Renaming either the parameter or the caller's variable
makes it work, which is the tell: the callee is reading a type it should not be
able to see.

The error is reported while the callee's body executes, but the *type* used to
resolve `b.x` comes from the caller's `b`. The call then yields `0.0` and the
process exits 0, so a program can carry on with a wrong value.

Found while reviewing #315 — pre-existing, not introduced by it.

### Minimal reproduction code

```c
gang Vec2 { chad x; chad y; };
gang Bag  { rizz n; };

chad getx(gang Vec2 b) {     🚽 parameter named `b`
    bussin b.x;
}

skibidi main {
    gang Vec2 v;
    v.x = 3.0;

    gang Bag b;              🚽 caller ALSO has a `b`, of a different type
    b.n = 1;

    yapping("%.1f", getx(v));
    bussin 0;
}
```

Rename either `b` and it prints `3.0`.

### Expected behavior

`3.0`. A parameter shadows any caller variable of the same name, so `b.x`
inside `getx` resolves against `gang Vec2` regardless of what `b` means in the
caller.

### Actual behavior

```shell
$ ./brainrot collide.brainrot
Error: Struct 'Bag' has no member 'x' at line 5
exit=0
```

With the caller's variable renamed to `other`:

```shell
$ ./brainrot control.brainrot
3.0
```

### Brainrot version / commit

`f98e90fa844937db93242a5564db84ac041d307d`

### Operating system

### Additional context

**Where to look.** The parameter is bound correctly — `enter_function_scope()`
(`ast.c`) creates the variable in the fresh scope and copies the tag from
`curr_param->desc.struct_name`. What goes wrong is the *tag lookup* during
member access: something on that path resolves the name through a scope that
still reaches the caller's binding, or reads a struct tag recorded outside the
callee's scope. `resolve_struct_access()`'s `NODE_IDENTIFIER` arm and the
static tag inference in `semantic_analyzer.c` are the two places that decide a
tag from a name, and they are the ones worth checking first.

**Suggested testing.** A `test_cases/*.brainrot` fixture with the two structs
above, printing the call's result rather than discarding it — so an
implementation that reports the error and then binds garbage still fails. Worth
covering three shapes, since only the first is currently broken:

- parameter and caller variable of **different** struct types (this bug)
- same name, **same** struct type (works today — should keep working)
- same name where the caller's variable is a **scalar** (worth pinning either
  way)

**Note on scope.** This is about struct *tag* resolution specifically. Plain
scalar parameters shadowing caller variables work correctly.

leo-aa88:
## Description

Two issues, **one root cause**.

| | Repro | Error |
| --- | --- | --- |
| #312 | `chad wid(gang P *bm)` called with main's `gang P bm;` in scope | *"expected pointer to struct/union 'P' (level 1), got struct pointer level 0"* — the **pointer level** came from main's variable |
| #321 | `chad getx(gang Vec2 b)` called with main's `gang Bag b;` in scope | *"Struct 'Bag' has no member 'x'"* — the **tag** came from main's variable, for a parameter whose own type *does* have `x` |

Both reproduce, both vanish if either variable is renamed, and both name a type
that is correct inside the callee — so the error sends you looking in the wrong
function. It also made **parameter names part of a library's public surface**: a
caller picking a natural local name for a struct it owns got a compile error
inside a file it did not write.

### Root cause

`skibidi_function` (`lang.y`) reduces to a **bare statement list**:

```
skibidi_function:
    SKIBIDI MAIN LBRACE statements RBRACE
        { $$ = $4; }
```

No function-def node — so main's body is analyzed with no
`current_function_name`, and `add_symbol()` records its locals with
`function_name == {0}`, **indistinguishable from a real global**.
`find_symbol()` accepted `is_global || same_function` and returned whichever
matched *first*, so main's local could answer a lookup for a same-named
parameter in an unrelated function.

I isolated it to confirm: the same collision against a **non-main** function's
local does **not** happen, because those locals are tagged with their function.

### Fix

`find_symbol()` now returns an in-function entry outright, and falls back to a
global only when no in-function declaration of that name exists.

That is the right rule independent of how `main` is represented — a parameter
or local shadows a global in any language with lexical scope — so it fixes the
symptom without depending on the grammar quirk that exposed it, and **without
changing what `skibidi main` parses to**.

## Related Issue

Fixes #312
Fixes #321

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement
- [ ] Refactor

## Checklist

**Test coverage.** `test_cases/param_shadows_main_local.brainrot` covers both
reported shapes (differing pointer level, differing struct tag) plus a scalar
one.

It prints the **values**, which is what makes it adversarial in both
directions: `getx(v)` giving `3.0` and `use(&bm)` giving `W` prove the callee
saw *its own* parameter, while main printing its own `b.n` / `n` / `bm.w`
afterwards proves the fix did not overcorrect and make main's locals
unreachable. **Mutation-verified:** restoring the first-match behavior fails the
fixture.

Also checked by hand that the fix does not go the other way: an undefined
variable is still caught, a parameter sharing a name *and* type with a main
local still works, and a scalar parameter still shadows a scalar local.

**Verification.** 482 tests pass (481 → +1). Full valgrind sweep **421/421

## Note

Last of #312, #313 and #319. #313 is PR #324, #319 is PR #325 — all three are
independent and touch different subsystems, so they can merge in any order.
This one additionally closes #321, which I had filed separately before finding
it shared #312's cause.

### DIFF
--- a/semantic_analyzer.c
+++ b/semantic_analyzer.c
@@ -3036,6 +3036,7 @@ SymbolEntry *find_symbol(SemanticAnalyzer *analyzer, const String name)
         return NULL;
 
     SymbolEntry *entry = analyzer->symbol_table;
+    SymbolEntry *fallback = NULL;
 
     while (entry)
     {
@@ -3059,15 +3060,52 @@ SymbolEntry *find_symbol(SemanticAnalyzer *analyzer, const String name)
                 analyzer->current_function_name.data &&
                 strcmp(entry->function_name.data,
                        analyzer->current_function_name.data) == 0;
-            if (is_global || same_function)
+            if (same_function)
             {
-                return entry; /* Symbol is accessible */
+                /* An entry belonging to the function being analyzed wins
+                   outright: it is the innermost declaration of that name,
+                   and it SHADOWS any global of the same name (#312, #321).
+
+                   This mattered because `skibidi main`'s locals are
+                   globals here. `skibidi_function` (lang.y) reduces to a
+                   bare statement list -- `$$ = $4`, with no function-def
+                   node -- so main's body is analyzed with no
+                   current_function_name, and add_symbol() records its
+                   locals with function_name == {0}, indistinguishable
+                   from a real global. Returning whichever matched first
+                   therefore let `gang P bm;` in main answer a lookup for
+                   the parameter `bm` inside an unrelated function:
+
+                     chad wid(gang P *bm) { ... }   <- pointer_level 1
+                     skibidi main { gang P bm; ... }  <- pointer_level 0
+
+                   which rejected a correct call with "expected pointer to
+                   struct/union 'P' (level 1), got struct pointer level 0",
+                   naming a type that is correct inside the callee; and in
+                   the member-access form gave "Struct 'Bag' has no member
+                   'x'" for a parameter whose own type does have it.
+                   Renaming either variable made both go away, which is the
+                   tell.
+
+                   Preferring the in-function entry is the right rule
+                   regardless of how main is represented -- a parameter or
+                   local shadows a global in any language with lexical
+                   scope -- so this fixes the symptom without depending on
+                   the grammar quirk that exposed it. */
+                return entry;
+            }
+            if (is_global && !fallback)
+            {
+                /* Remember the first global match, but keep looking: an
+                   in-function entry later in the table must still win. */
+                fallback = entry;
             }
         }
         entry = entry->next;
     }
 
-    return NULL; /* Symbol not found or not accessible */
+    /* No in-function declaration of this name -- a global is correct. */
+    return fallback;
 }
 
 void free_symbol_table(SymbolEntry *symbols)
--- a/semantic_analyzer.c
+++ b/semantic_analyzer.c
@@ -3097,7 +3097,38 @@ SymbolEntry *find_symbol(SemanticAnalyzer *analyzer, const String name)
             if (is_global && !fallback)
             {
                 /* Remember the first global match, but keep looking: an
-                   in-function entry later in the table must still win. */
+                   in-function entry later in the table must still win.
+
+                   The FIRST match, not the last, deliberately: that is
+                   the exact entry the old `return entry` handed back, so
+                   a program with no in-function declaration of the name
+                   behaves bit-for-bit as before. This is a refinement of
+                   the old rule, not a replacement for it.
+
+                   ── "Keep looking" is not free, and was measured ───────
+                   It costs the early exit: every lookup that resolves to
+                   a global now walks the table to the end. That is not a
+                   rare path here -- since main's locals ARE globals (see
+                   above), every variable reference inside main takes it.
+
+                   Timed on a generated main with N declarations plus N/4
+                   assignments, same build flags on both sides:
+
+                     N=2000   0.07s -> 0.11s
+                     N=6000   0.53s -> 0.93s
+
+                   so ~1.6-1.8x, a constant factor rather than a new
+                   complexity class: 3x the input costs ~8x the time on
+                   BOTH sides, because this lookup is already quadratic
+                   either way. It makes an existing scaling problem
+                   somewhat worse; it does not create one.
+
+                   Stated here because "keep looking" reads as free, and
+                   re-adding a `return entry` for the global case as an
+                   obvious optimization would quietly reintroduce #312.
+                   If the symbol table ever gets an index -- it is a
+                   linked list today, and lib/hm.c is right there -- this
+                   is one of the call sites that would benefit most. */
                 fallback = entry;
             }
         }
--- a/test_cases/param_shadows_main_local.brainrot
+++ b/test_cases/param_shadows_main_local.brainrot
@@ -1,0 +1,75 @@
+🚽 Test case: issues #312 and #321 -- a function's parameter must shadow a
+🚽 same-named variable in `skibidi main`.
+🚽
+🚽 Both were the same root cause. `skibidi_function` (lang.y) reduces to a
+🚽 BARE STATEMENT LIST -- `$$ = $4`, no function-def node -- so main's body
+🚽 is analyzed with no current_function_name, and add_symbol() records its
+🚽 locals with function_name == {0}: indistinguishable from a real global.
+🚽 find_symbol() returned whichever entry matched first, so main's local
+🚽 could answer a lookup for a parameter of the same name inside an
+🚽 unrelated function.
+🚽
+🚽 The two symptoms looked unrelated:
+🚽
+🚽   #312  the POINTER LEVEL came from main's variable, so a correct call
+🚽         was rejected -- "expected pointer to struct/union 'P' (level 1),
+🚽         got struct pointer level 0" -- naming a type that is correct
+🚽         inside the callee, which sends you looking in the wrong function.
+🚽
+🚽   #321  the STRUCT TAG came from main's variable, so a valid member
+🚽         access failed -- "Struct 'Bag' has no member 'x'" for a
+🚽         parameter declared `gang Vec2`, which does have `x`.
+🚽
+🚽 Renaming either variable made both disappear, which is the tell. It also
+🚽 made parameter names part of a library's public surface: a caller
+🚽 picking a natural local name for a struct it owns got a compile
+🚽 error inside a file it did not write.
+🚽
+🚽 ── Why the values are printed ────────────────────────────────────────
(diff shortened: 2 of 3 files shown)


semantic_analyzer.c
```diff
@@ -3059,15 +3060,52 @@ SymbolEntry *find_symbol(SemanticAnalyzer *analyzer, const String name)
...
+                   tell.
+
+                   Preferring the in-function entry is the right rule
+                   regardless of how main is represented -- a parameter or
+                   local shadows a global in any language with lexical
+                   scope -- so this fixes the symptom without depending on
+                   the grammar quirk that exposed it. */
+                return entry;
+            }
+            if (is_global && !fallback)
```
leo-aa88, line 3097: **MINOR**, and I measured it before raising it so it isn't a hand-wave.

Recording the global and continuing means `find_symbol()` no longer early-returns — every lookup that resolves to a "global" now walks the symbol table to the end. And in Brainrot that isn't a rare path: because `main`'s locals *are* globals here (as your comment explains), every variable reference inside `main` takes it.

Generated a `main` with N declarations plus N/4 assignments:

| N | `main` branch | this branch |
| --- | --- | --- |
| 2000 | 0.07 s | 0.11 s |
| 6000 | 0.62 s | 0.97 s |

So ~1.55–1.6× at both sizes — a stable constant factor, not a new complexity class. The lookup is already quadratic on both branches (3× the input costs ~9× the time either way), so this makes an existing scaling problem somewhat worse rather than creating one.

Not worth restructuring for. Two things worth doing:

- Say it in the comment. "Keep looking" reads as free, and it isn't; a sentence noting the lost early-exit and that it was measured stops someone from re-adding a `return entry` for the global case as an "obvious optimization" and quietly reintroducing #312.
- If the symbol table ever gets an index (it's a linked list today, and `lib/hm.c` is right there), this is one of the call sites that would benefit most — worth a note on whatever issue tracks that.

One thing I checked and liked: keeping the **first** global match rather than the last means the global-only path returns exactly the entry the old code did. That's what makes this a strict refinement instead of a behaviour change with a bug fix attached to it.

leo-aa88 (commented):
# APPROVE

This is the bug I tripped over writing probes during #315, and the fix is the
right rule rather than a patch on the symptom.

**Verified against my original repro:**

```
chad getx(gang Vec2 b) { bussin b.x; }
skibidi main { gang Bag b; ... getx(v); }
```

`main`: *"Struct 'Bag' has no member 'x'"* — twice, for both the by-value and
pointer forms. This branch: `getx: 7.0`, `wid: 7.0`, and `main`'s own `b.n`
and `bm.n` still read 5 and 6. Your fixture passes too.

**No regressions in what I could reach.** A function referencing a name that
exists only in `main` still fails at runtime exactly as before (identical
output on both branches), and a block-scoped redeclaration inside a function is
still rejected outright — so the "an in-function entry from a closed block now
outranks a global" case I went looking for isn't reachable. Preserving the
first-match choice for the fallback also means the global-only path is
bit-for-bit the old behaviour.

481 pass, `format-check` clean.

The comment is the valuable part: recording that `skibidi_function` reduces to
a bare statement list with no function-def node, so `main`'s locals are
indistinguishable from globals to `add_symbol()`, is the fact that makes this
bug comprehensible — and stating the rule holds "regardless of how main is
represented" is what keeps the fix correct if that grammar quirk is ever
fixed.

## One measured note

Inline: `find_symbol()` can no longer early-return on a global match. I
measured it rather than hand-waving — it's a ~1.6× constant factor, not a
complexity change.

# VERDICT

Right rule, honest comment, real fixture. Ship it.

leo-aa88 (commented):
# APPROVE (unchanged)

Rebased onto #320, and I checked that's all it is: `semantic_analyzer.c` is
byte-identical to the revision I approved, and the change against `main` is the
same three files.

Re-verified on the new base rather than assuming the rebase was clean —
my original repro still passes:

```
getx: 7.0
wid : 7.0
main b.n=5 bm.n=6
```

and the fixture still produces `getx 3.0 / use W / twice 6 / mine 41 100 1.0`.
483 pass, `format-check` clean.

Nothing further from me. The measured `find_symbol()` note from my last review
is still worth a sentence in the comment when convenient, but it isn't
blocking and doesn't need another round.

leo-aa88 (commented):
# APPROVE

Comment-only — verified, zero code lines changed — and it says the right thing.

Recording that the lost early-exit was **measured** rather than estimated is
what makes the note useful; and calling out that keeping the *first* global
match (not the last) is what makes the global-only path bit-for-bit identical
to the old `return entry` is the detail that stops someone "simplifying" it
later and quietly reintroducing #312.

483 collected, `format-check` clean. Nothing further.

leo-aa88 (commented):
# APPROVE (unchanged)

```
valid JSON, 412 entries
lost from main:   []
changed vs main:  []
added:            ['param_shadows_main_local']
```

with nothing going red.

Re-confirmed after the rebase: the fixture still produces
`getx 3.0 / use W / twice 6 / mine 41 100 1.0`. 487 collected, `format-check`
clean.

Approval stands.

leo-aa88 (commented):
# APPROVE (unchanged)

Rebased again, onto merged #324 and #325. Pure rebase — `semantic_analyzer.c`
and the fixture are byte-identical to the revision I approved.

`expected_results.json` merge is clean, which is the third time this file has
been the conflict for this PR and so worth re-checking each time:

```
valid JSON, 415 entries
lost from main:   []
changed vs main:  []
added:            ['param_shadows_main_local']
```

490 collected, fixture still gives `getx 3.0 / use W / twice 6 /
mine 41 100 1.0`, `format-check` clean.

Last of the batch. Nothing further from me.

PR #327, opened 2026-08-30
feat(stdrot): string stdlib v1 -- yaplen, yapcat, yapcmp, yapidx (#251)

### LINKED ISSUE #251: [Feature]: String stdlib v1 — yaplen / yapcat / yapcmp / yapidx + s[i] and s[i:j]
## Problem / motivation

Brainrot can *hold* strings (`rant`, backed by the length-prefixed
`String { char *data; size_t len; }` in [`lib/string_value.h`](lib/string_value.h))
and print them (`yapping`), but it can do essentially **nothing else** with
them. There is no way to measure a string, join two, pull out one character, or
find a substring — all of which are table stakes for writing real programs
(parsing input, formatting output, building up messages). The FFI substrate to
expose C-level string helpers as builtins already exists: the registry marshals
`STDROT_STRING` args and returns, and `STDROT_CSTRING` gives a strictly
non-escaping `const char *` view for calling into libc (see
[`stdrot/stdrot_api.h`](stdrot/stdrot_api.h)).

This issue tracks **v1** of a minimal string standard library: the irreducible
core that everything else (split/join/replace/trim/case, `strconv`) will build
on. It is intentionally small and unblocked — none of it depends on the
still-unsettled string-array return path flagged in the ROADMAP.

## Proposed solution

Six operations. Four are additive `stdrot/` builtins; two are **language
syntax** and touch `lang.l`/`lang.y`.

| # | Brainrot | C equivalent | Go equivalent | Kind |
|---|----------|--------------|---------------|------|
| 1 | `yaplen(s)` | `strlen` | `len(s)` | builtin |
| 2 | `yapcat(a, b)` | `strcat` / `snprintf` | `a + b` / `strings.Join` | builtin |
| 3 | `s[i:j]` | `memcpy` + bounds | `s[i:j]` | **syntax** |
| 4 | `yapcmp(a, b)` | `strcmp` | `strings.Compare` | builtin |
| 5 | `yapidx(hay, needle)` | `strstr` | `strings.Index` | builtin |
| 6 | `s[i]` | `s[i]` | `s[i]` | **syntax** |

### Semantic model (applies to all six)

- **Byte semantics, ASCII.** Length, indices, and slice bounds are all in
  **bytes**, not runes. UTF-8 / codepoint-aware variants are explicitly **out
  of scope for v1** (see below). This matches C and Go's `[]byte` view and
  keeps the length-prefixed `String` a straight byte buffer.
- **Immutable / return-new.** Nothing mutates its argument in place. `yapcat`
  and `s[i:j]` allocate a fresh `String` (arena-owned, via `lib/arena.c` /
  `lib/mem.c` as the surrounding stdlib code does). This matches Go's model and
  sidesteps the `STDROT_CSTRING` non-escaping ownership rule.
- **Out-of-bounds is a runtime error**, not UB and not silent clamping —
  raised the same way the existing builtins report (`ragequit`/`bet`-style,
  with `g_exec_context` line number), so ASan/UBSan stay clean.

---

### Builtins (additive, `stdrot/`)

Each is a new `stdrot/<name>.c` self-registering via `STDROT_EXPORT_SIG` with a
typed `StdrotParam` signature (see the header's "HOW TO ADD A NEW STDLIB
FUNCTION" and `stdrot/bet.c` for the pattern).

**1. `yaplen(rant s) -> rizz`** — string length.
- Returns `s.len` as an `int`. Zero for the empty string.
- Signature: `{STDROT_STRING}` → `STDROT_INT`, arity 1..1.

**2. `yapcat(rant a, rant b) -> rant`** — concatenation.
- Returns a new `String` of length `a.len + b.len`; does not touch `a` or `b`.
- `yapcat("", x) == x`, `yapcat(x, "") == x` (value-equal, freshly allocated).
- Signature: `{STDROT_STRING, STDROT_STRING}` → `STDROT_STRING`, arity 2..2.

**4. `yapcmp(rant a, rant b) -> rizz`** — lexicographic compare.
- Byte-wise; returns **-1 / 0 / 1** (normalized, not raw `memcmp` delta).
- Length-aware (must not stop at an embedded `\0` — `String` is not
  NUL-terminated): compare `min(a.len, b.len)` bytes, then shorter < longer.
- Signature: `{STDROT_STRING, STDROT_STRING}` → `STDROT_INT`, arity 2..2.

**5. `yapidx(rant hay, rant needle) -> rizz`** — first index of substring.
- Returns the byte index of the first occurrence of `needle` in `hay`, or
  **-1** if absent. Empty `needle` returns `0` (matches Go's `strings.Index`).
- Signature: `{STDROT_STRING, STDROT_STRING}` → `STDROT_INT`, arity 2..2.

---

### Syntax (touches `lang.l` / `lang.y` — public compatibility surface)

Per AGENTS.md, new grammar is easier to land than changing existing keywords,
but this still needs explicit sign-off before implementation. Neither construct
exists today.

**6. `s[i]` — character indexing.**
- Extends the existing `array_access` production
  ([`lang.y:1967`](lang.y#L1967), `IDENTIFIER multi_dimension_access`) so a
  **string-valued base** is legal, evaluating to a `yap` (char).
- Currently `array_access` resolves only against array-typed variables; the
  interpreter/visitor needs a branch: if the base is a `String`, bounds-check
  `i` against `s.len` and return the byte at `i`.
- `i` in `[0, len)`; otherwise runtime error.

**3. `s[i:j]` — slice / substring.**
- **New syntax** — no slice production exists. Add a grammar rule
  (`expression LBRACKET expression COLON expression RBRACKET`, or the analogous
  extension of `multi_dimension_access`), a new AST node type + constructor in
  `ast.c/h`, semantic-analyzer type rules, and interpreter evaluation.
- Half-open `[i, j)`, Python/Go semantics: result length `j - i`.
- Bounds for v1: require `0 <= i <= j <= len` (out of range → runtime error).
  Open question below on omitted bounds / negative indices.
- Returns a **new** `String` (copied into the arena), consistent with the
  immutability rule — no aliasing into the source buffer.
- Watch for grammar conflicts: `COLON` may already be tokenized (labels/ternary
  — check `lang.l`); reuse the token, keep the rule narrow, and build with
  `bison -Wcounterexamples` (shift/reduce = build failure).

## Example usage

```c
skibidi main() {
    rant greeting = "hello";
    rant name = "world";

    rizz n = yaplen(greeting);           // 5
    rant full = yapcat(greeting, name);  // "helloworld"
    yap first = greeting[0];             // 'h'
    rant sub = greeting[1:4];            // "ell"
    rizz where = yapidx(full, "world");  // 5
    rizz order = yapcmp("apple", "banana"); // -1

    yapping("%d %s %c %s %d %d", n, full, first, sub, where, order);
    bussin 0;
}
```

## Open questions (decide before implementation)

1. **Slice with omitted bounds** — do we want `s[:j]` / `s[i:]` / `s[:]` in v1,
   or require both bounds? (Requiring both is simpler grammar; omitted bounds
   are a nicety we can add later.)
2. **Negative indices** (`s[-1]`, Python-style)? Recommend **no** for v1 —
   keep C/Go byte-offset semantics; revisit with UTF-8 work.
3. **Return type of `yapidx`/`yaplen`** — `rizz` (`int`) is proposed. Confirm
   we don't want a `size_t`-ish unsigned type (Brainrot has none today, so
   `int` with -1 sentinel is the natural fit).
4. **Meme names** — `yaplen` / `yapcat` / `yapcmp` / `yapidx` follow the `yap`
   (char/print) family. Confirm these before they land in the README keyword
   table (a public compatibility surface).

## Out of scope (tracked for v2+)

`split` / `join` (need settled string-array returns — ROADMAP L9), `replace`,
`repeat`, `contains` / `hasPrefix` / `hasSuffix` (thin wrappers over `yapidx`),
`trim`, upper/lower case, and `strconv`-style `atoi`/`itoa`/`ParseFloat`
conversions. And **UTF-8 / rune-aware** variants of everything above.

## Implementation checklist

- [ ] `stdrot/yaplen.c`, `yapcat.c`, `yapcmp.c`, `yapidx.c` (+ `STDROT_EXPORT_SIG`)
- [ ] `lang.l` / `lang.y`: `s[i]` on strings, `s[i:j]` slice rule
- [ ] `ast.c/h`: slice AST node + constructor
- [ ] `semantic_analyzer.c/h`: type rules for string-index (→ `yap`) and slice (→ `rant`)
- [ ] `interpreter.c` / `visitor.c`: runtime eval + bounds checks for both
- [ ] `test_cases/*.brainrot` happy-path + `*_fail.brainrot` out-of-bounds cases, with `tests/expected_results.json` entries
- [ ] `examples/` sample + link from `examples/README.md`
- [ ] Docs: README keyword/builtin table + `docs/the-brainrot-programming-language.md`
- [ ] `make test`, `make valgrind`, `make format-check`, `make cppcheck` all green

leo-aa88:
## Description

Adds the measure / join / compare / search core of the string library as four standard-library builtins. No new syntax, no keywords, no `#cooked`.

```c
rizz yaplen(rant s);                  /* length in BYTES                     */
rant yapcat(rant a, rant b);          /* a joined to b, as a new string      */
rizz yapcmp(rant a, rant b);          /* -1 if a < b, 0 if equal, 1 if a > b */
rizz yapidx(rant hay, rant needle);   /* byte index of needle, or -1         */
```

**None of these is the obvious libc one-liner.** A `rant` is the length-prefixed `String { char *data; size_t len; }` of `lib/string_value.h` and is **not** NUL-terminated, so `strlen`/`strcmp`/`strstr` would each read past the buffer *and* each stop early at an embedded NUL. All four work from the length prefix instead.

Semantics worth stating explicitly, because each is a fork in the road rather than a detail:

- **Bytes, not runes.** `yaplen("é")` is `2`. Codepoint-aware variants are out of scope for v1.
- **Bytes compare unsigned** (memcmp's own rule), so non-ASCII sorts after ASCII. A hand-rolled loop over `char` — signed on x86-64 — would invert that for every UTF-8 byte.
- **`yapcmp` returns exactly `-1`/`0`/`1`**, never a raw memcmp delta, whose magnitude is unspecified beyond its sign. A prefix sorts before what extends it: `yapcmp("app", "apple")` is `-1`.
- **`yapidx` returns `0` for an empty needle**, matching Go's `strings.Index`, so `yapidx(h, n) >= 0` stays a correct "contains" test for *every* needle.
- **Strings are immutable.** `yapcat` touches neither argument and returns a value independent of both.

### The one interesting implementation problem

`yapcat` is the only builtin here that must produce storage the caller did not supply. The host does not free what a native returns, so a plain `malloc` would leak once per call, and the arena is unavailable because `libstdrot.so` links only `stdrot/*.c` + `lib/input.c`. It therefore uses a single file-static buffer, replaced per call — safe because the host copies the returned `STDROT_STRING` out before anything else can run, including a nested `yapcat`.

The final buffer is released from a **library destructor**, which is load-bearing rather than tidiness: `stdrot_unload()` `dlclose()`s this object, unmapping the static itself, so "still reachable through a static" is not true at leak-check time and LeakSanitizer reports a genuine direct leak. The destructor runs as part of that `dlclose`, while the code is still mapped — unlike an `atexit` handler registered from here.

### No defensive argument-type checks

None of the four re-checks its own argument types. `enforce_arg_type()` (stdrot.c) already does, and its comment states that *"a native wrapper is entitled to trust its own StdrotParam blindly"*. The declared signature **is** the check — so the declaration is what gets a test, rather than four unreachable branches no test could cover.

## Related Issue

Fixes #251

Scoped to the builtins. The `s[i]` / `s[i:j]` indexing syntax is deliberately left to a follow-up; the docs say so explicitly rather than leaving it implied.

## Type of Change

- [x] Documentation update
- [ ] Performance improvement
- [ ] Refactor

## Checklist

### Tests

Three fixtures, each pinning something the others cannot:

- **`string_stdlib`** — behaviour, including every empty-operand combination, needle-longer-than-haystack, first-match-wins, the prefix tie-break, and the byte/unsigned semantics.
- **`string_stdlib_result_independence`** — that an *earlier* `yapcat` result survives *later* calls. This is the one property the shared buffer could actually break, and `string_stdlib` would not notice: it consumes each result on the line that produced it, which a shared buffer still satisfies.
- **`string_stdlib_arg_type_fail`** — the declared parameter types. Exporting these with `STDROT_ANY` params would compile, load, run, and hand `yaplen` an int to dereference as a `char*`.

**Both copies** standing between `yapcat`'s buffer and the caller were mutation-tested, and they are not equally important:

| mutation | effect on the independence fixture |
| --- | --- |
| remove `evaluate_expression_string()`'s `safe_strdup` (ast.c) | **dies** with ASan heap-use-after-free |
| remove `finalize_native_string_result()`'s copy (stdrot.c) | **no change** — the first copy still covers it |

The fixture comment records both results, including that the first draft of that comment credited the wrong one.

### Gate results

| gate | result |
| --- | --- |
| `make test` | 485 passed |
| valgrind sweep | 424 fixtures, exit 0, zero leaks or errors |
| `make tidy` | pass (0 user findings) |

One environment note for reviewers reproducing the valgrind gate: on this WSL2 kernel ASan's shadow mapping collides with valgrind (`Shadow memory range interleaves with an existing memory mapping`), so the sweep was run against a build with `-fsanitize` dropped from `CFLAGS`. The sanitized build was restored before `make test`.

### DIFF
--- a/stdrot/yaplen.c
+++ b/stdrot/yaplen.c
@@ -1,0 +1,55 @@
+/* stdrot/yaplen.c -- yaplen(rant s) -> rizz, string length in BYTES.
+ *
+ * Byte semantics, not runes: a Brainrot `rant` is the length-prefixed
+ * String { char *data; size_t len; } from lib/string_value.h, and every
+ * string operation in this v1 (#251) measures and indexes it as a flat
+ * byte buffer, matching C and Go's []byte view. UTF-8/codepoint-aware
+ * variants are explicitly out of scope for v1.
+ *
+ * Reads String.len rather than calling strlen(): a `rant` is NOT
+ * NUL-terminated, so strlen() would run past the buffer for a string
+ * built by any path that does not happen to append one, and would stop
+ * early on a string containing an embedded NUL. The length prefix is the
+ * only correct source.
+ *
+ * No defensive check that args[0] really is a string: the declared
+ * StdrotParam below is enforced by enforce_arg_type() (stdrot.c) before
+ * fn() is entered, and that boundary exists precisely so that "a native
+ * wrapper is entitled to trust its own StdrotParam blindly" -- re-checking
+ * here would be unreachable code that no test could ever cover.
+ */
+#include "stdrot_api.h"
+#include <limits.h>
+#include <stdio.h>
+#include <stdlib.h>
+
+StdrotValue stdrot_yaplen(StdrotValue *args, int argc)
+{
+    (void)argc;
+    /* A String with .data == NULL is representable by the struct even
+       though no path here was found to produce one, so it is treated as
+       empty rather than trusting .len alongside a null buffer. */
+    size_t len = args[0].val.str.data ? args[0].val.str.len : 0;
+
+    /* The ABI's integer carrier is `int` (STDROT_INT), so a string longer
+       than INT_MAX has no representable length. Refusing is the honest
+       answer: silently truncating would report a negative or wrapped
+       length that every caller would then use as a bound. Unreachable in
+       practice, checked anyway because the cost is one comparison. */
+    if (len > (size_t)INT_MAX)
+    {
+        fprintf(stderr,
+                "Error: yaplen: string length %zu exceeds the largest "
+                "representable rizz at line %d\n",
+                len, g_exec_context.line_number);
+        exit(1);
+    }
+
+    return (StdrotValue){.type = STDROT_INT, .val = {.i = (int)len}};
+}
+
+static const StdrotParam yaplen_params[] = {
+    {STDROT_STRING, NULL, 0},
+};
+STDROT_EXPORT_SIG("yaplen", stdrot_yaplen, ((StdrotParam){STDROT_INT, NULL, 0}),
+                  yaplen_params, 1, 1, false);
--- a/docs/brainrot-user-guide.md
+++ b/docs/brainrot-user-guide.md
@@ -376,6 +376,10 @@ Brainrot includes some built-in functions for convenience:
 | **slorp**    | `stdin`     | -            | Reads user input.                                                     |
 | **bet**      | `stderr`    | No           | Tests conditions and terminates with error message if false.          |
 | **gamba**    | -           | -            | Cryptographically safe random integers (OpenSSL `RAND_bytes`).        |
+| **yaplen**   | -           | -            | Length of a `rant`, in bytes.                                         |
+| **yapcat**   | -           | -            | Joins two `rant`s into a new one.                                     |
+| **yapcmp**   | -           | -            | Lexicographic comparison: `-1`, `0` or `1`.                           |
+| **yapidx**   | -           | -            | Byte index of one `rant` inside another, or `-1`.                     |
 
 ## 10.1. yapping
 
@@ -617,6 +621,58 @@ skibidi main {
 
     rizz nonce = gamba();
     yapping("nonce %d", nonce);
+    bussin 0;
+}
+```
+
+---
+
+## 10.9. yaplen, yapcat, yapcmp, yapidx
+
+**Prototypes**
+
+```c
+rizz yaplen(rant s);                  /* length in BYTES                     */
+rant yapcat(rant a, rant b);          /* a joined to b, as a new string      */
+rizz yapcmp(rant a, rant b);          /* -1 if a < b, 0 if equal, 1 if a > b */
+rizz yapidx(rant hay, rant needle);   /* byte index of needle, or -1         */
+```
+
+**Description**
+
+- The v1 string library: measure, join, compare, search. All four are
+  standard-library builtins -- no `#cooked`, no keyword.
+- **Byte-oriented, not character-oriented.** `yaplen("é")` is `2`, because
+  that is two bytes of UTF-8. Comparison and searching work on bytes for the
+  same reason.
+- Bytes compare as **unsigned**, matching C's `strcmp` and Go's
+  `strings.Compare`, so non-ASCII bytes sort after every ASCII one.
+- **Strings are immutable.** `yapcat` modifies neither argument and returns a
+  new string that is independent of both, so an earlier result stays valid
+  after later calls.
+- `yaplen` reads the stored length rather than scanning, so it is O(1) and
+  stays correct for a string containing an embedded NUL.
+- `yapcmp` returns exactly `-1`, `0` or `1`. When one string is a prefix of
+  the other the **shorter sorts first**: `yapcmp("app", "apple")` is `-1`.
+- `yapidx` reports the **first** match. A missing needle gives `-1`; an
+  **empty** needle gives `0`, so `yapidx(h, n) >= 0` is a correct "contains"
+  test for every needle.
+- Not in v1: substring/slice, split, replace, case conversion, and `s[i]`
+  indexing syntax.
+
+### Example
+
+```c
+skibidi main {
+    rant name = yapcat(yapcat("Big", " "), "Chungus");
+
+    yapping("%s", name);              🚽 Big Chungus
+    yapping("%d", yaplen(name));      🚽 11
+    yapping("%d", yapidx(name, " ")); 🚽 3
+
+    edgy (yapidx(name, "Chungus") >= 0) {
+        yapping("certified");
+    }
     bussin 0;
 }
 ```
--- a/docs/brainrot-user-guide.md
+++ b/docs/brainrot-user-guide.md
@@ -647,6 +647,15 @@ rizz yapidx(rant hay, rant needle);   /* byte index of needle, or -1         */
   same reason.
 - Bytes compare as **unsigned**, matching C's `strcmp` and Go's
   `strings.Compare`, so non-ASCII bytes sort after every ASCII one.
+- ⚠️ **A `yap[N]` buffer always reports length `N`.** A `rant` carries a real
+  length, but a character buffer — what `slorp` fills — reports its declared
+  capacity, so with `yap buf[32]; slorp(buf);` and the user typing `hi`:
+  `yaplen(buf)` is `32`, `yapcmp(buf, "hi")` is `1` ("greater", because it is
+  32 long), and `yapcat(buf, "!")` builds a real 33-byte string that merely
+  *prints* as `hi` because output stops at the first NUL. Only `yapidx` is
+  unaffected. Measure and compare a `rant`, not a buffer — and there is no
+  conversion: `rant r = buf;` is a type error. This comes from how character
+  arrays are marshalled to builtins, not from these functions.
 - **Strings are immutable.** `yapcat` modifies neither argument and returns a
   new string that is independent of both, so an earlier result stays valid
   after later calls.
--- a/docs/the-brainrot-programming-language.md
+++ b/docs/the-brainrot-programming-language.md
@@ -1310,6 +1310,39 @@ Bytes compare as **unsigned**, which is what C's `strcmp` and Go's
 `strings.Compare` do. Every non-ASCII byte therefore sorts *after* every ASCII
 one.
 
+> ### ⚠️ A `yap[N]` buffer always has length `N`
+>
+> This is the neighbouring gotcha, and it bites harder than the byte/character
+> one. A `rant` carries a real length. A `yap[N]` character buffer — the thing
+> [`slorp`](#86-slorp) fills, and the usual way a program reads input —
+> reports its **declared capacity**, not the length of the text in it.
+>
+> ```c
+> yap buf[32];
+> slorp(buf);                        🚽 user types "hi"
+>
+> yaplen(buf)          🚽 32, not 2
+> yapcmp(buf, "hi")    🚽 1  -- "buf is greater", because it is 32 long
+> yapidx(buf, "i")     🚽 1  -- correct
+> yapcat(buf, "!")     🚽 a real 33-byte string; prints as "hi" because
+>                      🚽 yapping stops at the first NUL, so the "!" is
+>                      🚽 sitting 31 bytes past it
+> ```
+>
+> So measure, compare and join a **`rant`**, not a raw buffer — and note there
+> is currently no way to convert one into the other: `rant r = buf;` is
+> rejected outright with *"Type mismatch in initialization of 'r': expected
+> string, got char"*. A `yap[N]` stays a `yap[N]`.
+>
+> `yapidx` is the one that still behaves, because a bounded scan finds the
+> needle at the same offset whether or not the trailing bytes are counted. If
+> you only need "where is it" or "does it contain it", it works on a buffer.
+>
+> This is a property of how character arrays are passed to builtins, not of
+> these four functions; they are simply the first to read the stored length
+> rather than treating the bytes as a C string. Pinned by
+> `test_cases/string_stdlib_char_buffer.brainrot`, and expected to change.
+
 **Strings are immutable; `yapcat` returns a new one.** Neither argument is
 modified, and the result is independent of both -- storing it and then calling
 `yapcat` again leaves the first result untouched.
--- a/docs/the-brainrot-programming-language.md
+++ b/docs/the-brainrot-programming-language.md
@@ -33,6 +33,7 @@ A Meme-Fueled Journey into Compiler Design, Internet Slang, and Skibidi Toilets
    - 8.6. `slorp`
    - 8.7. `bet`
    - 8.8. `gamba`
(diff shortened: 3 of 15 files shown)


stdrot/yaplen.c
```diff
@@ -0,0 +1,55 @@
...
+#include <stdio.h>
+#include <stdlib.h>
+
+StdrotValue stdrot_yaplen(StdrotValue *args, int argc)
+{
+    (void)argc;
+    /* A String with .data == NULL is representable by the struct even
+       though no path here was found to produce one, so it is treated as
+       empty rather than trusting .len alongside a null buffer. */
+    size_t len = args[0].val.str.data ? args[0].val.str.len : 0;
```
leo-aa88, line 32: **MAJOR** — not a bug in this file, but this is where it's easiest to show.

`.len` for a `yap[N]` buffer is the **declared array size**, not the length of the string in it. `ast_expr_to_stdrot_value()` (stdrot.c:1105) sets `out->val.str.len = (size_t)var->array_length` for the `VAR_CHAR`/`is_array` case. So:

```c
skibidi main {
    yap buf[32];
    slorp(buf);                        🚽 user types "hi"
    yapping("%d", yaplen(buf));        🚽 32
    yapping("[%s]", yapcat(buf, "!")); 🚽 [hi]   -- the "!" is 31 bytes past a NUL
    yapping("%d", yapcmp(buf, "hi"));  🚽 1      -- "buf is greater", it's 32 long
    yapping("%d", yapidx(buf, "i"));   🚽 1      -- correct
}
```

Three of the four give an answer driven by the buffer's *capacity*. `yapcat` is the worst of them: the result is a genuinely 33-byte string with NULs in the middle, so it prints as `[hi]` and looks fine while being nothing of the sort — and it's a real 33-byte allocation, so nothing downstream flags it either.

This matters more than a normal edge case because `yap buf[N]; slorp(buf);` is *the* documented way to read input — `test_brainrot.py` has a dedicated regression test for that idiom — so "measure / compare / join what the user just typed" is the first thing anyone will point this library at, and it's the one shape that misbehaves.

The `.len` behaviour is pre-existing and I'm not asking you to change it here; changing it would touch every native that takes a string. What's new is that these are the first builtins to *read* `.len` rather than treat the data as a C string, which is precisely the distinction `yapidx.c`'s own comment draws when it refuses `strstr()`. The library knows a `rant` is length-prefixed; the docs and fixtures don't yet confront what that means for the buffers users have.

Minimum I'd want before merge:

- **A fixture.** None of the three covers a `yap[N]` argument. Whatever the intended behaviour is, pin it — otherwise the answer changes silently the first time someone touches `.len` handling.
- **A doc line.** The new sections are good on the byte-vs-character distinction (`yaplen("é")` is 2) — this is the neighbouring gotcha and belongs right next to it: a `rant` literal or variable has the length you expect, a `yap[N]` buffer always has length N.

If you'd rather fix it than document it, the honest fix is a NUL-scan for the `is_array` case in `ast_expr_to_stdrot_value()` — but that changes the ABI's meaning of `.len` for every existing native, so it's its own PR with its own blast radius, not something to slip in here.

leo-aa88 (commented):
# COMMENT

Careful work, and the part I most expected to be wrong isn't. One finding,
about the one string source users actually have.

## The `yapcat` scratch design holds up

That file comment makes a strong claim — a file-static buffer, reused across
calls, safe because the host materializes the result before anything else
runs. I tried to break it:

```
two live results at once      x = yapcat("aa","bb"); y = yapcat("cc","dd")  → aabb ccdd
nested both sides             yapcat(yapcat("1","2"), yapcat("3","4"))      → 1234
returned through a function   join3("p","q","r")                            → pqr
200-iteration loop            no accumulation
two results into one call     yapcmp(yapcat("a","b"), yapcat("a","b"))      → 0
```

and the earlier results stay intact throughout. Valgrind agrees — all three
fixtures plus that stress program are clean under the repo's own criteria
(`--leak-check=full --error-exitcode=100`), `definitely lost: 0` on every one.
Worth doing explicitly here rather than trusting the ASan suite: `libstdrot.so`
is built without sanitizers, so these four files are exactly the code the
normal test run cannot see.

Two other things I checked rather than assumed:

- The overflow guard's `total == (size_t)-1` arm is genuinely necessary, and
  the comment's reasoning about why `alen > SIZE_MAX - blen - 1` is wrong for
  `blen == SIZE_MAX` is correct.
- The destructor-over-`atexit` argument is right and non-obvious: an `atexit`
  handler registered from inside the `.so` can be invoked after `dlclose()`
  has unmapped its own code. Getting that backwards would be a crash at exit,
  not a leak.

`yapidx` refusing `strstr()` because a `rant` is length-prefixed and not
NUL-terminated is also correct — and it is the same property that produces my
finding.

## The finding

Inline. A `yap[N]` buffer — the thing `slorp` fills, and the only way a
program gets a string from the outside — presents to these builtins with
`.len == N`, not the length of what is in it.

# VERDICT

The implementation is more careful than most stdlib v1s: bounds, overflow,
NULL-data, empty-needle semantics, and an ownership model that survives
adversarial nesting. What it hasn't confronted is that the language's
length-prefixed `rant` and its `yap[N]` input buffers disagree about what
"length" means, and this is the first library where that disagreement produces
wrong answers instead of harmless slack.

leo-aa88 (commented):
# COMMENT

Build fix is correct and the diagnosis behind it is worth more than the
three-line change. The `yap[N]` finding is untouched.

## The fix, verified

Reproduced the pre-fix failure at the exact line:

```
yapcat.c:30:69: error: '/*' within block comment [-Werror,-Wcomment]
 * The arena is not an option either: libstdrot.so links only stdrot/*.c
                                                                    ^
```

and all four sources now compile clean under the flag set you cite. 485 pass,
`format-check` clean.

**The diagnosis checks out and is the interesting part.** I verified the claim
rather than taking it:

```
Makefile:11  CFLAGS    := -Wall -Wextra -Wpedantic -Werror -O2 -Wuninitialized -fsanitize=...
Makefile:13  SO_CFLAGS := -fPIC -shared
```

So *every* file in `stdrot/` is compiled with zero diagnostics in a native
build — not just these four, and not just `-Wcomment`. Combined with
`libstdrot.so` also being built without sanitizers, the standard library is the
one part of this repo that neither `-Werror` nor ASan/UBSan looks at natively,
and the wasm job is the only thing that does. That is worth its own issue: a
`-Wall -Wextra -Werror` pass over `SO_CFLAGS` would be cheap and would have
caught this before CI, and it's the same blind spot that makes valgrind the
only way to find memory bugs in these files.

## Still open

From my previous review, unchanged at this head:

```
yap buf[32]; slorp(buf);   (user types "hi")
→ yaplen=32   cat=[hi]   cmp=1
```

Not a criticism of this commit — it's a build fix and doesn't claim otherwise.
Flagging only so it doesn't get lost behind a green CI run.

# VERDICT

Right fix, and the write-up correctly identifies why local `make` couldn't see
it. The `yap[N]` question still needs a fixture and a doc line before this
merges.

leo-aa88 (commented):
# APPROVE

The `yap[N]` gap is closed the way I'd hoped — pinned *and* documented, and the
fixture is sharper than what I asked for.

**The fixture measures rather than prints, which is the whole point:**

```
slorp len 32      capacity, not content
slorp cmp 1       "greater", because it is 32 long
slorp idx 1       correct -- yapidx is the one that isn't affected
cat       [hi]    prints as "hi"...
cat len   33      ...and is really 33 bytes
rant len  2       the contrast, in the same fixture
rant cmp  0
```

`cat len 33` is the line that matters. Printing alone shows `[hi]` and hides
the problem completely — measuring the concatenation is what makes the
swallowed `!` visible, and it's exactly the assertion a fixture built around
`yapping` output would have missed. Putting the `rant` contrast in the same
file means the difference between the two string sources is the thing under
test, not a footnote.

Wiring it into `tests/stdin_fixtures.json` **and** `run_valgrind_tests.sh` is
the right completeness — a `slorp` fixture that the valgrind sweep skips would
be a hole in the one tool that can see `libstdrot.so`, since it's built without
sanitizers. Checked it directly: valgrind clean, `definitely lost: 0`,
`ERROR SUMMARY: 0 errors`.

**Both doc sites are accurate**, including the two details that make it
actionable rather than just a warning: that `yapidx` is unaffected, and that
there is no conversion — `rant r = buf;` is a type error, so "measure a `rant`,
not a buffer" is advice a reader can actually follow. Attributing it to how
character arrays are marshalled rather than to these four functions is also
correct and keeps the blame where a future fix would go.

486 collected, `format-check` clean.

# VERDICT

Three rounds on this one: a careful implementation, a build fix whose diagnosis
turned up that `stdrot/` compiles with no warning flags at all, and now the
`yap[N]` boundary pinned in a fixture that can actually fail. The remaining
`SO_CFLAGS` gap is worth its own issue, not a blocker here.

Ship it.

PR #329, opened 2026-08-30
feat: file I/O behind SAUCE * handles, and STDROT_HANDLE both ways (#213)

### LINKED ISSUE #213: [Roadmap] Phase 10 — File I/O
> **Roadmap tracking:** #214
> **Depends on:** #204. Can ship a file-only handle ahead of #205 STDROT_HANDLE.
> **Blocks / unblocks:** nothing

---

## Summary

`stdio.h`, rebranded. Like `sussybaka`, this is a **library, not keywords** — every operation here is "a call with arguments", so it belongs in `stdrot` alongside `yapping`, `slorp`, and `chill`, not in `lang.l`.

Unlike `sussybaka` it is **not** gated behind `#cooked` — file I/O is common enough to stay globally available, the same way `slorp` is today.

| | |
| --- | --- |
| **Roadmap** | Phase 10 — File I/O |
| **Milestone** | M12 — Files |
| **Priority** | P2 |
| **Status** | not started |
| **Depends on** | Phase 1 (`SAUCE *f = crackopen(...)` is an initializer). Can ship a file-only handle ahead of Phase 2's `STDROT_HANDLE` and adopt the general one later. |
| **Blocks** | nothing |
| **Baseline** | `0cfa958` |

Independent of M2–M6 — can proceed in parallel once Phase 1 lands.

## Vocabulary (library functions, not keywords)

`FILE *` → **`SAUCE *`**

| C | Brainrot | Why |
| - | -------- | --- |
| `fopen` | `crackopen` | you crack open the file |
| `fclose` | `peaceout` | perfect opposite lifecycle |
| `fread` | `doomscroll` | consuming data endlessly |
| `fwrite` | `shitpost` | putting content into the file |
| `fgets` | `skim` | silently reads a line |
| `fputs` | `yapto` | yapping into a file |
| `fseek` | `zoink` | literally moving through the file |
| `ftell` | `whereami` | cursor position |
| `rewind` | `throwback` | self-explanatory |
| `feof` | `itsjoever` | **this one is mandatory** |
| `ferror` | `bricked` | file operation got bricked |
| `fflush` | `bustcache` | flush buffered output |

`yapto` is deliberately distinct from raw `shitpost`: `shitpost` is `fwrite`-shaped (size/count, binary-safe), `yapto` is `fprintf`-shaped (a format string). That mirrors `yapping` (stdout, formatted, newline) vs `yappin` (stdout, formatted, no newline).

None of these twelve names collide with `lockin`/`logoff` (Phase 9) or `lurk` (Phase 8). An earlier draft of this table used those three names, which is why the family was renamed wholesale.

**README keyword table is not touched** — none of these are grammar keywords.

## Example usage (target)

```c
skibidi main {
    SAUCE *f = crackopen("classified_lore.txt", "r");
    edgy (!f) {
        yapping("file got negative aura");
        ragequit(1);
    }

    goon (!itsjoever(f)) {
        rant line = skim(f);
        yapping("%s", line);
    }

    peaceout(f);
    bussin 0;
}
```

```c
SAUCE *manifesto = crackopen("schizo.txt", "w");
yapto(manifesto, "aura = %d", aura);
peaceout(manifesto);
```

## Work

- [ ] A `SAUCE *` handle type wrapping `FILE *`, following the handle pattern Phase 2 formalizes (`STDROT_HANDLE`). Ship a narrower, file-only handle if Phase 2 is not in yet; adopt the general one once it lands.
- [ ] `stdrot/file.c` implementing the twelve functions via `STDROT_EXPORT` (or `STDROT_EXPORT_SIG` if Phase 2 has landed)
- [ ] Binary-safe `doomscroll` / `shitpost` plus formatted `yapto` / line-oriented `skim`
- [ ] Falsy / negative handle from `crackopen` on a missing file, so `edgy (!f)` works
- [ ] No leaked `FILE *` on any exit path — including `ragequit` / failed `bussin` if that's reachable. `make valgrind` is the relevant gate.
- [ ] `docs/` reference page for the file I/O vocabulary
- [ ] Do **not** add these names to `lang.l`

## Files likely to change

- `stdrot/file.c` (new), `stdrot/stdrot_api.h` if a handle type is added
- `ast.c` / analyzer only if `SAUCE *` needs a new type spelling — prefer a handle that already fits the type system (or a typedef-like alias if `lit` / Phase 2 handles cover it)
- `test_cases/file_io.brainrot` + a missing-file case
- `tests/expected_results.json`
- `docs/` (not the README keyword table)

## Definition of done

- [ ] `test_cases/file_io.brainrot` covers open / read / write / close and the `itsjoever` loop idiom
- [ ] Missing-file case: `crackopen` returning a negative/falsy handle, exercised through `edgy (!f)`
- [ ] No leaked `FILE *` on any exit path (`make valgrind`)
- [ ] `docs/` reference page for the vocabulary
- [ ] README keyword table **not** touched

## References

- [docs/ROADMAP.md — Phase 10](https://github.com/Brainrotlang/brainrot/blob/main/docs/ROADMAP.md#phase-10--file-io)
- Appendix B Q6 — ownership of native resources (files are the simplest handle case)

leo-aa88:
## Description

Phase 10 — twelve file operations, plus the ABI capability they needed.

```c
skibidi main {
    SAUCE *f = crackopen("classified_lore.txt", "r");
    edgy (!f) {
        yapping("file got negative aura");
        ragequit(1);
    }

    goon (!itsjoever(f)) {
        rant line = skim(f);
        yapping("%s", line);
    }

    peaceout(f);
    bussin 0;
}
```

`crackopen` `peaceout` `doomscroll` `shitpost` `skim` `yapto` `zoink` `whereami` `throwback` `itsjoever` `bricked` `bustcache` — library functions, not keywords, and not gated behind `#cooked`. **The README keyword table is untouched**; the only thing this adds to the grammar is the *type* name `SAUCE`.

## STDROT_HANDLE, and the question it was blocked on

Handle returns were rejected outright, because a handle *"needs a resource-identity/ownership model (type_name-based type checking, who frees it, GC vs manual release) that Phase 2 deliberately hasn't designed yet"* — the roadmap's **Appendix B Q6**. Q6's own hint turns out to be the answer, and it's now written down on `STDROT_HANDLE` itself:

1. **Ownership stays in C.** Brainrot holds a token, never a freeable pointer.
2. **Release is manual** (`peaceout`) — no destructors, no GC.
3. **The owning library keeps a registry of live handles** and validates against it.
4. **Anything still live at unload is released by the library.**

**(3) is the load-bearing part**, and the reason a handle beats the raw pointer underneath it. A program can produce an address that was never a file — a stale token kept past `peaceout`, arithmetic, anything — and the type system can't tell, because it sees "opaque pointer". Passing one to `fclose()` is UB. Resolving through the registry turns use-after-release, double-release, and operating on a failed `crackopen`'s null handle into diagnostics instead.

Retagging a pointer as a handle *without* a registry behind it would be security theatre, and the code says so where it does the retagging: the coercion **asserts** the declared kind, it does not **verify** it. Kinds are kept apart statically by `SAUCE *` being its own type spelling, and at runtime by the library refusing any address that isn't one of its own.

Handles work in **both directions**. `STDROT_CSTRING` and `STDROT_STRUCT` returns stay rejected, and the reason is now sharper than "ownership is unsolved": a handle is a *token*, not memory the caller must free, so it never had their question to answer.

## `itsjoever` peeks, deliberately

The one departure from the C original, and it's in the issue's headline example. `feof()` reports that a read has **already** failed, so `while (!feof(f))` runs one iteration too many — the most common file-reading bug in C. A literal `feof` would make that loop print a spurious blank line at the end of every file. `itsjoever` peeks one byte and puts it back, answering the question its name asks. `bricked` stays a plain `ferror()`, because there C's semantics are the right ones.

## `make valgrind` was not actually the gate the DoD assumed

The issue requires *"no leaked `FILE *` on any exit path (`make valgrind`)"*. **It wasn't.** A `FILE *` that is never `fclose`d does **not** show as `definitely lost` — glibc keeps every open stream on its own list, so the allocation stays reachable to the very end. Verified by mutation: removing the release path left `definitely lost: 0 bytes` and a **passing** sweep.

So `run_valgrind_tests.sh` now also runs `--track-fds=yes` and fails on any descriptor open at exit beyond the standard three:

| build | valgrind reports |
| --- | --- |
| release path removed | `FILE DESCRIPTORS: 5 open (3 std) at exit.` → sweep fails |
| as shipped | `FILE DESCRIPTORS: 3 open (3 std) at exit.` → sweep passes |

The whole corpus is clean under the new gate — 438 fixtures, exit 0, no pre-existing descriptor leaks anywhere.

While adding it I had to switch the stderr capture off a `tee` process substitution: that runs asynchronously, so the log could still be being written when the check read it, which made the first version report a phantom leak on the first fixture.

## Two other things this turned up

- **The deprecated write-back convention would have corrupted a handle.** `peaceout(f)` returns `fclose`'s status, and writing that back into `f` turns a live SAUCE into `0`. A `STDROT_HANDLE` first parameter is now exempt — for a *stronger* reason than the `yap[N]` exemption beside it, where write-back is merely redundant rather than destructive.
- **`yapping`'s formatter now takes its stream** instead of assuming stdout, so `yapto` shares it rather than growing a second copy that drifts. Declared in a new `stdrot/stdrot_format.h`, deliberately **not** in `stdrot_api.h` — that's the external ABI contract, not a place for libstdrot's internals.

## Related Issue

Fixes #213. Also resolves **Appendix B Q6** for the handle case.

## Type of Change

- [x] Documentation update
- [ ] Performance improvement
- [ ] Refactor

Additive: new grammar (one type name), a new ABI capability that was previously rejected, and a new stdrot file. 496 pre-existing tests are untouched.

## Checklist

### Tests

| fixture | pins |
| --- | --- |
| `file_io` | all twelve operations, the read-loop idiom printing **no** trailing blank, binary-safe `doomscroll` length, and a missing file being falsy rather than fatal |
| `file_io_use_after_close_fail` | the registry catching a stale handle — the property that makes handles worth having |
| `file_io_ragequit_no_leak` | two unclosed handles, in front of the valgrind sweep |

`file_io_ragequit_no_leak`'s own comment says plainly that **its stdout proves nothing** — `make test` would pass identically on a build that leaked every file. The sweep is what can see the difference, and the fixture's value is existing *inside* it.

`file_io` is self-contained: it writes the file it reads, so there's no checked-in data and re-running gives the same answer.

### Gate results

| gate | result |
| --- | --- |
| `make test` | 499 passed (was 496) |
| valgrind sweep | 438 fixtures, exit 0 — **now including the fd gate** |
| `make tidy` | pass |
| `bison -Wcounterexamples` | no conflicts |

One thing I could **not** verify locally: the wasm fixture run. There's no `emcc` on this box, and the file fixtures write to `/tmp`. Emscripten's MEMFS creates `/tmp` by default so this should be fine, but the "Release dry-run wasm" job is the first real check of it.

As before, the sweep ran against a build with `-fsanitize` dropped — ASan's shadow mapping collides with valgrind on this WSL2 kernel. The sanitized build was restored before `make test`.

### DIFF
--- a/stdrot/file.c
+++ b/stdrot/file.c
@@ -1,0 +1,439 @@
+/* stdrot/file.c -- stdio.h, rebranded (#213).
+ *
+ * Twelve file operations behind a `SAUCE *` handle. Not gated behind
+ * `#cooked`: file I/O is common enough to stay globally available, the same
+ * way slorp is.
+ *
+ *   crackopen  fopen      peaceout   fclose     doomscroll fread
+ *   shitpost   fwrite     skim       fgets      yapto      fprintf
+ *   zoink      fseek      whereami   ftell      throwback  rewind
+ *   itsjoever  feof       bricked    ferror     bustcache  fflush
+ *
+ * ── The live-handle registry, and why it is the whole point ────────────
+ * This file implements the ownership model documented on STDROT_HANDLE
+ * (stdrot_api.h), which is this project's answer to roadmap Appendix B Q6
+ * ("Textures, sockets, and map entries all outlive statements... Handles
+ * sidestep this by keeping ownership in C -- is that the general answer?").
+ * Files are the simplest case to prove it on.
+ *
+ * The registry is not bookkeeping. A Brainrot `SAUCE *` is an address, and
+ * a Brainrot program can produce an address that was never a file: a stale
+ * token kept past peaceout(), a pointer from somewhere else entirely, the
+ * result of arithmetic. Handing any of those to fclose()/fread() is
+ * undefined behaviour that no amount of static typing prevents, because
+ * the type system sees only "opaque pointer".
+ *
+ * So every entry point resolves its handle through resolve_handle(), which
+ * accepts ONLY an address currently in this table. Use-after-release and
+ * double-release become diagnostics instead of a double free(); a
+ * fabricated token becomes an error instead of a wild fclose(). That is
+ * what makes a handle genuinely safer than the raw pointer underneath it,
+ * and it is why the ABI comment says retagging a pointer as a handle
+ * without a registry behind it would be security theatre.
+ *
+ * ── No leaked FILE * on any exit path ─────────────────────────────────
+ * A destructor closes everything still open when libstdrot.so is unloaded.
+ * That covers the paths a program cannot clean up after: ragequit(), a
+ * native that exit()s, or simply forgetting peaceout(). It is the same
+ * mechanism yapcat.c uses for its scratch buffer, and it must be a
+ * DESTRUCTOR rather than atexit() for the same reason: stdrot_unload()
+ * dlclose()s this object, and an atexit handler registered from here could
+ * be invoked after its own code was unmapped.
+ */
+#include "stdrot_api.h"
+#include "stdrot_format.h"
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+
+/* Grows by doubling; there is no fixed ceiling on open files beyond what
+   the OS itself enforces. */
+static FILE **live_files = NULL;
+static size_t live_count = 0;
+static size_t live_capacity = 0;
+
+/* doomscroll's read scratch. File-scope rather than function-static so the
+   destructor below can release it: libstdrot.so is dlclose()d by
+   stdrot_unload(), which unmaps the static itself, so a buffer merely left
+   "still reachable" is genuinely unreachable by the time LeakSanitizer
+   scans and is reported as a real leak. Same reasoning as yapcat.c's. */
+static char *read_buf = NULL;
+static size_t read_buf_size = 0;
+
+static void registry_add(FILE *fp)
+{
+    if (live_count == live_capacity)
+    {
+        size_t next = live_capacity ? live_capacity * 2 : 8;
+        /* sizeof a FILE* is exactly what is wanted here -- this array holds
+           pointers, not FILE objects, whose size is not even knowable.
+           bugprone-sizeof-expression flags every sizeof of a pointer-to-
+           aggregate because the usual mistake is meaning the aggregate;
+           that is not the case here. */
+        // NOLINTNEXTLINE(bugprone-sizeof-expression)
+        FILE **grown = realloc(live_files, next * sizeof(*live_files));
+        if (!grown)
+        {
+            fprintf(stderr,
+                    "Error: crackopen: out of memory tracking open "
+                    "files at line %d\n",
+                    g_exec_context.line_number);
+            fclose(fp);
+            exit(1);
+        }
+        live_files = grown;
+        live_capacity = next;
+    }
+    live_files[live_count++] = fp;
+}
+
+static bool registry_remove(FILE *fp)
+{
+    for (size_t i = 0; i < live_count; i++)
+    {
+        if (live_files[i] == fp)
+        {
+            live_files[i] = live_files[--live_count];
+            return true;
+        }
+    }
+    return false;
+}
+
+static bool registry_contains(const FILE *fp)
+{
+    for (size_t i = 0; i < live_count; i++)
+    {
+        if (live_files[i] == fp)
+            return true;
+    }
+    return false;
+}
+
+/* Runs when libstdrot.so is unloaded, while this object is still mapped.
+   See the file comment for why a destructor and not atexit(). */
+__attribute__((destructor)) static void file_release_all(void)
+{
+    for (size_t i = 0; i < live_count; i++)
+        fclose(live_files[i]);
+    free(live_files);
+    live_files = NULL;
+    live_count = 0;
+    live_capacity = 0;
+    free(read_buf);
+    read_buf = NULL;
+    read_buf_size = 0;
+}
+
+/* The one gate every operation below passes through. A handle that is not
+   currently live is an error, never a pointer to act on -- see the file
+   comment for why this is the load-bearing part of the design. */
+static FILE *resolve_handle(const char *who, const StdrotValue *v)
+{
+    FILE *fp = (FILE *)v->val.handle.handle;
+    if (!fp || !registry_contains(fp))
+    {
+        fprintf(stderr,
+                "Error: %s: not an open SAUCE -- it was never opened, or "
+                "was already closed with peaceout, at line %d\n",
+                who, g_exec_context.line_number);
+        exit(1);
+    }
+    return fp;
+}
+
+/* A `rant` is length-prefixed and NOT NUL-terminated, so paths and modes
+   have to be copied into a terminated buffer before reaching libc. */
+static bool to_cstr(const String s, char *out, size_t out_size)
+{
+    if (!s.data || s.len >= out_size)
+        return false;
+    memcpy(out, s.data, s.len);
+    out[s.len] = '\0';
+    return true;
+}
+
+/* crackopen(rant path, rant mode) -> SAUCE*
+   Returns a null handle when the file cannot be opened, so `edgy (!f)`
+   is the idiomatic check -- a null pointer is falsy in Brainrot. This is
+   deliberately not an error: a missing file is an ordinary, expected
+   outcome a program should be able to handle, unlike being handed a
+   fabricated handle, which is a bug. */
+static StdrotValue stdrot_crackopen(StdrotValue *args, int argc)
+{
+    (void)argc;
+    char path[4096];
+    char mode[16];
+    StdrotValue out = {STDROT_HANDLE, {0}};
+    out.val.handle.type_name = "SAUCE";
+    out.val.handle.handle = NULL;
+
+    if (!to_cstr(args[0].val.str, path, sizeof(path)) ||
+        !to_cstr(args[1].val.str, mode, sizeof(mode)))
+    {
+        return out; /* unusable path/mode reads as "could not open" */
+    }
+
+    FILE *fp = fopen(path, mode);
+    if (!fp)
+        return out;
+
+    registry_add(fp);
+    out.val.handle.handle = fp;
+    return out;
+}
+
+/* peaceout(SAUCE *f) -> rizz -- 0 on success, matching fclose. */
+static StdrotValue stdrot_peaceout(StdrotValue *args, int argc)
+{
+    (void)argc;
+    FILE *fp = resolve_handle("peaceout", &args[0]);
+    registry_remove(fp);
+    return (StdrotValue){STDROT_INT, {.i = fclose(fp)}};
+}
+
+/* doomscroll(SAUCE *f, rizz n) -> rant -- fread of up to n bytes.
+   Binary-safe: the result's length is the byte count actually read, so
+   embedded NULs survive, which is exactly what a rant's length prefix is
+   for. Returns a short (or empty) rant at end of file. */
+static StdrotValue stdrot_doomscroll(StdrotValue *args, int argc)
+{
+    (void)argc;
+    FILE *fp = resolve_handle("doomscroll", &args[0]);
+    int want = args[1].val.i;
+    if (want < 0)
+    {
+        fprintf(stderr,
+                "Error: doomscroll: byte count must not be negative "
+                "(got %d) at line %d\n",
+                want, g_exec_context.line_number);
+        exit(1);
+    }
+
+    size_t need = (size_t)want + 1;
+    if (need > read_buf_size)
+    {
+        char *grown = realloc(read_buf, need);
+        if (!grown)
+        {
+            fprintf(stderr,
+                    "Error: doomscroll: out of memory reading %d "
+                    "bytes at line %d\n",
+                    want, g_exec_context.line_number);
+            exit(1);
+        }
+        read_buf = grown;
+        read_buf_size = need;
+    }
+
+    size_t got = fread(read_buf, 1, (size_t)want, fp);
+    read_buf[got] = '\0';
+    /* The host deep-copies a returned STDROT_STRING before anything else
+       runs, so handing back this reused scratch is safe -- the same
+       contract yapcat.c relies on and documents at length. */
+    return (StdrotValue){STDROT_STRING,
+                         {.str = {.data = read_buf, .len = got}}};
+}
+
+/* shitpost(SAUCE *f, rant data) -> rizz -- fwrite, binary-safe.
+   Writes data.len bytes, NOT up to a terminator, so a rant containing an
+   embedded NUL round-trips through doomscroll unchanged. Returns the
+   number of bytes written. */
+static StdrotValue stdrot_shitpost(StdrotValue *args, int argc)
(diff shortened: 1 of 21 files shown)


stdrot/file.c
```diff
@@ -0,0 +1,439 @@
...
+        if (live_files[i] == fp)
+        {
+            live_files[i] = live_files[--live_count];
+            return true;
+        }
+    }
+    return false;
+}
+
+static bool registry_contains(const FILE *fp)
```
leo-aa88: **BLOCKING.** This answers "is some live file at this address?", but every caller needs "is *this* the file the program opened?". Those differ the moment `malloc` hands the same address out twice, which it does almost immediately.

```c
SAUCE *a = crackopen("/tmp/fa.txt", "r");   🚽 AAAA
peaceout(a);
SAUCE *b = crackopen("/tmp/fb.txt", "r");   🚽 BBBB
skim(a)   🚽 -> "BBBB", no diagnostic
```

and the close path corrupts across identities:

```
peaceout(a) → 0     closes fa
b = crackopen(fb)   same address
peaceout(a) → 0     silently closes fb
peaceout(b) → Error: not an open SAUCE
```

Measured, not assumed — `edgy (b == a)` across fifty close/reopen cycles:

```
release build:  50 of 50 reused
ASan build:     0 of 50
```

ASan's quarantine is why the suite is green, and valgrind's allocator hides it the same way. `file_io_use_after_close_fail` uses the stale handle with no intervening `crackopen`, which is the only shape that behaves identically under both.

**The fix doesn't touch the ABI.** `StdrotValue.val.handle.handle` is a `void *`, so it can carry a token instead of the pointer:

- keep the registry as a `FILE **` slot array, and hand back `(void *)(uintptr_t)((generation << 16) | slot)` — or simply a monotonically increasing `uint64_t` id mapped to its `FILE *`
- `resolve_handle()` looks the token up instead of scanning for pointer equality; a stale token either falls outside the table or fails the generation check
- reuse of a *slot* is then fine, because the generation makes the token unique

A monotonic id is the simpler of the two and never collides at all; the slot+generation version keeps lookup O(1) without a map. Either way the guarantee becomes structural rather than allocator-dependent.

One thing worth keeping from the current code: rejecting `!fp` before the lookup. A failed `crackopen` should stay distinguishable from a fabricated token.

leo-aa88 (commented):
# REQUEST CHANGES

The registry is the right idea and the rest of this is strong work — but it
validates **membership, not identity**, and the allocator recycles addresses.
The guarantee the whole design rests on, and that the ABI header now states as
a permanent contract, does not hold on a release build.

## The blocker

```c
SAUCE *a = crackopen("/tmp/fa.txt", "r");   🚽 contains AAAA
peaceout(a);                                 🚽 a is now stale
SAUCE *b = crackopen("/tmp/fb.txt", "r");   🚽 contains BBBB
skim(a)                                      🚽 -> "BBBB"
```

No diagnostic. The stale handle silently read a different file, because
`fclose()` freed the `FILE` and the next `fopen()` got the same address back,
so `registry_contains()` says yes.

Closing is worse than reading:

```
peaceout(a)   → 0        closes fa
b = crackopen(fb)        same address
peaceout(a)   → 0        silently closes fb -- someone else's file
peaceout(b)   → Error: not an open SAUCE
```

A stale token closed a live file, and then the rightful owner's `peaceout`
was rejected.

**How reliably?** 50 of 50 on a release build. I measured it rather than
assuming — `edgy (b == a)` after close-and-reopen, fifty iterations:

```
release build (as shipped):  address reuse hits: 50 of 50
ASan build (make):           address reuse hits: 0 of 50
```

That gap is why nothing caught it. ASan's quarantine delays reuse, so the
whole suite — which runs the ASan build — sees the good case. Valgrind's
allocator does the same; my double-release repro errors correctly under
valgrind and misbehaves outside it. And
`file_io_use_after_close_fail.brainrot` closes and then uses **with no
intervening `crackopen`**, which is the one shape that works either way.

Details and the fix inline. It doesn't touch the ABI.

## The rest is good, and two parts are better than good

**The `--track-fds=yes` catch is excellent and I verified the reasoning.** A
leaked `FILE *` really doesn't show as `definitely lost` — glibc keeps open
streams on its own list, so the allocation stays reachable. Discovering that
the DoD's stated gate wasn't actually a gate, and then building the real one,
is exactly the right response to "the requirement says `make valgrind`". The
`tee`-process-substitution race you hit and fixed is a genuine trap too.

**The `itsjoever` peek-instead-of-`feof` decision is right** and the
justification is the correct one: `while (!feof(f))` running one iteration
long is the single most common C file-reading bug, and a literal port would
have shipped it into every Brainrot program. Keeping `bricked` as plain
`ferror()` shows the line was drawn on reasoning rather than taste.

Also verified: the deprecated write-back exemption for a `STDROT_HANDLE` first
parameter is necessary — `peaceout(f)` returning `fclose`'s status into `f`
would zero a live handle — and 499 collected, suite green.

# VERDICT

The architecture is right: ownership in C, manual release, a registry, and
release-at-unload. What's wrong is the key. Validating an address against a
set of live addresses cannot distinguish "this handle" from "some handle that
happens to live where this one used to", and `malloc` makes that collision the
normal case rather than the rare one. Give out a token that is never reused
and the design delivers what it claims.

stdrot/stdrot_api.h
```diff
@@ -175,7 +175,57 @@ typedef enum
...
+                           validates every handle it is given against it.
+                           This is the part that matters, and it is why a
+                           handle is genuinely safer than the raw STDROT_PTR
+                           it superficially resembles: a Brainrot program can
+                           fabricate an address (integer arithmetic, a stale
+                           token kept past release) and hand it back, and a
+                           raw pointer would be dereferenced or free()d. A
+                           registered handle is looked up first and rejected
+                           if it is not live -- so use-after-release and
+                           double-release are diagnosed, not undefined.
```
leo-aa88: This sentence is the contract the whole handle design is sold on, and it's in the file that outlives every implementation behind it:

> so use-after-release and double-release are diagnosed, not undefined.

As implemented that holds only while the allocator hasn't recycled the address. On a release build it recycles 50 times out of 50, so both cases go undiagnosed — see my comment on `stdrot/file.c`. A use-after-release silently reads another file; a double-release silently closes one.

I'd rather this sentence stay and the implementation catch up to it — it's the right guarantee to promise, and a token-based registry delivers it structurally. But if any part of it has to be weakened, weaken it here rather than leaving the ABI header asserting something a caller can't rely on: this is the document a future binding author reads when deciding whether `STDROT_HANDLE` is safe enough to build a socket or a database cursor on.

Worth adding either way, once the token change lands: a line saying the handle value is an **opaque token, not a pointer**, and that a binding must not dereference it or compare it against addresses of its own. Right now `SAUCE *` reads like a pointer and behaves like one, which is what made assuming address identity feel natural in the first place.

leo-aa88 (commented):
# REQUEST CHANGES (blocker unchanged)

The new commit is a CI fix for the fd gate, not the handle-identity blocker.
Both verified.

## The fd-gate fix is correct, and it kept its teeth

The failure mode is real — a CI runner hands its child inherited descriptors
and the summary line lumps them in with genuine leaks, so a fixture that opens
no files looked like it leaked two. Discriminating on the
`<inherited from parent>` marker per entry rather than on the summary count is
the right shape.

I checked the property that actually matters — that skipping inherited
descriptors didn't also skip real ones — by mutation, removing the destructor's
release path:

```
mutated (release path removed):   stray fds: 1     definitely lost: 0 bytes
as shipped:                       stray fds: 0
```

Still catches it, and that `definitely lost: 0` beside `stray fds: 1` is the
original insight restated: leak-check alone would have passed this build. The
gate is doing the job it was added for.

## The blocker is untouched

Same repro, same head, release build:

```
SAUCE *a = crackopen("/tmp/fa.txt", "r");   🚽 AAAA
peaceout(a);
SAUCE *b = crackopen("/tmp/fb.txt", "r");   🚽 BBBB
skim(a)                                      🚽 -> "BBBB"
```

`registry_contains()` still matches on address, and `fclose`/`fopen` still hand
back the same one. My two inline comments stand: the fix is a token instead of
the raw pointer in `val.handle.handle`, which needs no ABI change.

# VERDICT

Unchanged. The fd gate is now genuinely sound and I'd merge that part on its
own. The handle registry still can't distinguish a live handle from a stale one
pointing at a recycled address, which is the guarantee the ABI header commits
to in writing.

leo-aa88 (commented):
# APPROVE

The blocker is fixed structurally rather than probabilistically, and the fixture
that proves it is the sharpest kind — one I verified fails against the old
implementation.

## The blocker, on the build where it actually bit

Release build, both shapes:

```
skim(stale a)      → Error: not an open SAUCE -- it was already closed with peaceout...
peaceout(a) twice  → Error: same
```

Previously the first silently returned the *other file's* contents and the
second silently closed it. A counter issued once and never reissued makes that
dead however the allocator behaves, which is the difference between a guarantee
and a coincidence. Starting at 1 so `0` is never valid — and keeping the failed-
`crackopen` arm its own message — is a nice bonus: "you never had a file" stays
distinguishable from "you had one and it's gone". Refusing the wrap instead of
calling it improbable is the right instinct given the whole point was not
resting on improbability.

## The fixture discriminates, and I checked

Reverted `next_token++` to `(uintptr_t)fp` — the old address identity — on a
release build:

```
mutated:      same handle 1     stale write SUCCEEDS, prints "unreachable", exit 0
as shipped:   same handle 0     Error: yapto: not an open SAUCE ...
```

It signals on two independent things — the `same handle` line *and* the presence
of the diagnostic — so neither a message change nor an allocator change can make
it pass vacuously.

**The ordering insight is the part worth keeping.** Using the retired handle
*while `b` is still open* is what makes the broken version succeed; close `b`
first and even address-membership errors, for the wrong reason. Noticing that a
first draft "agreed with itself" is exactly the discipline that was missing when
the original green suite meant nothing.

## Recorded where it will be read

`stdrot_api.h` now carries the requirement — a handle **MUST** be a token, and
a binding must not dereference it, compare it against its own addresses, or
assume anything about equality. That's the line a future socket or cursor
binding needs, and it's in the file that outlives this implementation. Writing
down that ASan and valgrind are *precisely* the two builds that conceal the
difference is the most valuable sentence in the change: it explains why a green
suite was no evidence, which generalises well past file handles.

## Verified

500 collected. Valgrind with the fd gate on all four file fixtures:
`definitely lost: 0`, `stray_fds=0` on every one — including
`file_io_ragequit_no_leak` and both `*_fail` fixtures, which `exit(1)` with a
file still open, so the destructor is genuinely covering the abnormal paths.
`format-check` and `tidy` clean.

# VERDICT

Three rounds: a well-argued design whose key turned out to be wrong, a CI fix
that sharpened the fd gate, and now the key replaced. Appendix B Q6 is answered
in a way that a socket or a database cursor can be built on — because the
answer no longer depends on what `malloc` happens to do.

Ship it.

PR #336, opened 2026-08-31
fix: support member access on pointer expressions

### LINKED ISSUE #322: [Bug]: member access on a parenthesized expression, e.g. `(p + 1).x`, is unsupported
### Description

Member access directly on a parenthesized expression is rejected:

```
Error: Unsupported struct member access expression
```

`resolve_struct_access()` (`ast.c`) dispatches on four base shapes —
`NODE_IDENTIFIER`, `NODE_STRUCT_ACCESS`, `NODE_ARRAY_ACCESS` and
`NODE_FUNC_CALL` — and falls through to that error for anything else. A
`NODE_OPERATION` base such as `(p + 1)` is "anything else".

This is a **pre-existing limitation**, not a regression. It was documented as
such when struct pointer arithmetic landed (#316) precisely because that change
made the shape natural to reach for: once `p = p + 1` works, `(p + 1).x` is the
obvious next thing to write.

Both workarounds are one line, which is why this is a papercut rather than a
blocker.

### Minimal reproduction code

```c
gang E { chad x; };

skibidi main {
    gang E pool[2];
    pool[1].x = 5.0;
    gang E *p = &pool[0];

    yapping("%.1f", (p + 1).x);   🚽 Error
    bussin 0;
}
```

Both of these do work today:

```c
    yapping("%.1f", p[1].x);      🚽 5.0
    p = p + 1;
    yapping("%.1f", p.x);         🚽 5.0
```

### Expected behavior

`5.0` — `(p + 1).x` should mean what `p[1].x` means, as in C (where it is
spelled `(p + 1)->x`).

### Actual behavior

```shell
$ ./brainrot paren.brainrot
Error: Unsupported struct member access expression at line 8
0.0
exit=0
```

Note it reports and continues, yielding `0.0`, rather than aborting.

### Brainrot version / commit

`f98e90fa844937db93242a5564db84ac041d307d`

### Operating system

### Additional context

**Where to look.** `resolve_struct_access()` needs a `NODE_OPERATION` arm that
evaluates the base as a pointer (`evaluate_expression_pointer()`) and resolves
the pointee's tag, the way the existing pointer-variable path does. The
arithmetic itself already works and already scales by the struct's size —
`pointer_arith_scale()` (`ast.c`) handles that — so this is about giving member
access a base shape it currently refuses, not about computing a new address.

**There are THREE dispatches, not one.** This is the trap that has produced
several bugs in this area already, and there is a comment above
`resolve_by_value_struct_source()` in `ast.c` describing it. Adding an
expression shape that can denote a struct means updating:

1. `resolve_struct_access()` (`ast.c`) — which struct is this member read out
   of, and at what address? Runtime.
2. `resolve_by_value_struct_source()` (`ast.c`) — where is a by-value struct
   *value* to copy, and **who owns it**? Runtime.
3. the `NODE_STRUCT_ACCESS` case of `semantic_analyze_with_scope_tracking()`
   (`semantic_analyzer.c`) — is this expression struct-typed at all? Static.

Updating only the first makes `(p + 1).x` work for one hop and fail on
`(p + 1).nested.field`. Updating only the first two leaves the analyzer
rejecting it. Both of those have happened before.

**Suggested testing.** A `test_cases/*.brainrot` fixture covering: the one-hop
read, a **write** through it (`(p + 1).x = 9.0;` — element addresses must be
lvalues, not readable copies), a second hop off it, and passing `(p + 1)`
somewhere by value if that is also made to work. A struct whose size is neither
`sizeof(void *)` (8) nor `sizeof(rizz)` (4) is worth using, so a wrong stride
cannot coincidentally produce the right answer.

w3lld1:
## Description

I added support for member access on parenthesized struct-pointer arithmetic expressions such as `(p + 1).field`.

I preserve the struct/union pointee tag through pointer arithmetic in both static analysis and runtime resolution, evaluate the base expression once, and reject null results cleanly. This also covers nested member access, lvalue writes, and struct fields consumed by value.

I added regression fixtures for:

- one-hop and nested reads;
- writes through the resolved lvalue;
- compound pointer arithmetic and non-pointer-sized struct strides;
- by-value arguments, copy initialization, and returns;
- neighboring-element integrity;
- null pointer arithmetic followed by member access.

## Related Issue

Fixes #322

## Type of Change

- [ ] Breaking change (fix or feature which changes existing behavior)
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Refactor

## Checklist

- [x] I confirm the documentation-only exception is not applicable because my PR changes program behavior and includes regression tests

## Validation

- `make test` — 498 passed, 5 optional raylib tests skipped
- focused exact-head regression run — 2 passed
- `make valgrind` with a sanitizer-free local build — all 442 fixtures reported 0 errors and no leaked file descriptors
- `make format-check`
- `make cppcheck`
- `make tidy`
- `git diff --check`

### DIFF
--- a/ast.c
+++ b/ast.c
@@ -44,6 +44,7 @@ extern void cleanup(void);
 extern const char *vartype_to_string(VarType type);
 extern int yylineno;
 static int get_function_return_pointer_level(const String name);
+static StructDef *get_struct_def_for_expression(ASTNode *expr);
 String evaluate_expression_string(ASTNode *node);
 
 /* ── Native-call memo cache ──────────────────────────────────────────────
@@ -1166,6 +1167,35 @@ bool resolve_struct_access(ASTNode *node, StructDef **def_out, void **base_out,
                 parent_base = element_addr;
             }
         }
+    }
+    else if (obj && obj->type == NODE_OPERATION)
+    {
+        /* `(p + i).field`: pointer arithmetic already preserves the
+           struct-pointer type and scales by the pointee's complete layout.
+           Resolve that same pointee definition here, then evaluate the
+           expression once to obtain the element address. */
+        if (get_expression_pointer_level(obj) != 1)
+        {
+            if (report_errors)
+                yyerror("Member access on a non-struct/union expression");
+            return false;
+        }
+        parent_def = get_struct_def_for_expression(obj);
+        if (!parent_def)
+        {
+            if (report_errors)
+                yyerror("Member access on an expression with an unknown "
+                        "struct or union type");
+            return false;
+        }
+        uintptr_t target = evaluate_expression_pointer(obj);
+        if (!target)
+        {
+            if (report_errors)
+                yyerror("Null pointer dereference in struct member access");
+            return false;
+        }
+        parent_base = (void *)target;
     }
     else if (obj && obj->type == NODE_FUNC_CALL)
     {
@@ -2969,6 +2999,19 @@ static StructDef *get_struct_def_for_expression(ASTNode *expr)
             return get_struct_def_for_expression(operand);
         }
         return NULL;
+    case NODE_OPERATION:
+    {
+        /* Preserve the pointee layout through pointer arithmetic. The
+           numeric operand contributes an offset, not a new pointee type. */
+        int left_level = get_expression_pointer_level(expr->data.op.left);
+        int right_level = get_expression_pointer_level(expr->data.op.right);
+        if (left_level > 0 && right_level == 0 &&
+            (expr->data.op.op == OP_PLUS || expr->data.op.op == OP_MINUS))
+            return get_struct_def_for_expression(expr->data.op.left);
+        if (right_level > 0 && left_level == 0 && expr->data.op.op == OP_PLUS)
+            return get_struct_def_for_expression(expr->data.op.right);
+        return NULL;
+    }
     default:
         return NULL;
     }
--- a/ast.c
+++ b/ast.c
@@ -1188,6 +1188,10 @@ bool resolve_struct_access(ASTNode *node, StructDef **def_out, void **base_out,
                         "struct or union type");
             return false;
         }
+        if (!report_errors)
+        {
+            return false;
+        }
         uintptr_t target = evaluate_expression_pointer(obj);
         if (!target)
         {
--- a/ast.c
+++ b/ast.c
@@ -477,6 +477,12 @@ static StructField *static_struct_field(ASTNode *node)
         }
         parent = get_struct_def(outer->desc.struct_name);
     }
+    else if (obj && obj->type == NODE_OPERATION)
+    {
+        /* Pointer arithmetic has no storage to inspect during a no-eval
+           query, but its pointee layout is still statically available. */
+        parent = get_struct_def_for_expression(obj);
+    }
 
     if (!parent)
     {
--- a/test_cases/parenthesized_struct_access.brainrot
+++ b/test_cases/parenthesized_struct_access.brainrot
@@ -1,0 +1,63 @@
+🚽 Regression for #322: member access on a parenthesized struct-pointer
+🚽 expression. Wide is 24 bytes so the second element only resolves when
+🚽 pointer arithmetic uses the pointee's complete struct layout.
+gang Inner {
+    chad value;
+    rizz marker;
+};
+
+gang Wide {
+    gigachad left;
+    gang Inner nested;
+    gigachad right;
+};
+
+skibidi show_inner(gang Inner value) {
+    yapping("by value: %.1f %d", value.value, value.marker);
+}
+
+gang Inner pick_inner(gang Wide *p) {
+    bussin (p + 1).nested;
+}
+
+skibidi main {
+    gang Wide pool[3];
+    pool[0].left = 10.0;
+    pool[0].nested.value = 11.0;
+    pool[0].nested.marker = 12;
+    pool[0].right = 13.0;
+    pool[1].left = 20.0;
+    pool[1].nested.value = 21.0;
+    pool[1].nested.marker = 22;
+    pool[1].right = 23.0;
+    pool[2].left = 30.0;
+    pool[2].nested.value = 31.0;
+    pool[2].nested.marker = 32;
+    pool[2].right = 33.0;
+
+    gang Wide *p = &pool[0];
+
+    🚽 One-hop read and a second member-access hop.
+    yapping("read: %.1f %.1f %d", (p + 1).right,
+            (p + 1).nested.value, (p + 1).nested.marker);
+
+    🚽 The resolved element must remain an lvalue.
+    (p + 1).nested.value = 91.0;
+    (p + 1).right = 93.0;
+    yapping("write: %.1f %.1f", pool[1].nested.value, pool[1].right);
+
+    🚽 A compound operation still carries the original pointee tag.
+    yapping("compound: %.1f", (p + 2 - 1).left);
+
+    🚽 Struct-valued fields off the expression work at each by-value
+    🚽 consumer, not only for an immediate scalar read.
+    show_inner((p + 1).nested);
+    gang Inner copy = (p + 1).nested;
+    yapping("copy: %.1f %d", copy.value, copy.marker);
+    gang Inner returned = pick_inner(p);
+    yapping("return: %.1f %d", returned.value, returned.marker);
+
+    🚽 Neighboring elements must remain intact after the writes.
+    yapping("neighbors: %.1f %.1f", pool[0].right, pool[2].right);
+    bussin 0;
+}
--- a/test_cases/parenthesized_struct_access.brainrot
+++ b/test_cases/parenthesized_struct_access.brainrot
@@ -57,6 +57,11 @@ skibidi main {
     gang Inner returned = pick_inner(p);
     yapping("return: %.1f %d", returned.value, returned.marker);
 
+    🚽 Type queries must resolve the parenthesized pointer expression without
+    🚽 evaluating it, just like the equivalent pointer-index expression.
+    yapping("no-eval: %lu %lu", maxxing((p + 1).nested),
+            maxxing(p[1].nested));
+
     🚽 Neighboring elements must remain intact after the writes.
     yapping("neighbors: %.1f %.1f", pool[0].right, pool[2].right);
     bussin 0;
--- a/docs/the-brainrot-programming-language.md
+++ b/docs/the-brainrot-programming-language.md
@@ -615,8 +615,8 @@ skibidi main {
   bounds-checked, unlike the array form (`pts[i]`) which is. Only
   single-level pointers and a single index are accepted: `gang E **pp` needs
   an explicit dereference, and `p[0][1]` would be indexing into a struct.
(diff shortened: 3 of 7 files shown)


ast.c
```diff
@@ -1167,6 +1168,35 @@ bool resolve_struct_access(ASTNode *node, StructDef **def_out, void **base_out,
...
+        }
+        parent_def = get_struct_def_for_expression(obj);
+        if (!parent_def)
+        {
+            if (report_errors)
+                yyerror("Member access on an expression with an unknown "
+                        "struct or union type");
+            return false;
+        }
+        uintptr_t target = evaluate_expression_pointer(obj);
```
leo-aa88, line 1201: **MAJOR (latent, not currently reachable).** This arm calls `evaluate_expression_pointer(obj)` unconditionally — it does **not** honor the `report_errors == false` "do not execute" invariant that the `NODE_FUNC_CALL` arm 20 lines up spends a whole paragraph defending ("running a user function from any of them would execute the program out of order, or while it is still being parsed").

Today that's safe, but only by accident: `get_struct_def_for_expression()` has **no `NODE_FUNC_CALL` case**, so `(get_p() + 1).x` resolves `parent_def == NULL` and this returns false *before* the `evaluate_expression_pointer` line ever runs. The "evaluate once" is guarded by a side effect of an unrelated omission in another function.

The moment someone adds a `NODE_FUNC_CALL` case to `get_struct_def_for_expression()` — the obvious next feature: member access on a call that returns a struct *pointer*, `(make_p() + 1).x` — this arm will start executing `make_p()` inside every `report_errors == false` type probe (`get_expression_type`, `infer_expression_pointer_level`, `handle_sizeof`, …). Out of order, possibly at analysis time, possibly several times per expression. That is precisely the failure mode the `NODE_FUNC_CALL` arm exists to prevent, and this arm reproduces none of its guard.

Please either (a) add the same `if (!report_errors) return false;` gate this arm's sibling has and resolve the eval'd base only in execute mode, or (b) at minimum document, right here and in `get_struct_def_for_expression()`, that the "no executable sub-expression is statically resolvable to a pointee tag" property is load-bearing for the do-not-execute contract — so the next person adding a call case knows they're arming a landmine.

leo-aa88 (commented):
Correct, and it does what #322 asked. I fetched the branch, built with ASAN/UBSAN, and both fixtures pass; I also ran an adversarial matrix beyond the fixtures — reversed operand `(1 + p).x`, `chungus` (union) pointee read+write, multi-level `(pp + 1).x` (cleanly rejected by the analyzer), non-pointer `(a + b).x` (rejected), null base, and nested/by-value/return hops — all behave. The struct/union tag really is preserved through the arithmetic and the stride uses the full pointee layout (the 24-byte `Wide` fixture proves a wrong stride can't fluke the right answer).

On the "THREE dispatches" trap the issue warns about: `resolve_by_value_struct_source()` is correctly handled *without* a new arm, because `(p + 1).nested` is a `NODE_STRUCT_ACCESS` that delegates to `resolve_struct_access()`, and `(p + 1)` itself is a pointer (never a by-value struct). Good call not adding a redundant arm there.

One latent, forward-looking concern below. It is **not** a current bug — I traced that no reachable expression triggers it today — but it's a landmine for the next feature in this area, and worth a comment or a guard before merge.

(Note: `make test` shows unrelated failures on my checkout from uncommitted local WIP in the native-module subsystem; the two fixtures here pass in isolation.)

leo-aa88 (changes requested):

w3lld1:
Thanks — addressed the speculative-evaluation concern in `1da3279`. The `NODE_OPERATION` path now returns without calling `evaluate_expression_pointer()` whenever `report_errors == false`, matching the no-execution invariant used for function-call member access.

Validation:
- `clang-format --dry-run --Werror ast.c`
- `git diff --check`

I could not run the focused executable test locally because this checkout requires `bison`, which is unavailable in the environment; CI has started on the pushed head.

ast.c
```diff
@@ -1167,6 +1168,39 @@ bool resolve_struct_access(ASTNode *node, StructDef **def_out, void **base_out,
...
+        }
+        parent_def = get_struct_def_for_expression(obj);
+        if (!parent_def)
+        {
+            if (report_errors)
+                yyerror("Member access on an expression with an unknown "
+                        "struct or union type");
+            return false;
+        }
+        if (!report_errors)
```
leo-aa88, line 1197: **[bug]** This gate is correct — do not evaluate here when `report_errors` is false. It also makes `static_struct_field()` the only remaining no-eval answer for this shape, and that helper still has no `NODE_OPERATION` arm (only `NODE_FUNC_CALL` and `NODE_STRUCT_ACCESS`).

`infer_runtime_expression_type_noeval()` (used by `maxxing` so the operand is never run) does:

```c
if (!resolve_struct_access(..., false)) {
    fld = static_struct_field(expr);
    if (!fld)
        return NONE;
}
```

I built this commit and ran:

```
maxxing((p + 1).nested)  → error, 0
maxxing(p[1].nested)     → 8
```

The analyzer already infers the tag (`infer_struct_def_static` got a `NODE_OPERATION` case). The runtime no-eval path does not. That is a type that becomes NONE halfway through the pipeline. `f().x` needed `static_struct_field` for exactly this reason; `(p + 1).x` needs the same arm (pointer_level == 1, then `get_struct_def_for_expression(obj)`).

Add `maxxing((p + 1).nested)` / `maxxing((p + 1).left)` to the fixture. The current tests cannot fail this implementation.

test_cases/parenthesized_struct_access.brainrot
```diff
@@ -0,0 +1,63 @@
...
+    pool[1].right = 23.0;
+    pool[2].left = 30.0;
+    pool[2].nested.value = 31.0;
+    pool[2].nested.marker = 32;
+    pool[2].right = 33.0;
+
+    gang Wide *p = &pool[0];
+
+    🚽 One-hop read and a second member-access hop.
+    yapping("read: %.1f %.1f %d", (p + 1).right,
```
leo-aa88, line 41: **[bug]** Covers reads, writes, compound arithmetic, by-value, and neighbors. Does not cover a no-eval type query on the new shape. `maxxing((p + 1).nested)` vs `maxxing(p[1].nested)` is the case that actually exercises `static_struct_field` / `infer_runtime_expression_type_noeval`, and it fails on this head.

leo-aa88 (changes requested):
## Summary

The `report_errors == false` gate on the new `NODE_OPERATION` arm is the right fix for the previous speculative-eval finding. Runtime `(p + 1).field` reads/writes, nested hops, by-value copy/return, and the 24-byte stride fixture still do what #322 asked.

The no-eval type path was not updated. `static_struct_field()` still only understands call bases and member-access chains. `maxxing()` / `infer_runtime_expression_type_noeval()` therefore reports NONE for `(p + 1).nested` even though the analyzer already knows the tag. I built this head with ASAN/UBSAN and ran it: `maxxing((p + 1).nested)` errors and prints 0; `maxxing(p[1].nested)` prints 8. CI is green. This is not a failing-test problem. The fixtures never call `maxxing`.

## Issue counts by severity

- bugs: 1
- suggestions: 0
- nits: 0

w3lld1:
Thanks — fixed the no-eval type path in `3c1330b`. `static_struct_field()` now resolves the pointee layout for a `NODE_OPERATION` base without evaluating it, and the regression fixture compares `maxxing((p + 1).nested)` with `maxxing(p[1].nested)`; both report 8.

Validation on the pushed commit:
- targeted `brainrot` build with ASAN/UBSAN and `-Werror`;
- `./brainrot test_cases/parenthesized_struct_access.brainrot` (output matches the fixture exactly);
- `clang-format --dry-run -Werror ast.c`;
- `git diff --check`.

test_cases/parenthesized_struct_access.brainrot
```diff
@@ -0,0 +1,68 @@
+🚽 Regression for #322: member access on a parenthesized struct-pointer
```
leo-aa88, line 1: **[bug]** This fixture is now the proof that `(p + 1).field` works. `docs/the-brainrot-programming-language.md` §7.9 still lists that exact shape as a current limitation, and `test_cases/struct_pointer_arithmetic.brainrot` still calls it "deliberately absent" / "still unsupported."

The tests prove the implementation agrees with itself. The public spec still agrees with the bug. Update both in this PR: delete the limitation bullet (replace with the `p[1].x` equivalence), and rewrite the sibling fixture's "Deliberately absent" block so it points here instead of documenting a hole that no longer exists.

leo-aa88 (changes requested):
# REQUEST CHANGES

The runtime, analyzer, and no-eval type paths now agree on `(p + 1).field`. I built `3c1330b` with ASAN/UBSAN and the two fixtures match expected output. Reversed operand, union pointee, sizeof-must-not-evaluate, multi-level rejection, and non-pointer rejection all behave. That is the feature #322 asked for.

It still cannot merge. The public language spec continues to document the bug as a current limitation, and the sibling pointer-arithmetic fixture still tells the next reader the shape is "deliberately absent." The implementation is done. The contract is not.

## 1. Language reference still documents the bug as intended behavior

**BLOCKING.**

`docs/the-brainrot-programming-language.md` §7.9 "Current Limitations" still says:

> Member access directly on a parenthesized expression (`(p + 1).x`) is not supported; assign first (`p = p + 1; p.x`) or index (`p[1].x`).

That sentence is the specification of the bug this PR claims to fix. After this change, the runtime does the opposite. The next engineer who treats the language reference as true will keep writing the workarounds and will believe `resolve_struct_access()` still has no `NODE_OPERATION` arm.

Same lie, second site: `test_cases/struct_pointer_arithmetic.brainrot` lines 40–46 still call `(p + 1).x` "deliberately absent" and "still unsupported." That comment is now a landmine sitting on top of a fixture whose whole job is to describe what struct pointer arithmetic can and cannot do.

This is not optional documentation. AGENTS.md requires `docs/the-brainrot-programming-language.md` to track implementation status. The PR checklist claims the change was documented. Code comments are not a substitute for the public spec still stating the pre-fix limitation.

Delete the limitation bullet. Replace it with the actual rule: `(p + 1).x` is `p[1].x`, lvalue, scaled by the pointee layout. Rewrite the "Deliberately absent" block in `struct_pointer_arithmetic.brainrot` to point at `parenthesized_struct_access.brainrot` instead of asserting the old hole.

## Issue counts by severity

- bugs: 0
- specification: 1 (blocking)
- suggestions: 0

# VERDICT

This is a specification problem, not an implementation problem. The three dispatches, the `report_errors == false` gate, and `static_struct_field()`'s `NODE_OPERATION` arm now do what they claim. The language reference and the sibling fixture comment still claim the opposite.

Do not merge a feature whose public spec says it does not exist.

w3lld1:
Thanks — updated the public contract in `b4b6ebd`:

- the language reference now documents `(p + 1).x` as equivalent to `p[1].x` for reads and assignments;
- the sibling pointer-arithmetic fixture now points to the dedicated parenthesized-access regression instead of describing the old limitation.

Validation: `git diff HEAD^ --check`. The focused executable test could not be rerun in this environment because `bison` is unavailable; the change is limited to documentation and fixture comments.

leo-aa88 (approved):

Issue #337, opened 2026-08-31
[Feature]: Native Windows build support


leo-aa88:
## Problem / motivation

Brainrot has **no native Windows build**. The release workflow
(`.github/workflows/release.yml`) ships native artifacts for linux/amd64,
linux/arm64, darwin/amd64, and darwin/arm64 only, and says so explicitly in its
header:

> Windows is omitted: the loader and stdrot are POSIX (dlfcn.h, libstdrot.so, unistd.h).

This blocks Windows users from running Brainrot without WSL, and it blocks
downstream projects from shipping Windows builds. Concretely, it just blocked
the [`tung-tung-sahur`](https://github.com/Brainrotlang/tung-tung-sahur) game
release: its per-platform bundles cover Linux and macOS, and Windows is a
WSL-only stopgap precisely because there is no `brainrot.exe` to bundle.

This issue proposes making the interpreter build and run natively on Windows,
and adding a `windows/amd64` leg to the release matrix.

## Current state — the POSIX surface

Reading the source, the Windows-blocking surface is small and contained — four
spots, not a rewrite. Everything else (Flex lexer, Bison parser, AST, semantic
analyzer, tree-walking interpreter) is portable C.

| # | Where | POSIX dependency | Windows equivalent |
| --- | --- | --- | --- |
| 1 | `stdrot.c` (~L541) | `dlopen("libstdrot.so")` to load core builtins | — (already solvable, see below) |
| 2 | `stdrot.c` `stdrot_load_module()` + `lib/module_path.c` | `dlopen`/`dlsym` + hardcoded `.so` for `#cooked <name>` native modules | `LoadLibraryW` / `GetProcAddress`, `.dll` |
| 3 | `lib/module_path.c` (L12/L17/L18) | `<libgen.h>` (`dirname`/`basename`), `<sys/stat.h>`, `<unistd.h>`, `/` separators | `_splitpath_s`, `GetFileAttributes`, `\` |
| 4 | `stdrot/ragequit.c` (L17) | `sleep()` from `<unistd.h>` | `Sleep()` (milliseconds) |

### #1 is already half-solved by `STDROT_STATIC`

The hardest blocker — `dlopen`'ing `libstdrot.so` at startup — already has a
compile-time escape hatch. The **wasm** build defines `-DSTDROT_STATIC`
(`make wasm`, `Makefile`), which compiles every builtin directly into the binary
instead of loading a `.so`. Under `STDROT_STATIC`:

[...]

PR #338, opened 2026-08-31
feat: native Windows build — core interpreter (issue #337, Stage 1)

leo-aa88:
## Description

**Stage 1 of native Windows support (#337): the core interpreter builds and runs natively on Windows.**

Windows had no native build — `release.yml` omits it and the loader/stdrot were POSIX. This makes a single, statically-linked `brainrot.exe` (`-DSTDROT_STATIC`, exactly like the `wasm` target — `stdrot` compiled in, no `libstdrot.so`/`dlopen`) via MinGW-w64. It runs pure-Brainrot programs and `#cooked` **prelude** modules. It does **not** run `#cooked <native.dll>` native modules — `STDROT_STATIC` compiles the dynamic-module loader out — which is **Stage 2** (Win32 `LoadLibraryW`), tracked in #337.

The POSIX surface turned out to be four spots, all now behind `_WIN32` guards so **POSIX builds are byte-for-byte unchanged**:

- **`lib/module_path.c`** — `GetModuleFileNameA` for the running-executable probe (vs `/proc/self/exe`); a `path_canonical()` wrapper (`realpath` → `_fullpath` + a `GetFileAttributesA` existence check); an in-place directory split (vs POSIX `dirname`); and a `;`-separated `$BRAINROT_PATH` on Windows.
- **`lang.l`** — the `#cooked` subsystem also used `realpath`/`dirname`/`basename`; ported with local helpers. It deliberately does **not** include `<windows.h>`: that header `typedef`s `CHAR`, which collides with the grammar's token enum in `lang.tab.h` (this cost me a debugging round — the cross-compile caught it). The Win32 canonicalizer uses `_fullpath` (`<stdlib.h>`) + `stat` instead.
- **`stdrot/ragequit.c`** — `chill()` uses `Sleep()` (ms), with an overflow cap so it can never become `INFINITE`.
- **`stdrot/gamba.c`** — a `BCryptGenRandom` CSPRNG backend for Windows, taking precedence over the `STDROT_STATIC` stub, so `gamba` stays a real CSPRNG with **no OpenSSL dependency** (`-lbcrypt`).

Build/CI/docs:
- **Makefile** — `make windows` target (mirrors `wasm`, MinGW-w64 gcc, `-lbcrypt`). `-Wpedantic` is dropped for it (it rejects the compound-literal `.return_type` initializer in `STDROT_EXPORT_SIG`, which the native `libstdrot.so` build also compiles without `-Wpedantic`; Clang/emcc accept it).
- **ci.yml** — a `windows` job (MSYS2/MINGW64) that builds `brainrot.exe` and smoke-tests it against the runnable examples + `#cooked` prelude modules, and uploads the exe.
- **README** — a "Native Windows build" section.

## Related Issue

Part of #337 (Stage 1 of 3). Does not fully close it — Stages 2 (Win32 native-module loading) and 3 (release-matrix `windows/amd64`) remain.

## Type of Change

## Checklist

- [x] I have added tests that prove my change works, at the lowest appropriate layer — **the appropriate layer here is CI**: Windows portability can't be proven by a `test_cases/*.brainrot` fixture, so this adds a `windows` CI job that builds `brainrot.exe` on `windows-latest` and runs it against the examples + `#cooked` prelude modules. All existing `test_cases` continue to run unchanged on POSIX.
- [x] Those tests cover the happy path and error/edge cases — the smoke test asserts exact `hello_world` output, runs every dependency-free example, and exercises both relative and `$BRAINROT_PATH` `#cooked` resolution (the code paths I ported).
- [x] If this PR cannot affect program behavior … — it *can* affect the `#cooked`/module-path code on all platforms, so I validated that below rather than skipping.
- [x] I have run `make format-check` locally — passes.
- [x] I have run the unit tests locally — full `pytest` suite **503 passed** against a native (system-gcc) build, which is **ASan+UBSan-instrumented**, so it covers memory safety on every `#cooked`/module-path path I changed.
- [ ] I have run the valgrind memory tests locally — **could not run in this environment** (the default toolchain here can't link the sanitizer binary — a pre-existing `libstdc++`/glibc mismatch unrelated to this change — and local `cppcheck` is 2.7 < the required 2.13). The ASan/UBSan pytest run above is the memory-safety evidence; CI's `test` (valgrind) and `static-analysis` (cppcheck) jobs are the authoritative check.
- [x] All new and existing tests pass (locally, on POSIX)

### How the Windows paths were validated locally

Beyond the POSIX suite: cross-compiled the full `make windows` target with MinGW-w64 (`x86_64-w64-mingw32-gcc`) — it links cleanly (incl. `-lbcrypt`) to a valid `PE32+ x86-64` `brainrot.exe`, exercising every `_WIN32` branch at compile+link time. Runtime on real Windows is what the new CI job covers (no Wine available here).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -387,16 +387,19 @@ jobs:
           out="$(./brainrot.exe examples/hello_world.brainrot)"
           test "$out" = "Hello, World!" \
             || { echo "hello_world mismatch: [$out]"; exit 1; }
-          # Every dependency-free example must run and exit 0. Excludes
-          # file_wordcount (reads a path) and examples/raylib/ (native module).
+          # Every dependency-free STANDALONE example must run and exit 0.
+          # Excludes: file_wordcount (reads a path), examples/raylib/ (native
+          # module), and mathutils (a #cooked library with no `skibidi main` --
+          # it's exercised via modules_named below, not run on its own).
           for f in fibonacci fizz_buzz bubble_sort sieve_of_eras twoSum \
-                   array_of_structs string_parsing string_toolkit mathutils \
+                   array_of_structs string_parsing string_toolkit \
                    heat_equation_1d; do
             echo "-- $f --"
             ./brainrot.exe "examples/$f.brainrot" >/dev/null \
               || { echo "example $f failed"; exit 1; }
           done
           # #cooked prelude modules (relative + BRAINROT_PATH) work statically.
+          # modules_named cooks <mathutils>, so that module is covered here.
           ./brainrot.exe examples/modules.brainrot >/dev/null
           BRAINROT_PATH=examples ./brainrot.exe examples/modules_named.brainrot >/dev/null
           echo "windows smoke tests passed"
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -165,6 +165,64 @@ jobs:
           os: ${{ matrix.os }}
           artifact-name: brainrot-release-${{ matrix.os }}-${{ matrix.arch }}
 
+  # Windows release dry-run (issue #337). Windows can't share the POSIX
+  # release-dry-run matrix above: that path runs `make release` and packages
+  # brainrot + libstdrot.so via the native-release-build composite action,
+  # whereas Windows is a single self-contained brainrot.exe from `make windows`
+  # (STDROT_STATIC, no libstdrot.so) built under MSYS2 -- hence a sibling job.
+  # Gated on the `windows` smoke job, mirroring how the POSIX dry-run needs
+  # `test`. No publish; just proves the shippable artifact builds and is
+  # self-contained. Wiring this into the real release (release.yml) is Stage 3.
+  release-dry-run-windows:
+    name: Release dry-run windows/amd64
+    if: github.event_name == 'pull_request'
+    runs-on: windows-latest
+    needs: windows
+    timeout-minutes: 20
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v7
+
+      - name: Set up MSYS2 (MinGW-w64 toolchain)
+        uses: msys2/setup-msys2@v2
+        with:
+          msystem: MINGW64
+          update: true
+          install: >-
+            mingw-w64-x86_64-gcc
+            make
+            bison
+            flex
+
+      - name: Build release brainrot.exe
+        shell: msys2 {0}
+        run: make windows
+
+      # A shipped .exe must import only system DLLs. A dependency on
+      # libgcc_s_seh-1.dll / libwinpthread-1.dll would mean it only runs where
+      # MSYS2's bin is on PATH -- exactly the "works on the builder" trap the
+      # POSIX job's otool check guards against. -static (Makefile) prevents it;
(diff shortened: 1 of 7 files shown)


leo-aa88 (commented):
LGTM.

I traced the whole change and validated it locally rather than trusting the description:

- **POSIX unchanged, verified.** Full `make test` → **501 passed**, no regressions. The claim that matters isn't "byte-for-byte" (it isn't — `libgen` `dirname()`/`basename()` are replaced by hand-rolled `br_path_last_sep`/`strrchr` helpers on *all* platforms), it's that behavior is identical for the inputs that actually reach these paths. It is: `current_filename`/`exe_path` are always canonical absolute paths by the time dirname/basename run (`br_path_canonical`/`realpath`), so they always contain a separator and the helpers produce exactly what libgen did. The divergences (`dirname("/foo")` → `""` vs `"/"`, trailing-slash basename) are unreachable and, where near-reachable, wash out under the subsequent canonicalization.
- **Windows compiles, links, and is self-contained.** Cross-built `make windows` with `x86_64-w64-mingw32-gcc` → valid `PE32+ x86-64`, links `-lbcrypt`, and `objdump -p` imports only `bcrypt.dll` / `KERNEL32` / the UCRT `api-ms-win-crt-*` — no `libgcc`/`libwinpthread`/`libstdc++`. Every `_WIN32` branch is exercised at compile+link. The `release-dry-run-windows` objdump guard will hold.
- **`gamba` CSPRNG** — `BCryptGenRandom(NULL, …, BCRYPT_USE_SYSTEM_PREFERRED_RNG)`, NTSTATUS `!= 0` = failure, ordered *before* the `STDROT_STATIC` stub so the Windows static build gets the real CSPRNG. Correct, and the `#endif` closes the three-way cleanly.
- **`chill()` overflow cap** — the `> 4294967` gate is exactly right: `4294967 * 1000 = 4,294,967,000 < 2^32`, so the un-capped branch never wraps, and the cap is `0xFFFFFFFE`, so a finite `chill()` can never degrade into `INFINITE`. Good.
- **`lang.l` deliberately not pulling in `<windows.h>`** (the `CHAR` token-enum collision with `lang.tab.h`) is the right call and well-commented; `_fullpath` + `stat` cover it.
- CI smoke list: every referenced example exists and runs; `modules_named` genuinely cooks `<mathutils>`, so the angle-name + `BRAINROT_PATH` path is covered.

Two non-blocking nits, take or leave:
1. The PR text's "POSIX builds are byte-for-byte unchanged" overstates it — the *source path* changed for everyone; it's the *behavior* that's unchanged. Worth softening so a future reader doesn't assume POSIX still routes through libgen.
2. The Windows smoke test sets `BRAINROT_PATH=examples` (single entry), so the `;`-vs-`:` list split is never exercised with more than one component. A two-entry `BRAINROT_PATH` in the smoke job would actually cover the separator difference you introduced.

Scope is honest (Stage 1, native `#cooked <native.dll>` explicitly out), guards are clean, no memory-safety concern on the changed paths (ASan/UBSan suite is green). Ship it — Stage 2's `LoadLibraryW` is the real test of this foundation.

(Formally submitting as a comment, not an Approve: GitHub blocks approving your own PR.)

PR #339, opened 2026-08-31
feat: native modules on Windows (#cooked <name> via LoadLibraryA) — issue #337, Stage 2

leo-aa88:
## Description

**Stage 2 of native Windows support (#337): `#cooked <name>` native modules load on Windows.**

Stage 1 shipped a static `brainrot.exe` but `STDROT_STATIC` compiled out *all* dynamic loading, so native modules were unavailable. This decouples "core builtins static" from "native modules loadable" and adds a Win32 loader path, so a `#cooked <name>` module works on Windows while the core stays statically linked.

- **`stdrot.c`** — a `STDROT_DYNAMIC_MODULES` capability macro (a loader exists on any POSIX build *or* on Windows — just not wasm) and a small `br_module_open`/`sym`/`close`/`error` shim over `dlopen` and the Win32 loader (`LoadLibraryA`/`GetProcAddress`/`FreeLibrary`/`FormatMessageA`). The cooked-module state, `stdrot_load_module()`, and the module-freeing half of `stdrot_unload()` now key on that capability instead of `!STDROT_STATIC`; the core-library load/unload still keys on `STDROT_STATIC`. On POSIX the shim *is* `dlopen`/`dlsym`/`dlclose`, so behavior there is unchanged.
- **`lib/module_path.c`** — resolves a native module under the same predicate (`MODULE_HAS_NATIVE`), as `.dll` on Windows and `.so` elsewhere.
- **`Makefile`** — a `tests/nativemodules/%.dll` rule + a `windows-test-module` target (same `registry.c`-per-file shape as the `.so` rule).
- **`.github/workflows/ci.yml`** — the `windows` job now builds a native module DLL and proves `#cooked <testnative>` `LoadLibraryA`-loads it and `tripled(14) → 42`.
- **`README.md`** — Windows section updated: native modules now supported.

### Why the core can stay static but modules still load

`dlopen` of `libstdrot.so` was the hard part of a Windows core port; Stage 1 sidestepped it by compiling the core in (`STDROT_STATIC`). Native modules are a *separate* concern: they're loaded by explicit handle (`RTLD_LOCAL` / the Win32 per-module scope), reference no main-binary globals, and self-register via `registry.c`'s linker-section collection — which MinGW/PE supports (`__start_/__stop_stdrot_exports`). A module `.dll` links standalone and exports `brainrot_module_init_v3`, exactly what the loader looks up.

## Related Issue

Part of #337 (Stage 2 of 3). Stage 3 (publishing a Windows artifact from `release.yml`) remains.

## Type of Change

## Checklist

- [x] I have added tests that prove my change works, at the lowest appropriate layer — the `windows` CI job now builds a native module DLL and runs a `#cooked <testnative>` program end-to-end (`tripled(14) → 42`), which exercises resolution (`.dll`), `LoadLibraryA`, `GetProcAddress` of `brainrot_module_init_v3`, the registry-section enumeration on PE, and a native call/return. The POSIX native-module suite (`test_native_module_*`, `test_native_struct_*`, `test_bad_registry_*`, `test_old_abi_*`) continues to run unchanged and still exercises the shared `stdrot_load_module()` path.
- [x] Those tests cover the happy path and error/edge cases — the POSIX loader-rejection fixtures (malformed registry, missing/incompatible `brainrot_module_init_v3`, duplicate exports, pre-v2 ABI) all still pass through the refactored code.
- [x] If this PR cannot affect program behavior … — it does affect the shared loader path on all platforms, validated by the full suite below.
- [x] I have run `make format-check` locally — passes.
- [x] I have run the unit tests locally — full `pytest` **501 passed** against a native (system-gcc) build, ASan+UBSan-instrumented, covering the refactored `stdrot_load_module`/`stdrot_unload`/`module_path_resolve`.
- [ ] I have run the valgrind memory tests locally — **could not** in this environment (the default toolchain can't link the sanitizer binary — a pre-existing `libstdc++`/glibc mismatch — and local `cppcheck` is 2.7 < the required 2.13). The ASan/UBSan pytest run is the memory evidence; CI's `test` (valgrind) and `static-analysis` (cppcheck) jobs are authoritative.
- [x] All new and existing tests pass (locally, on POSIX)

### Windows validation done locally

Cross-compiled with MinGW-w64: `brainrot.exe` (self-contained, still only system-DLL imports) and `tests/nativemodules/testnative.dll` (a valid PE DLL that **exports `brainrot_module_init_v3`**). The one thing a cross-compile can't check — the runtime module load/call and PE section enumeration — is what the new CI job proves on a real `windows-latest` runner (no Wine available here).

### DIFF
--- a/stdrot.c
+++ b/stdrot.c
@@ -34,9 +34,93 @@
 #include <stdlib.h>
 #include <string.h>
 #include <stdarg.h>
-#ifndef STDROT_STATIC
-#include <dlfcn.h>
+
+/* A dynamic loader for #cooked <name> native modules exists wherever the
+ * platform provides one -- POSIX dlopen (any non-wasm build) or the Win32
+ * loader -- INDEPENDENTLY of whether the CORE library is statically linked.
+ * On Windows the core is compiled in (STDROT_STATIC, no libstdrot.so) yet
+ * native modules still load via LoadLibraryA (issue #337 Stage 2). Only the
+ * wasm build (STDROT_STATIC && !_WIN32) has no loader at all. lib/module_path.c
+ * mirrors this exact predicate to decide whether to resolve a native module. */
+#if !defined(STDROT_STATIC) || defined(_WIN32)
+#define STDROT_DYNAMIC_MODULES 1
 #endif
+
+#ifndef STDROT_STATIC
+#include <dlfcn.h> /* core library loader (dlopen/dlsym) -- POSIX native build */
+#elif defined(_WIN32)
+#include <stdint.h> /* uintptr_t, for the GetProcAddress round-trip below */
+#include <windows.h> /* Win32 native-module loader (LoadLibraryA/GetProcAddress) */
+#endif
+
+/* ── Native-module loader shim ────────────────────────────────────────────
+ * One trio -- open/sym/close -- over dlopen and the Win32 loader, so
+ * stdrot_load_module() and stdrot_unload() read identically on both. The
+ * existing void* handle fields hold either a dlopen handle or an HMODULE
+ * (a pointer). Only compiled where a loader exists. */
+#ifdef STDROT_DYNAMIC_MODULES
+#if defined(_WIN32)
+static void *br_module_open(const char *path)
+{
+    return (void *)LoadLibraryA(path);
+}
+static void *br_module_sym(void *handle, const char *name)
+{
+    /* GetProcAddress returns FARPROC; round-trip through uintptr_t to a data
+     * pointer, the same shape dlsym() hands back. */
+    return (void *)(uintptr_t)GetProcAddress((HMODULE)handle, name);
+}
+static void br_module_close(void *handle)
+{
+    FreeLibrary((HMODULE)handle);
+}
+/* Formats the last loader error into a static buffer (single-threaded loader
+ * path, same as dlerror()'s own not-thread-safe contract). */
+static const char *br_module_error(void)
+{
+    static char buf[256];
+    DWORD err = GetLastError();
+    DWORD n = FormatMessageA(FORMAT_MESSAGE_FROM_SYSTEM |
+                                 FORMAT_MESSAGE_IGNORE_INSERTS,
+                             NULL, err, 0, buf, (DWORD)sizeof(buf), NULL);
+    if (n == 0)
+    {
+        snprintf(buf, sizeof(buf), "Win32 error %lu", (unsigned long)err);
+    }
+    else
+    {
+        /* Trim the trailing CR/LF FormatMessage appends. */
+        while (n > 0 && (buf[n - 1] == '\n' || buf[n - 1] == '\r'))
+        {
+            buf[--n] = '\0';
+        }
+    }
+    return buf;
+}
+#else
+/* RTLD_LOCAL, unlike the core library's own dlopen(): a cooked module is
+ * looked up entirely by explicit handle (br_module_sym below searches that
+ * object only), so its exports -- including its own brainrot_module_init_v3,
+ * shared by every module -- never enter the process-wide symbol scope, and
+ * two modules exporting that name never collide. */
+static void *br_module_open(const char *path)
+{
+    return dlopen(path, RTLD_LAZY | RTLD_LOCAL);
+}
+static void *br_module_sym(void *handle, const char *name)
+{
+    return dlsym(handle, name);
+}
+static void br_module_close(void *handle)
+{
+    dlclose(handle);
+}
+static const char *br_module_error(void)
+{
+    return dlerror();
+}
+#endif
+#endif /* STDROT_DYNAMIC_MODULES */
 
 /* ── Global execution context ────────────────────────────────────────────── */
 ExecutionContext g_exec_context = {0, {NULL, 0}, {NULL, 0}};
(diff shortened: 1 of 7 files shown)


stdrot.c
```diff
@@ -34,9 +34,93 @@
...
 #include <stdarg.h>
+
+/* A dynamic loader for #cooked <name> native modules exists wherever the
+ * platform provides one -- POSIX dlopen (any non-wasm build) or the Win32
+ * loader -- INDEPENDENTLY of whether the CORE library is statically linked.
+ * On Windows the core is compiled in (STDROT_STATIC, no libstdrot.so) yet
+ * native modules still load via LoadLibraryA (issue #337 Stage 2). Only the
+ * wasm build (STDROT_STATIC && !_WIN32) has no loader at all. lib/module_path.c
+ * mirrors this exact predicate to decide whether to resolve a native module. */
+#if !defined(STDROT_STATIC) || defined(_WIN32)
```
leo-aa88: **MINOR (maintainability).** This is now the *third* copy of the same predicate: `STDROT_DYNAMIC_MODULES` here, `MODULE_HAS_NATIVE` in `lib/module_path.c`, and `COOKED_NATIVE_EXT`'s guard in `lang.l` — all three literally `#if !defined(STDROT_STATIC) || defined(_WIN32)`, and two of them additionally re-derive the `.dll`/`.so` suffix. You've documented the "must match stdrot.c's predicate exactly" contract in each comment, which is the right mitigation given they're in separate TUs — but a comment is not a compiler. The day someone adds a fourth platform axis (a `STDROT_STATIC` build that *does* want a loader, say), these drift silently and the failure mode is ugly: `module_path.c` resolves a `<name>.so` that `stdrot.c` then refuses to load, or advertises the wrong suffix in the not-found diagnostic. Consider hoisting the predicate + suffix into a single shared header (`stdrot_api.h` already crosses all three) so there's one definition, not three that a reviewer has to eyeball for equality. Non-blocking.

leo-aa88 (commented):
LGTM. Clean decoupling of "core is static" from "modules can load" — the refactor is the right shape and I verified it end-to-end rather than trusting the writeup.

**Verified across all three build configs:**
- **POSIX: `make test` → 501 passed.** The shared `stdrot_load_module`/`stdrot_unload`/runtime-query paths are exercised unchanged by the full native-module suite; no regression from re-keying them on `STDROT_DYNAMIC_MODULES`.
- **wasm config compiles clean** (`gcc -DSTDROT_STATIC -Werror -fsyntax-only stdrot.c`, exit 0) — the `#else` stub `stdrot_load_module` is reached and every `cooked_modules[]` reference is correctly compiled out, so the DYNAMIC_MODULES-undefined path has no dangling symbol.
- **Windows: the actual crux is proven statically.** Cross-built `make windows` + `make windows-test-module` → `brainrot.exe` (PE32+, self-contained: imports only `bcrypt`/`KERNEL32`/UCRT, `LoadLibraryA`/`GetProcAddress` come from the already-linked `KERNEL32`) and `testnative.dll` (PE32+ DLL). `objdump -p` on the DLL confirms it exports **exactly `brainrot_module_init_v3`** — the name the loader `GetProcAddress`es — and `objdump -h` shows the **`stdrot_exports` linker section is present in the PE**, which is what makes `registry.c`'s `__start_/__stop_` collection work on MinGW/PE. Everything the CI `tripled(14) → 42` job depends on has its prerequisites satisfied; the only thing left is the runtime call, which is exactly what that job covers (no Wine here).

**Details that were easy to get wrong and aren't:**
- The preprocessor nesting is balanced and, more importantly, *correctly split*: core dynamic state (`lib_handle`, `symbol_cache`, `cache_count`) stays under `#ifndef STDROT_STATIC`, cooked-module state under `#ifdef STDROT_DYNAMIC_MODULES`, and `stdrot_unload()` uses **two independent guards** so the Windows-static build frees modules without ever touching `cache_count`.
- `br_module_error()` reads `GetLastError()` as the first thing after a failed `br_module_open()`, with no intervening Win32 call to clobber the thread-error — so the diagnostic reports the real load failure, not a stale code.
- The Stage-1→2 fix (`19d7b27`) switching `is_builtin_function`/`get_native_function` to `STDROT_DYNAMIC_MODULES` is load-bearing and correct: without it a Windows-loaded module would register but never resolve at call time. Good catch.

One inline maintainability nit below (the predicate now lives in three files). Non-blocking.

(Submitting as a comment, not an Approve — GitHub blocks approving your own PR.)

leo-aa88 (commented):
Re-reviewed against HEAD (`be5d6fc`). The single new commit does exactly what the prior inline nit asked — hoists the predicate and suffix into `lib/module_path.h` as `MODULE_NATIVE_LOADER`/`MODULE_NATIVE_SUFFIX`, one definition, and points all three TUs (`stdrot.c`, `lib/module_path.c`, `lang.l`) at it. Nit resolved.

I re-ran the full matrix against this HEAD, since a shared macro that resolves differently per TU would be worse than the duplication it replaces:

- **The include-order footgun is genuinely closed.** I verified all three TUs `#include "lib/module_path.h"` (`stdrot.c:33`, `lang.l:17`, `module_path.c:11`), and — the load-bearing part — `STDROT_STATIC` and `_WIN32` are both `-D`/compiler-predefined, so the `#if` in the header resolves *before any include* and identically in every TU regardless of include order. The header comment calls this out explicitly, and it's correct. Concretely: had `lang.l` not pulled in the header, `#ifndef MODULE_NATIVE_LOADER` would silently flip its "cannot find module" diagnostic to the wasm-only ".brainrot only" wording on native/Windows — it does include it, so that doesn't happen.
- **POSIX:** `make test` → **501 passed**, unchanged.
- **wasm config:** `gcc -DSTDROT_STATIC -Werror -fsyntax-only stdrot.c` → clean; loader blocks still compile out, stub reached.
- **Windows:** `make windows` + `make windows-test-module` cross-compile; `brainrot.exe` still self-contained (no `libgcc`/`libwinpthread`/`libstdc++`); `testnative.dll` still exports exactly `brainrot_module_init_v3` and carries the `stdrot_exports` PE section.
- **`make format-check`** passes (the CI `lint` gate), including the new `module_path.h` block.

Pure refactor, no behavior change, no regression. Same as before: LGTM, submitting as a comment because GitHub won't let you Approve your own PR.

PR #340, opened 2026-08-31
feat: ship the Windows artifact in releases (issue #337, Stage 3)

### LINKED ISSUE #337: [Feature]: Native Windows build support

leo-aa88:
## Description

**Stage 3 of native Windows support (#337): the Windows artifact ships in releases.** This completes #337 — Stage 1 (core interpreter), Stage 2 (native modules), Stage 3 (release).

`release.yml` previously omitted Windows. This adds it:

- **`windows` job** (MSYS2/MinGW-w64, `needs: version`) — builds the self-contained `brainrot.exe` via `make windows`, **verifies it imports only system DLLs** (no `libgcc_s_seh-1.dll`/`libwinpthread-1.dll` — the Windows analogue of the Darwin `otool` check in `native-release-build`), smoke-runs `hello_world`, and uploads it as `brainrot-windows-amd64`.
- **`publish`** now `needs` it and packages it as **`brainrot-<tag>-windows-amd64.zip`** — a `.zip` (the conventional Windows archive) carrying just `brainrot.exe` (no `libstdrot.so`; the core is statically linked in). Built with `python3 -m zipfile`, which — like `tar` — is always present on the runner, rather than the `zip` binary which isn't guaranteed. The zip flows into `SHA256SUMS.txt` and the release assets automatically.
- **`README.md`** — the prebuilt-binaries list documents the Windows zip; the "Windows is not a native target" note is gone.

Windows is a separate job (not the `build` matrix) for the same reason it's a separate CI job: it can't use the `native-release-build` composite action (`make release` → `brainrot` + `libstdrot.so`); it's one static `brainrot.exe` from `make windows` under MSYS2, packaged as a zip.

## Related Issue

Closes #337 (Stages 1–2 already merged: #338, #339).

## Type of Change

## Checklist

- [x] I have added tests that prove my change works, at the lowest appropriate layer — this is release *packaging*, so the proof is the `windows` job itself: it builds the shipped `brainrot.exe`, asserts self-containment (only system-DLL imports), and runs it. `make windows` is the same target already exercised end-to-end (including native modules) by the merged CI `windows` job.
- [x] If this PR cannot affect program behavior … — it's CI/release + docs only; it changes no interpreter code, so the interpreter test suite is unaffected. (Confirmed: no `.c`/`.l`/`.y`/`Makefile` build-logic changes — only `release.yml` and `README.md`.)
- [x] I have run `make format-check` locally — N/A to the changed files, but unchanged and still passing on `main`.
- [ ] I have run the unit tests locally — no interpreter code changed; the release `version` job gates packaging on `make test` as before.
- [ ] I have run the valgrind memory tests locally — N/A (no C changes).

### Validation done locally

`release.yml` is valid YAML. I simulated the publish `Package archives` Windows branch end-to-end against a MinGW-cross-built `brainrot.exe`: `python3 -m zipfile -c` produces `brainrot-<tag>-windows-amd64.zip` containing exactly `brainrot.exe`, and it flows through the existing `sha256sum` step. The `windows` job mirrors the already-green CI `windows` build (`make windows` + the self-containment `objdump` check). A `workflow_dispatch` **dry run** (`dry_run: true`) is the right first exercise of the full release path before cutting a real tag.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -263,10 +263,19 @@ jobs:
           ref: ${{ needs.version.outputs.checkout_ref }}
 
       - name: Set up MSYS2 (MinGW-w64 toolchain)
-        uses: msys2/setup-msys2@v2
+        # SHA-pinned (v2.32.0): this builds a PUBLISHED artifact
+        # (brainrot-<tag>-windows-amd64.zip), so the same reasoning that
+        # SHA-pins setup-emsdk and action-gh-release applies -- a retag
+        # upstream must not silently change the shipped .exe. `update: true`
+        # is deliberately dropped for the same reason: it would pull whatever
+        # MSYS2 serves that day, reintroducing the nondeterminism the pin
+        # removes. The MinGW package version is left to the pinned action's
+        # bundled MSYS2 snapshot rather than pinned explicitly -- MSYS2 is a
+        # rolling distro that garbage-collects old package versions, so an
+        # exact package pin would eventually fail to resolve.
+        uses: msys2/setup-msys2@66cd2cce69caa17b53920067426061ca1de3a884 # v2.32.0
         with:
           msystem: MINGW64
-          update: true
           install: >-
             mingw-w64-x86_64-gcc
             make
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -7,8 +7,14 @@
 # Native binaries are built on NATIVE runners (linux/amd64, linux/arm64,
 # darwin/amd64, darwin/arm64) because the interpreter is C + Flex/Bison and
 # dlopen's libstdrot.so at runtime — there is no useful cross-compile path
-# that still produces a runnable artifact. Windows is omitted: the loader
-# and stdrot are POSIX (dlfcn.h, libstdrot.so, unistd.h).
+# that still produces a runnable artifact.
+#
+# Windows (windows/amd64) is a separate job too: it ships a SINGLE
+# self-contained brainrot.exe from `make windows` (STDROT_STATIC, the core
+# compiled in — no libstdrot.so — and -static so the MinGW runtime is baked
+# in), built under MSYS2. Native #cooked <name> modules load via LoadLibraryA
+# (issue #337). It's packaged as a .zip, not a .tar.gz, and carries no
+# libstdrot.so.
 #
 # wasm stays a separate job (Emscripten, statically linked stdrot) and still
 # publishes brainrot.wasm / brainrot.mjs under those exact names so existing
@@ -245,9 +251,63 @@ jobs:
           if-no-files-found: error
           retention-days: 1
 
+  windows:
+    name: Build windows/amd64
+    needs: version
+    runs-on: windows-latest
+    timeout-minutes: 30
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v7
+        with:
+          ref: ${{ needs.version.outputs.checkout_ref }}
+
+      - name: Set up MSYS2 (MinGW-w64 toolchain)
+        uses: msys2/setup-msys2@v2
+        with:
+          msystem: MINGW64
+          update: true
+          install: >-
+            mingw-w64-x86_64-gcc
+            make
+            bison
+            flex
+
+      - name: Build brainrot.exe
+        shell: msys2 {0}
+        run: make windows
+
+      # The shipped .exe must import only system DLLs. A libgcc_s_seh-1.dll /
+      # libwinpthread-1.dll dependency would make it run only where MSYS2's bin
+      # is on PATH -- the Windows analogue of the Darwin otool check in the
+      # native-release-build action. -static (Makefile) prevents it; fail
+      # loudly here if that ever regresses. Also smoke-run it.
+      - name: Verify + smoke-test
+        shell: msys2 {0}
+        run: |
+          set -euo pipefail
+          objdump -p brainrot.exe | grep -i 'DLL Name'
+          if objdump -p brainrot.exe | grep -iqE 'libgcc|libwinpthread|libstdc\+\+'; then
+            echo "ERROR: brainrot.exe depends on a MinGW runtime DLL;" \
+              "it must be statically linked (-static)." >&2
+            exit 1
+          fi
+          out="$(./brainrot.exe examples/hello_world.brainrot)"
+          test "$out" = "Hello, World!" \
+            || { echo "hello_world mismatch: [$out]"; exit 1; }
+          echo "windows release artifact verified"
+
+      - name: Upload brainrot.exe
+        uses: actions/upload-artifact@v7
+        with:
+          name: brainrot-windows-amd64
+          path: brainrot.exe
+          if-no-files-found: error
+          retention-days: 1
+
   publish:
     name: Package and publish release
-    needs: [version, build, wasm]
+    needs: [version, build, wasm, windows]
     runs-on: ubuntu-latest
     timeout-minutes: 15
     steps:
@@ -280,6 +340,15 @@ jobs:
               tar -czf "$out/brainrot-${VERSION}-wasm.tar.gz" \
                 -C "$dir" brainrot.wasm brainrot.mjs
               cp "$dir/brainrot.wasm" "$dir/brainrot.mjs" "$out/"
+              continue
+            fi
+            # Windows ships a single self-contained brainrot.exe (no
+            # libstdrot.so) as a .zip, the conventional Windows archive.
+            # python3 -m zipfile (always present, like tar) rather than the
+            # `zip` binary, which isn't guaranteed on the runner.
+            if [ "$name" = "brainrot-windows-amd64" ]; then
+              (cd "$dir" && python3 -m zipfile -c \
+                "$out/brainrot-${VERSION}-windows-amd64.zip" brainrot.exe)
               continue
             fi
             platform="${name#brainrot-}"
@@ -336,6 +405,7 @@ jobs:
           prerelease: ${{ needs.version.outputs.prerelease == 'true' }}
           files: |
             dist/*.tar.gz
+            dist/*.zip
             dist/brainrot.wasm
             dist/brainrot.mjs
             dist/SHA256SUMS.txt
--- a/README.md
+++ b/README.md
@@ -161,13 +161,18 @@ Each `v*` GitHub release attaches native archives plus the wasm module:
   linker's `libstdrot.so` search path. Release binaries include an rpath
   (`$ORIGIN` / `@loader_path`) so that second lookup can find the
   `libstdrot.so` next to `brainrot` when no cwd copy is loaded first.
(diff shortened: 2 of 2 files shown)


.github/workflows/release.yml
```diff
@@ -245,9 +251,63 @@ jobs:
...
+    runs-on: windows-latest
+    timeout-minutes: 30
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v7
+        with:
+          ref: ${{ needs.version.outputs.checkout_ref }}
+
+      - name: Set up MSYS2 (MinGW-w64 toolchain)
+        uses: msys2/setup-msys2@v2
```
leo-aa88: **MINOR (supply-chain consistency).** This job builds a **published release artifact** (the `brainrot.exe` that ships in `brainrot-<tag>-windows-amd64.zip`), but its builder is pinned more loosely than the file's own stated bar. Twelve lines up, the `wasm` job SHA-pins `emscripten-core/setup-emsdk` with the explicit rationale *"so a retagged release can't silently change what builds the artifact this workflow publishes"* — and `action-gh-release` is SHA-pinned too. Here the toolchain is `msys2/setup-msys2@v2` (a moving major tag) **plus** `update: true` and a rolling `mingw-w64-x86_64-gcc` package, so the exact compiler and action that produce a shipped binary are whatever MSYS2 serves that day. By the very reasoning the emsdk pin cites, a released Windows `.exe` deserves at least as reproducible a builder: pin `setup-msys2` to a SHA (and ideally drop `update: true` / pin the mingw package version) so a retag upstream can't change the published artifact silently. Non-blocking — `msys2/setup-msys2` is a reputable action and the rest of the file uses moving tags for `checkout`/`upload-artifact` — but it's the one place this PR touches where the publish-time trust bar the file sets for itself isn't met.

leo-aa88 (commented):
LGTM. Purely additive release wiring for the Windows artifact, and it reuses the `make windows` + self-containment + smoke pattern already merged and green in `ci.yml`. I validated the parts that could actually break a release rather than trusting the writeup:

- **YAML is valid** (`yaml.safe_load` clean).
- **The zip packaging is correct.** Simulated the `publish` Windows branch exactly (`cd "$dir" && python3 -m zipfile -c "$out/…zip" brainrot.exe`): the archive contains exactly `brainrot.exe` at the root (no stray directory prefix), and it flows into `SHA256SUMS.txt` — the `sha256sum ./*` glob expands before the `tee` creates that file, so the sums file correctly isn't hashing itself while the new `.zip` is. `dist/*.zip` is added to the release `files:` list, so it actually attaches.
- **The smoke assertion matches reality.** `hello_world.brainrot` emits `Hello, World!\n`; `$(...)` strips the trailing newline, so `test "$out" = "Hello, World!"` holds — and this is the *same* assertion already running in the merged `ci.yml` windows job (line 447), i.e. a pattern proven on a real `windows-latest` runner, not a new one being tried for the first time here.
- **Job wiring is right:** `windows` `needs: version` and checks out `checkout_ref` like `build`/`wasm`; `publish` now `needs` it, so a release can't be cut without the Windows artifact; naming (`brainrot-<tag>-windows-amd64.zip`) matches the tarball scheme.

One inline MINOR note below on builder pinning — the only spot where this PR's publish-time trust posture is looser than the file's own stated standard. Non-blocking.

Same process caveat: submitting as a comment because GitHub blocks approving your own PR. Substance is an approval — this closes #337 cleanly.

leo-aa88 (commented):
Re-reviewed against HEAD (`c7282b0`). The one new commit resolves the prior inline note exactly and with the right nuance.

- **The pin is genuine.** I checked against the upstream repo: `msys2/setup-msys2`'s `v2.32.0` is a lightweight tag pointing directly at commit `66cd2cce69caa17b53920067426061ca1de3a884` — the exact SHA now in the workflow. So the published Windows `.exe` is now built by a fixed action, matching the trust bar the file already sets for `setup-emsdk` and `action-gh-release`.
- **Dropping `update: true` is correct, not a regression.** It only skipped the full `pacman -Syu`; the `install:` list (`mingw-w64-x86_64-gcc make bison flex`) still installs from the pinned action's bundled MSYS2 snapshot, which is *more* deterministic, not less — exactly the point.
- **The decision to NOT pin the mingw package version is the right call and well-justified in the comment:** MSYS2 is a rolling distro that garbage-collects old package versions, so an exact `=version` pin would eventually fail to resolve. Pinning the action (and thus its bundled repo snapshot) is the reproducibility you can actually sustain; a package-version pin would rot. Good judgment.
- **YAML re-validated; nothing else changed** — the zip packaging, checksum flow, smoke assertion, job wiring, and README from my prior pass are untouched and still hold.

Nit resolved, no new issues. LGTM — closes #337. Submitting as a comment again since GitHub blocks self-approval.

PR #344, opened 2026-09-02
rayrot: window controls + render-texture framebuffer scaling

leo-aa88:
## Description

Nine additive functions in the hand-written `rayrot` binding (`rayrot/raylib.c`), for desktop-release polish in downstream games — a pause menu (an ESC that doesn't kill the process) and scaling a fixed logical resolution to any window / fullscreen:

| Function | raylib | Purpose |
|---|---|---|
| `rl_set_exit_key(key)` | `SetExitKey` | Disable ESC-as-quit (`0` = `KEY_NULL`) so a game can make ESC a pause |
| `rl_set_config_flags(flags)` | `SetConfigFlags` | e.g. `FLAG_WINDOW_RESIZABLE` before `InitWindow` |
| `rl_toggle_borderless_windowed()` | `ToggleBorderlessWindowed` | Friendlier fullscreen, no mode switch |
| `rl_is_window_resized()` | `IsWindowResized` | → bool |
| `rl_load_render_texture(w, h)` | `LoadRenderTexture` | → handle, or `-1` |
| `rl_begin_texture_mode(h)` | `BeginTextureMode` | Draw the scene into the framebuffer |
| `rl_end_texture_mode()` | `EndTextureMode` | |
| `rl_draw_render_texture_fit(h)` | `DrawTexturePro` | Blit letterboxed to the live window size |
| `rl_unload_render_texture(h)` | `UnloadRenderTexture` | |

Render textures use the same Road A integer-handle table as textures (a `RenderTexture2D` is a struct that can't cross the ABI, so C owns the array; the generated Road B binding can't return one, same as `LoadTexture`). `rl_close_window()` now unloads every still-owned render texture before destroying the GL context, mirroring the existing texture cleanup. `rl_draw_render_texture_fit` is a composite that keeps the scale/letterbox maths in C (reads `GetScreenWidth/Height`, blits with `DrawTexturePro`, source height negated for the render-texture y-flip) rather than exposing a 16-argument `DrawTexturePro` across the boundary.

Additive only — no existing signature changes. First consumer: `Brainrotlang/tung-tung-sahur` (ESC pause menu + 1280×720 framebuffer scaled to the window).

**On testing:** these are thin wrappers over raylib window/GL calls, which require a live GL context (a window + display). The `make test` / `make valgrind` harness deliberately does not build `rayrot` and CI has no display, so — exactly like every existing `rayrot` wrapper (`rl_load_texture`, `rl_begin_drawing`, …) — they are not reachable by a `test_cases/*.brainrot` fixture or a host-side C test. They are verified by: a clean `make rayrot` build under `-Wall -Wextra` (warnings are errors), `make format-check`, and the downstream game's `make lint-native` (which type-checks every `rl_*` call against this module).

## Related Issue

N/A — desktop-release polish for the downstream `tung-tung-sahur` game; no tracking issue.

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement
- [ ] Refactor

## Checklist

- [ ] I have added tests — **not reachable by the harness**: these wrap raylib window/GL calls that need a live GL context, which `make test` (no `rayrot` build, no display) cannot provide, same as every existing `rayrot` wrapper. Verified instead by the `-Wall -Wextra` build, `make format-check`, and the downstream `make lint-native`. See the testing note above.
- [ ] Adversarial/edge cases — the handle-table wrappers validate range + liveness (out-of-range / unloaded handle is a no-op), mirroring the existing texture/audio wrappers.
- [x] If this PR cannot affect program behavior via the test harness, I explained that in the Description
- [x] I have run `make format-check` locally
- [ ] I have run the unit tests locally — `make test` does not build `rayrot`, so it is unaffected by this change (CI runs it regardless)
- [ ] I have run the valgrind memory tests locally — same: `make valgrind` does not build `rayrot`
- [x] All new and existing tests pass (CI)

### DIFF
--- a/docs/rayrot.md
+++ b/docs/rayrot.md
@@ -371,6 +371,10 @@ Colors are always the trailing `r, g, b, a` integers (0–255, clamped).
 | `rl_set_target_fps(fps)` | `SetTargetFPS` | |
 | `rl_get_screen_width()` | `GetScreenWidth` | returns `rizz` |
 | `rl_get_screen_height()` | `GetScreenHeight` | returns `rizz` |
+| `rl_set_config_flags(flags)` | `SetConfigFlags` | before `rl_init_window`; e.g. `4` = `FLAG_WINDOW_RESIZABLE` |
+| `rl_set_exit_key(key)` | `SetExitKey` | `0` (`KEY_NULL`) stops ESC closing the window |
+| `rl_toggle_borderless_windowed()` | `ToggleBorderlessWindowed` | borderless fullscreen, no mode switch |
+| `rl_is_window_resized()` | `IsWindowResized` | returns `cap` |
 | `rl_begin_drawing()` | `BeginDrawing` | |
 | `rl_end_drawing()` | `EndDrawing` | |
 | `rl_clear_background(r, g, b, a)` | `ClearBackground` | |
@@ -389,6 +393,11 @@ Colors are always the trailing `r, g, b, a` integers (0–255, clamped).
 | `rl_draw_texture(handle, x, y, r, g, b, a)` | `DrawTexture` | `r,g,b,a` = tint |
 | `rl_draw_texture_rec(handle, sx, sy, sw, sh, x, y, r, g, b, a)` | `DrawTextureRec` | one sub-rectangle; see below |
 | `rl_unload_texture(handle)` | `UnloadTexture` | |
+| `rl_load_render_texture(width, height)` | `LoadRenderTexture` | offscreen framebuffer; returns handle or `-1` |
+| `rl_begin_texture_mode(handle)` | `BeginTextureMode` | draw the scene into the framebuffer |
+| `rl_end_texture_mode()` | `EndTextureMode` | |
+| `rl_draw_render_texture_fit(handle)` | `DrawTexturePro` | present the framebuffer scaled-to-fit and letterboxed; between begin/end drawing |
+| `rl_unload_render_texture(handle)` | `UnloadRenderTexture` | |
 | `rl_init_audio_device()` | `InitAudioDevice` | see below |
 | `rl_is_audio_device_ready()` | `IsAudioDeviceReady` | returns `cap` |
 | `rl_close_audio_device()` | `CloseAudioDevice` | unloads every live stream first |
--- a/rayrot/raylib.c
+++ b/rayrot/raylib.c
@@ -66,6 +66,19 @@
 static Texture2D g_textures[RAYROT_MAX_TEXTURES];
 static bool g_texture_used[RAYROT_MAX_TEXTURES];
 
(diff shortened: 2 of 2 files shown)


leo-aa88 (commented):
LGTM. Nine additive Road A bindings, and they're a faithful copy of the machinery that's already proven — I read them against the existing texture path and validated the whole thing against a real raylib rather than trusting "built clean."

**Verified:**
- **Builds clean.** `make rayrot` against raylib 6.0.0 with `-Wall -Wextra` → zero warnings, every raylib symbol resolves (`SetExitKey`, `SetConfigFlags`, `ToggleBorderlessWindowed`, `IsWindowResized`, `LoadRenderTexture`, `Begin/EndTextureMode`, `DrawTexturePro`, `UnloadRenderTexture`).
- **All 9 register**, and the module loads: I cooked `<raylib>` and called them from a Brainrot program — the semantic analyzer accepts the arities, and dispatch works end-to-end.
- **The failure path is real, not theoretical.** Running without a GL context, `rl_load_render_texture(1280,720)` hit raylib's "FBO cannot be created", so `rt.id == 0` → the handler returned **-1** with no slot consumed, and the program exited 0. That's exactly the "-1 without consuming a slot, so a live handle is always a real target" contract, exercised. A subsequent `-1` handle into begin/draw/unload just trips the `h < 0` guard and no-ops — safe.
- **The handle table mirrors `br_load_texture`/`br_unload_texture` exactly** — `.id == 0` → -1, linear scan for a free slot, `UnloadRenderTexture` + -1 when full (no leak), and every consumer bounds-checks `h >= 0 && h < MAX && g_render_texture_used[h]`. No OOB, no double-free.
- **`rl_close_window` now unloads render textures before `CloseWindow()`**, in the same shape as the texture cleanup right above it — GL resources freed before the context dies. Good.
- **The composite `rl_draw_render_texture_fit` is the canonical letterbox idiom** — `scale = min(sw/fw, sh/fh)`, centred dst, `src.height` negated for the render-texture y-flip. Keeping that maths in C instead of exposing a 16-arg `DrawTexturePro` across the ABI is the right call, consistent with why these live in the hand-written binding at all.
- **Every `STDROT_EXPORT_SIG` matches its handler** — arity, param descriptors (`p_win_int1`/`p_render_wh`), and return kinds (`R_NONE`/`R_INT`/`R_BOOL`) all line up.

No test fixture, but that's consistent with the rest of `rayrot/raylib.c`: it's GL/display-dependent and excluded from `make test`/`valgrind` by design (the `test_rayrot_gen.py` suite covers the *generated* Road B binding, not this hand-written one). Nothing here is testable in headless CI beyond the load/dispatch/failure-path check I ran, which passes.

Additive, no signature changes, mirrors a proven pattern, builds clean, handles its own failure modes. Submitting as a comment since GitHub blocks self-approval; substance is an approval.

PR #347, opened 2026-09-06
test: guard #180 UBSan misaligned array-access load with a regression fixture

### LINKED ISSUE #180: UBSan: misaligned load in evaluate_expression_short/int NODE_ARRAY_ACCESS
## What happens

`examples/sieve_of_eras.brainrot` trips two UBSan misaligned-load errors on the native (ASan/UBSan) build. Both come from `evaluate_expression_*` reinterpreting a `void *` returned by `evaluate_multi_array_access` as a wider type and dereferencing it directly:

```
ast.c:2238:16: runtime error: load of misaligned address 0x... for type 'short int', which requires 2 byte alignment
ast.c:2375:16: runtime error: load of misaligned address 0x... for type 'int', which requires 4 byte alignment
```

```c
// ast.c:2238, inside evaluate_expression_short's NODE_ARRAY_ACCESS case
return *(short*)evaluate_multi_array_access(node);

// ast.c:2375, same shape inside evaluate_expression_int
return *(int*)evaluate_multi_array_access(node);
```

Repro: `make && ./brainrot examples/sieve_of_eras.brainrot` (native build, which carries `-fsanitize=undefined`). Output is still correct — x86-64 tolerates misaligned loads — but it's undefined behavior per the C standard and would misbehave or fault outright on a strict-alignment target.

## How I found this

Not part of the wasm work in #178 directly — found while adding `tests/run_wasm_examples_check.mjs`, which runs every `examples/*.brainrot` file (not just the ones already covered by `test_cases/`) to verify wasm output matches native. `sieve_of_eras.brainrot` isn't in `test_cases/`, so nothing in the existing `make test`/`make valgrind` suite ever exercised this path — Valgrind alone doesn't catch misaligned-load UB the way UBSan does, and UBSan only runs as part of the normal `./brainrot` binary, which nothing was pointing at this specific example.

## Why I'm not fixing it here

Root-causing this means understanding `evaluate_multi_array_access`'s actual memory layout (why would an offset into that buffer ever land on an odd/non-4-aligned address? packed array-of-smaller-type storage? pointer arithmetic in bytes without accounting for the target element's alignment?) well enough to fix it without changing array semantics elsewhere. That's real interpreter-internals work, unrelated to #178's build-target scope.

## Suggested next step

Worth adding `sieve_of_eras` (and ideally the rest of `examples/`) to `test_cases/`/`tests/expected_results.json` so `make valgrind`/UBSan actually exercise every example the README points people at — right now there's a real coverage gap between "programs documented as working" and "programs actually tested."

leo-aa88:
## Description

Issue #180 reported UBSan **misaligned-load** errors in
`evaluate_expression_short`/`evaluate_expression_int`'s `NODE_ARRAY_ACCESS`
case, reproduced by `examples/sieve_of_eras.brainrot`: a `cap` (bool, 1-byte)
array element sitting at an odd address was read through a `*(int*)` / `*(short*)`
cast — undefined behavior on strict-alignment targets even though x86-64
tolerates it.

**The load itself is already fixed.** The scalar array-access paths now go
through the `numeric_load()` helper, which dispatches on the element's `VarType`
(`VAR_BOOL` → a 1-byte read) instead of reinterpreting the byte address as a
wider type. Building native (`-fsanitize=address,undefined`) and running the
example produces **no** UBSan errors.

**What was still missing — and what the issue's "Suggested next step" explicitly
asked for — is coverage.** `sieve_of_eras.brainrot` lived only in `examples/`,
so nothing in `make test` / `make valgrind` ever exercised this path; a
regression could silently reintroduce the UB. This PR adds it as a `test_cases/`
fixture (byte-identical to the example, following the existing
`hello_world`/`fizz_buzz` convention of mirroring an example into `test_cases/`)
with its expected output, so the sanitized `make test` binary now runs the exact
reproducer on every build.

## Related Issue

Fixes #180

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement
- [ ] Refactor

## Checklist

> The fixture is the original UBSan reproducer, so it exercises exactly the
> misaligned-load path that regressed. `make test` (505 passed) and Valgrind on
> the new fixture are clean, and the sanitized binary reports no UBSan errors on
> it. No C source changed, so `cppcheck`/`clang-format` are unaffected.

### DIFF
--- a/test_cases/sieve_of_eras.brainrot
+++ b/test_cases/sieve_of_eras.brainrot
@@ -1,0 +1,22 @@
+skibidi main {
+        rizz i;
+        rizz p;
+        cap prime[105];
+        flex(i = 1; i <= 100; i = i + 1){
+                prime[i] = W;
+        }
+
+        flex(p = 2; p * p <= 100; p++){
+                edgy(prime[p] == W){
+                        flex(i = p * p; i <= 100; i = i+p){
+                                prime[i] = L;
+                        }
+                }
+        }
+
+        flex(p = 2;p <= 100; p = p + 1){
+                edgy(prime[p]) { yapping("%d", p); }
+        }
+
+        bussin 0;
+}
--- a/tests/expected_results.json
+++ b/tests/expected_results.json
@@ -427,5 +427,6 @@
     "file_io_use_after_close_fail": "opened\nclose 0\nclosed\nStderr:\nError: yapto: not an open SAUCE -- it was already closed with peaceout, or was never a handle at all, at line 47\n",
     "file_io_ragequit_no_leak": "ExitCode:4",
     "file_io_stale_handle_fail": "close a 0\nsame handle 0\nb works, pos 5\nnow using the retired handle, while b is still open:\nStderr:\nError: yapto: not an open SAUCE -- it was already closed with peaceout, or was never a handle at all, at line 62\n",
-    "freed_keywords_as_identifiers": "functions 42 42\nfields    2 3\nlocals    5 7\n"
+    "freed_keywords_as_identifiers": "functions 42 42\nfields    2 3\nlocals    5 7\n",
+    "sieve_of_eras": "2\n3\n5\n7\n11\n13\n17\n19\n23\n29\n31\n37\n41\n43\n47\n53\n59\n61\n67\n71\n73\n79\n83\n89\n97"
 }


leo-aa88 (commented):
# APPROVE

**LGTM.** Adversarial pass over this test-only PR; I checked out the head, built, and ran the fixture end-to-end (sanitized + valgrind). Every claim in the description holds.

## Findings

Traced and verified:
- **CONFIRMED** `test_cases/sieve_of_eras.brainrot` is byte-identical to `examples/sieve_of_eras.brainrot` (`diff` clean), honoring the existing "mirror an example into `test_cases/`" convention.
- **CONFIRMED** `tests/expected_results.json` remains valid JSON; the `sieve_of_eras` entry is the correct output — the 25 primes ≤ 100 (2…97).
- **CONFIRMED (regression guard works)** the sanitized binary (`-fsanitize=address,undefined`) runs the fixture with exit 0 and empty stderr — no misaligned-load UBSan error. This is exactly the #180 path (`cap`/1-byte element read via a wider cast) now being exercised by `make test` on every build, which the example-only location never did.
- **CONFIRMED (test passes)** `pytest -k sieve_of_eras` → 1 passed.
- **CONFIRMED (valgrind clean)** under the project's exact gate (`--leak-check=full --track-fds=yes --error-exitcode=100`, unsanitized `brainrot-valgrind`): exit 0, ERROR SUMMARY 0 errors, definitely/possibly lost 0 bytes, only the 3 std FDs open.

## QUESTION (resolved, not a finding)

The expected value ends at `97` with no trailing newline, while the interpreter emits a trailing `\n`. Not a defect: the harness `.strip()`s **both** `actual_output` and `expected_output` before comparing (`tests/test_brainrot.py:59,64`), so the comparison is unaffected. Consistent with neighboring entries either way.

## VERDICT

Does exactly what it claims: converts the #180 UBSan reproducer from an unexercised example into a guarded regression fixture, with correct expected output and clean sanitizer + valgrind runs. No C source touched, so cppcheck/clang-format are unaffected. Deserves to merge. (Posting as COMMENT rather than a formal approval only because the reviewer account is also the PR author — GitHub rejects self-approval.)

PR #350, opened 2026-09-06
fix: error on struct/union tag redefinition instead of silently overwriting (#189)

### LINKED ISSUE #189: struct/union tag redefinition silently overwrites instead of erroring
## Summary

The docs (`docs/the-brainrot-programming-language.md` §7.10) state that `gang` and `chungus` tags share a namespace, so a `gang Foo` and a `chungus Foo` can't coexist — matching real C's tag-namespace rule. In practice, `register_struct_def` (`ast.c`) just overwrites the hashmap entry on a name clash, with no diagnostic:

```
gang Foo { rizz a; };
chungus Foo { chad f; };

skibidi main {
    gang Foo g;
    g.a = 1;   🚽 "Foo" now resolves to the union definition
}
```

This produces `Error: Undefined variable at line 1` rather than a clear redefinition error, and the line number is misleading regardless (see the related, now-fixed, struct-access line-number issue in PR #188 — this is a separate case since the *declaration* checking, not member access, is what's silently wrong here).

Same behavior already existed for two `gang Foo` definitions before `chungus` was added — this isn't union-specific, just newly re-surfaced by giving `gang` a second keyword (`chungus`) that shares its namespace.

## Suggested direction

Either:
- Raise a semantic/parse error on a duplicate tag (`register_struct_def` or the `struct_def` grammar action in `lang.y` could check `get_struct_def(name)` before overwriting), or
- If "last definition wins" is the intended, simpler behavior, soften the docs to say so explicitly instead of claiming C's stricter rule.

Flagged during review of PR #188 (chungus/union implementation) — not fixed there to keep that PR's scope to the union feature itself.

leo-aa88:
## Description

`gang` and `chungus` share one tag namespace — the language reference
(`docs/the-brainrot-programming-language.md` §7.10) says *"Struct and union tags
share the same namespace, so a `gang` and a `chungus` cannot use the same
name."* But `register_struct_def()` (`ast.c`) just `hm_put`'d over the existing
entry on a name clash, with **no diagnostic**. The issue's repro:

```
gang Foo { rizz a; };
chungus Foo { chad f; };

skibidi main {
    gang Foo g;
    g.a = 1;   // "Foo" had silently become the union
}
```

...failed with a misleading `Error: Undefined variable` instead of a clear
redefinition error. The same silent-overwrite existed for two `gang Foo`
definitions even before `chungus` was added.

### Fix

Guard the registration in `register_struct_def()` — the single choke point both
definition paths flow through:

- If the tag already exists, emit `Redefinition of tag '<name>' -- gang and
  chungus share one tag namespace` via `yyerror_current_line()` and set
  `struct_def_had_error` (so `main` fails the parse before semantic
  analysis/execution, matching the other parse-phase rejections).
- Keep the **first** definition in the lookup table (don't overwrite).
- Still thread the rejected `def` onto `struct_registry_list` so
  `free_struct_registry()` reclaims it — **no leak** (verified under valgrind).

The `lit` alias path in `lang.y` already guarded duplicates before constructing
a def; this closes the gap for the plain `gang Foo {...}` / `chungus Foo {...}`
grammar action. No doc change needed — this makes the implementation match the
already-documented contract.

### Behavior change note

This turns a previously-silent overwrite into a hard error. Any program that
relied on "last definition wins" was relying on behavior the docs already
forbade; such a program now gets a clear diagnostic at parse time.

## Related Issue

Fixes #189

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement
- [ ] Refactor

## Checklist

### Tests added

- `duplicate_tag_gang_chungus_fail` — the issue's exact repro (`gang` then `chungus`, same tag) → redefinition error.
- `duplicate_tag_gang_gang_fail` — two `gang Foo` (the pre-`chungus` form of the same bug) → redefinition error.
- `gang_chungus_distinct_tags` — distinct tags (`Point`/`Num`) still coexist and are usable → guards against over-rejection.

`make test` → **508 passed**; full `make valgrind` sweep → exit 0 (the error path is leak-clean). `make format-check` clean. `make cppcheck` couldn't run locally (this box has cppcheck 2.7 < the required 2.13); the change is a `snprintf` + existing-tag lookup, left for CI's `static-analysis` job.

### DIFF
--- a/ast.c
+++ b/ast.c
@@ -8586,6 +8586,27 @@ void register_struct_def(StructDef *def)
     }
     if (!struct_registry)
         struct_registry = hm_new();
+    /* gang and chungus share one tag namespace (see docs
+       the-brainrot-programming-language.md 7.10), so a second definition of an
+       existing tag is a redefinition error, not a silent overwrite (#189).
+       Keep the first definition in the lookup table; still thread the rejected
+       def onto the teardown list so free_struct_registry() reclaims it rather
+       than leaking it. The `lit` alias path in lang.y guards this before
+       constructing a def; this covers the plain `gang Foo {...}` /
+       `chungus Foo {...}` grammar action, which did not. */
+    if (get_struct_def(def->name))
+    {
+        char msg[MAX_BUFFER_LEN];
+        snprintf(msg, sizeof(msg),
+                 "Redefinition of tag '%s' -- gang and chungus share one tag "
+                 "namespace",
+                 def->name.data ? def->name.data : "?");
+        yyerror_current_line(msg);
+        struct_def_had_error = true;
+        def->next_def = struct_registry_list;
+        struct_registry_list = def;
+        return;
+    }
     size_t len = def->name.len;
     hm_put(struct_registry, def->name.data, len, def, sizeof(StructDef));
     def->next_def = struct_registry_list;
--- a/test_cases/duplicate_tag_gang_chungus_fail.brainrot
+++ b/test_cases/duplicate_tag_gang_chungus_fail.brainrot
@@ -1,0 +1,8 @@
+gang Foo { rizz a; };
+chungus Foo { chad f; };
+
+skibidi main {
+    gang Foo g;
+    g.a = 1;
+    bussin 0;
+}
--- a/test_cases/duplicate_tag_gang_gang_fail.brainrot
+++ b/test_cases/duplicate_tag_gang_gang_fail.brainrot
@@ -1,0 +1,6 @@
+gang Foo { rizz a; };
+gang Foo { rizz b; };
+
+skibidi main {
+    bussin 0;
+}
--- a/test_cases/gang_chungus_distinct_tags.brainrot
+++ b/test_cases/gang_chungus_distinct_tags.brainrot
@@ -1,0 +1,14 @@
+gang Point { rizz x; rizz y; };
+chungus Num { rizz i; chad f; };
+
+skibidi main {
+    gang Point p;
+    p.x = 3;
+    p.y = 4;
+    chungus Num n;
+    n.i = 42;
+    yapping("%d", p.x);
+    yapping("%d", p.y);
+    yapping("%d", n.i);
+    bussin 0;
+}
--- a/tests/expected_results.json
+++ b/tests/expected_results.json
@@ -428,5 +428,8 @@
     "file_io_ragequit_no_leak": "ExitCode:4",
     "file_io_stale_handle_fail": "close a 0\nsame handle 0\nb works, pos 5\nnow using the retired handle, while b is still open:\nStderr:\nError: yapto: not an open SAUCE -- it was already closed with peaceout, or was never a handle at all, at line 62\n",
     "freed_keywords_as_identifiers": "functions 42 42\nfields    2 3\nlocals    5 7\n",
-    "sieve_of_eras": "2\n3\n5\n7\n11\n13\n17\n19\n23\n29\n31\n37\n41\n43\n47\n53\n59\n61\n67\n71\n73\n79\n83\n89\n97"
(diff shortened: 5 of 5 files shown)


leo-aa88 (commented):
# APPROVE

**LGTM.** Adversarial pass on this semantic fix. I checked out the head in an isolated worktree, traced the ownership/error paths in source, and built + ran the shipped fixtures plus several stress cases under sanitizers and valgrind. Everything holds; no double-free, no double-report, no over-rejection.

## Findings

### Contract / ownership traced end-to-end
- **CONFIRMED — single choke point, no double-report.** `register_struct_def` has exactly two callers (`lang.y:351`, `:706`). The `lit`-alias caller (`register_aggregate_typedef`) already guards duplicates and **`return`s before** its `register_struct_def` call (`lang.y:331–337`), so its "…already defined" path and the new "Redefinition of tag…" path are mutually exclusive — never both. The plain grammar action (`:706`) had no pre-guard; that's the gap this closes. Verified live: a plain-vs-`lit` tag collision emits a single error, not two.
- **CONFIRMED — leak-clean, no double-free.** `free_struct_registry` shallow-frees the hashmap index (`hm_free_shallow`) and frees the actual defs via `struct_registry_list`. The rejected def is threaded onto that list but never `hm_put`, so it's freed exactly once; the kept (first) def stays the sole map entry. valgrind: **0 def / 0 indirect lost, 0 errors** on both fail fixtures — and also on a triple-definition case (two rejected defs on the list) and the cross-path collision.

### Behavior verified live (worktree @ 819dbfd, isolated from the user tree)
- The 3 shipped fixtures produce byte-exact expected output; `pytest -k "duplicate_tag or gang_chungus_distinct"` → 3 passed.
- **Adversarial extras**, all correct + valgrind-clean: reverse order (`chungus` then `gang`) → error at line 2; triple definition → clean diagnostics at **both** lines 2 and 3, no crash; plain-vs-inline-`lit` collision → single "already defined" error.
- **No over-rejection:** distinct tags coexist (`3\n4\n42`), and all 29 existing `lit_*` tests still pass.

## QUESTION (non-blocking)

"Type of Change" is ticked *Bug fix (non-breaking)*, but this is strictly a behavior change: a program that previously parsed (silent last-def-wins) now hard-errors. The PR's own "Behavior change note" acknowledges this and correctly argues the old behavior was already forbidden by §7.10 of the language reference, so nothing conformant regresses. I'd have ticked "Breaking change" for bookkeeping, but the substance is right and it does not affect mergeability.

## VERDICT

Correct, minimal, and it makes the implementation match an already-documented contract, replacing a misleading `Undefined variable` failure with a precise parse-time diagnostic. Guard placement is right (the one path both definitions share), ownership is sound, and the error paths are leak-clean under every combination I could throw at them. Deserves to merge. (Posting as COMMENT rather than a formal approval only because the reviewer account is also the PR author — GitHub rejects self-approval.)

leo-aa88 (commented):
# APPROVE

**LGTM.** Re-review of new head `5af90de` (was `819dbfd`).

The only change is a **merge commit** integrating current `origin/main` (needed to resolve the shared `tests/expected_results.json` append-conflict). The content commit `819dbfd` — the one I reviewed and stress-tested (guard in `register_struct_def`, three fixtures) — is unchanged in the history.

Verified the merge resolved cleanly:
- Merged `expected_results.json` is valid JSON; this PR's net diff vs main is exactly its own three keys (`gang_chungus_distinct_tags`, `duplicate_tag_gang_chungus_fail`, `duplicate_tag_gang_gang_fail`), appended after main's latest entry — nothing clobbered.
- Net code diff vs main is the same +21-line `ast.c` guard I already traced; nothing newly merged into main touches `register_struct_def` or struct registration, so no semantic interaction.

## VERDICT

Same APPROVE as before — pure merge-from-main, content and behavior unchanged, resolution clean. Deserves to merge. (COMMENT, not formal approval, only because the reviewer account is also the PR author.)

PR #351, opened 2026-09-06
test: guard maxxing(*q)/maxxing(q) on a struct pointer parameter (#200)

### LINKED ISSUE #200: maxxing(*q) on a pointer-to-struct/union parameter still returns 0
## Summary

#198 teaches `maxxing` (`sizeof`) to handle a dereferenced struct/union
pointer (`maxxing(*q)`), but only when `q` is a local or global. The
same operand on a **function parameter** still errors and prints `0`.

Leftover from reviewing #198 (the #195 fix). Locals work; parameters
don't, because the new deref path needs `var->struct_name` and pointer
parameters never get one.

## Reproduction

```c
gang Point {
    rizz x;
    rizz y;
};

skibidi size_of_ptr(gang Point *q) {
    yapping("%lu", maxxing(*q));   // expected 8
    yapping("%lu", maxxing(q));    // pointer size, works
}

skibidi main {
    gang Point p;
    gang Point *qp = &p;
    size_of_ptr(qp);
}
```

Output on the #198 branch:

```
Error: Invalid type in sizeof at line 15
0
8
```

`maxxing(*qp)` in `main` (a local) already returns `8`. `maxxing(p)` on
a by-value `gang Point p` parameter also works. Only the pointer
parameter is broken.

## Root cause

`handle_sizeof()`'s new `VAR_STRUCT` deref branch looks up the operand
variable and does `get_struct_def(var->struct_name)`.

`enter_function_scope()` (ast.c) binds `pointer_level > 0` parameters by
copying only the address:

```c
if (curr_param->pointer_level > 0)
{
    Variable *bound = get_variable(curr_param->name);
    if (bound)
    {
        bound->pointer_level = curr_param->pointer_level;
        bound->value.pvalue = arg_values[i].pvalue;
    }
    /* ... continue; never copies struct_name */
}
```

`var->var_type` is `VAR_STRUCT`, so the deref path is taken, but
`struct_name` is empty → `get_struct_def` returns NULL →
`"Invalid type in sizeof"`. By-value struct params go through the
`VAR_STRUCT` binding case, which *does* `safe_strdup` the tag, which is
why `maxxing(p)` works there.

Locals get `struct_name` from `interpreter_visit_declaration`, which is
why the #198 fixture (`gang Point *q = &p` in `main`) is green.

#193 already flagged pointer-to-struct parameters as "may partially work
via the generic pointer-parameter mechanism — worth verifying
explicitly." This is that gap, hit from sizeof.

## Suggested direction

When binding a pointer-to-struct/union parameter, copy
`curr_param->struct_name` onto the bound variable (same as the by-value
`VAR_STRUCT` case). After that, #198's deref path should just work.

If `get_struct_def` is still NULL, prefer `"Unknown struct or union
type"` over `"Invalid type in sizeof"` — #198 already made that change
on the identifier path in `get_type_size()`.

## Tests

Extend `test_cases/sizeof_struct.brainrot` (or add a sibling fixture)
with a `gang Point *` parameter and `maxxing(*q)`. Also pin
`maxxing(q)` on that parameter (pointer size) so the descriptor path
doesn't regress.

**Testing note:** passing a `gang Point *` parameter currently
double-frees on scope exit (ASan: `hm_free` in `exit_scope` vs
`free_scope` in `cleanup`), even with no `maxxing` at all. Pre-existing,
not introduced by #198. The sizeof test may need that crash fixed or
avoided before it can land under the default ASan build.

## Also relevant

- #195 / #198 — the sizeof-on-struct fix this is leftover from
- #193 — pointer-to-struct params/returns still incomplete
- #196 — pointer-to-struct variables don't follow the pointer for
  member access (separate, but the same "pointer-to-struct storage
  isn't wired through")

leo-aa88:
## Description

#200 reported that `maxxing(*q)` on a **pointer-to-struct parameter** errored and
printed `0`:

```
skibidi size_of_ptr(gang Point *q) {
    yapping("%lu", maxxing(*q));   // errored -> 0
    yapping("%lu", maxxing(q));    // pointer size, worked
}
```

The #198 deref path needed `var->struct_name`, which pointer parameters never
received — the same operand on a *local* already worked.

**On current `main` this is fixed.** `aa001b9` (#336, "support member access on
pointer expressions") wired pointer-to-struct parameter storage through, so the
issue's exact repro now prints the pointee struct size (`8`) for `maxxing(*q)`
and the pointer size for `maxxing(q)`, and the pre-existing double-free on
`gang Point *` scope exit that the issue flagged as a potential blocker no longer
reproduces (clean under ASan and valgrind).

**What was still missing is the regression coverage the issue explicitly asks
for.** `test_cases/sizeof_struct.brainrot` only exercises `maxxing(*q)` on a
*local* `gang Point *q`; nothing pinned the *parameter* path. This adds
`test_cases/sizeof_struct_pointer_param.brainrot`, which pins both:

- `maxxing(*q)` → pointee struct size (`8`), and
- `maxxing(q)` → pointer size, so the descriptor path can't silently regress.

Since `maxxing(q)` is pointer-width, its value differs by data model, so I added
the wasm32 ILP32 override (`8\n4`) to `run_wasm_tests.mjs` next to the existing
pointer-size overrides — exactly like `native_sizeof_ptr_result`. The first line
(`maxxing(*q)`, the struct size) is platform-independent.

## Related Issue

Fixes #200

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement
- [ ] Refactor

## Checklist

> This PR adds coverage for behavior already fixed by #336; the new fixture is
> the regression guard the issue requested. It changes no interpreter/stdlib C
> source — only a `test_cases/` fixture, its `expected_results.json` entry, and a
> wasm expected-value override — so `cppcheck`/`clang-format` are unaffected.
> `make test` → **507 passed**; full `make valgrind` sweep → exit 0; the fixture
> is leak- and error-clean under both ASan and valgrind.

### DIFF
--- a/test_cases/sizeof_struct_pointer_param.brainrot
+++ b/test_cases/sizeof_struct_pointer_param.brainrot
@@ -1,0 +1,21 @@
+🚽 Regression for #200: maxxing (sizeof) on a struct POINTER PARAMETER.
+🚽 maxxing(*q) must report the pointee struct size, and maxxing(q) the pointer
+🚽 size, when q is a `gang Point *` function parameter -- not just a local
+🚽 (which #198 added and sizeof_struct.brainrot already pins). The parameter
+🚽 path was wired through by #336 (member access on pointer expressions); this
+🚽 guards it from regressing.
+gang Point {
+    rizz x;
+    rizz y;
+};
+
+skibidi size_of_ptr(gang Point *q) {
+    yapping("%lu", maxxing(*q));   🚽 pointee struct size -> 8
+    yapping("%lu", maxxing(q));    🚽 pointer size -> 8 (native LP64)
+}
+
+skibidi main {
+    gang Point p;
+    gang Point *qp = &p;
+    size_of_ptr(qp);
+}
--- a/tests/expected_results.json
+++ b/tests/expected_results.json
@@ -429,5 +429,6 @@
     "file_io_stale_handle_fail": "close a 0\nsame handle 0\nb works, pos 5\nnow using the retired handle, while b is still open:\nStderr:\nError: yapto: not an open SAUCE -- it was already closed with peaceout, or was never a handle at all, at line 62\n",
     "freed_keywords_as_identifiers": "functions 42 42\nfields    2 3\nlocals    5 7\n",
     "sieve_of_eras": "2\n3\n5\n7\n11\n13\n17\n19\n23\n29\n31\n37\n41\n43\n47\n53\n59\n61\n67\n71\n73\n79\n83\n89\n97",
-    "parse_error_unsized_array_arity_fail": "Error: Initializer count does not match array dimensions at line 1\nParsing failed"
+    "parse_error_unsized_array_arity_fail": "Error: Initializer count does not match array dimensions at line 1\nParsing failed",
+    "sizeof_struct_pointer_param": "8\n8"
 }
--- a/tests/run_wasm_tests.mjs
+++ b/tests/run_wasm_tests.mjs
@@ -118,6 +118,11 @@ const WASM_EXPECTED_OVERRIDES = {
   giga: "4\n4",
   giga_array: "1\n2\n3\n12",
   native_sizeof_ptr_result: "4",
+  // sizeof_struct_pointer_param's second line is maxxing(q) on a `gang Point *`
+  // parameter -- a pointer, so 4 bytes on wasm32 ILP32 vs 8 on native LP64
+  // (same root cause as native_sizeof_ptr_result). The first line, maxxing(*q),
+  // is the pointee struct size (2 rizz = 8) and is platform-independent.
+  sizeof_struct_pointer_param: "8\n4",
   native_identity_abi_type_char_array: "8",
   native_void_pointer_struct_field: "8",
   // struct_field_long_modifier's `Meters` field is a `lit giga rizz`


leo-aa88 (commented):
# APPROVE

**LGTM.** Adversarial pass on this regression-coverage PR. Checked out the head in an isolated worktree, built, and ran the fixture end-to-end (sanitized + valgrind) plus a semantic sanity check that the test guards the *right* value. Everything holds.

## Findings

Verified:
- **CONFIRMED — values are right.** Native run prints `8\n8`: `maxxing(*q)` = pointee struct size (`gang Point` = 2×`rizz` = 8), `maxxing(q)` = pointer width (8, LP64). Matches the `sizeof_struct_pointer_param: "8\n8"` entry.
- **CONFIRMED — the test actually pins the pointee path, not a coincidence.** With a 3-`rizz` struct pointer parameter, `maxxing(*q)` returns `12`, so the deref genuinely resolves `var->struct_name` for a *parameter* (the #200 gap), not the pointer size or a constant. This is exactly the regression the issue asked to guard.
- **CONFIRMED — #200's failure modes are gone.** `maxxing(*q)` no longer errors/prints `0` (exit 0, clean output), and valgrind on the `gang Point *` scope exit reports **0 def / 0 indirect lost, 0 errors** — the double-free the issue flagged as a potential blocker does not reproduce.
- **CONFIRMED — harness plumbing.** `expected_results.json` is valid JSON; `run_wasm_tests.mjs` passes `node --check`; `pytest -k sizeof_struct_pointer_param` → passed.
- **CONFIRMED — wasm override is consistent.** `8\n4` correctly keeps the platform-independent struct size (line 1) and switches only the pointer width to ILP32's 4 (line 2), matching the neighboring `native_sizeof_ptr_result: "4"` convention.

## VERDICT

Test-only, no C source touched, and it closes exactly the coverage gap #200 called out — the parameter path for `maxxing(*q)`/`maxxing(q)` on a struct pointer, with a wasm override that mirrors the established pointer-size pattern. Correct values, leak-clean, and it genuinely exercises the pointee resolution rather than a lucky constant. Deserves to merge. (Posting as COMMENT rather than a formal approval only because the reviewer account is also the PR author — GitHub rejects self-approval.)

PR #352, opened 2026-09-06
fix: maxxing (sizeof) of a struct-returning call (#201)

### LINKED ISSUE #201: maxxing still rejects struct-returning calls and double-derefs
## Summary

#198 adds a `VAR_STRUCT` case to `handle_sizeof()`, but it only
special-cases two node shapes:

- `NODE_STRUCT_ACCESS` (e.g. `l.start`)
- `OP_DEREFERENCE` of a `NODE_IDENTIFIER` (e.g. `*q`)

Any other struct/union-typed expression still falls through to
`"Invalid type in sizeof"` and prints `0`. Two that come up immediately:

1. A function call returning a struct/union by value —
   `maxxing(make_point(1, 2))`
2. A double (or deeper) dereference — `maxxing(**pp)` where
   `pp` is `gang Point **`

Leftover from reviewing #198 (the #195 fix). Same operand class, just
not the two shapes that PR wired up.

## Reproduction

```c
gang Point {
    rizz x;
    rizz y;
};

gang Point make_point(rizz a, rizz b) {
    gang Point p;
    p.x = a;
    p.y = b;
    bussin p;
}

skibidi main {
    gang Point p;
    gang Point *q = &p;
    gang Point **pp = &q;

    yapping("%lu", maxxing(make_point(1, 2)));  // expected 8
    yapping("%lu", maxxing(**pp));              // expected 8
}
```

Output on the #198 branch:

```
Error: Invalid type in sizeof at line 21
0
Error: Invalid type in sizeof at line 21
0
```

(`yylineno` is already at EOF by the time these evaluate — a pre-existing
runtime-error line-number quirk, not new here.)

For comparison, these already work after #198 and should stay working:

| Operand | Result |
|---|---|
| `maxxing(p)` | 8 |
| `maxxing(*q)` (local) | 8 |
| `maxxing(q)` / `maxxing(&p)` | pointer size (8 on LP64) |
| `maxxing(l.start)` | 8 |
| `maxxing(make_point)` as a *variable*, not a call | n/a |

## Root cause

`handle_sizeof()`'s non-identifier branch switches on
`get_expression_type(expr)`. For `VAR_STRUCT` with pointer-level 0 it
then pattern-matches the AST node:

- `NODE_FUNC_CALL` returning `VAR_STRUCT` is neither a field access nor
  a single identifier deref, so it misses both arms. The return tag is
  already on `Function.return_struct_name` — sizeof must use that and
  **must not** execute the call (C `sizeof` does not evaluate its
  operand).
- `**pp` is `OP_DEREFERENCE` whose operand is itself a deref, not a
  `NODE_IDENTIFIER`, so the identifier-only arm misses it.

Adding another `if (expr->type == ...)` for each new shape will keep
leaving holes. A small `get_expression_struct_def(ASTNode *)` (or
equivalent) that walks identifier / field access / deref / func-call
and returns the `StructDef` would close this class in one place.
`get_expression_type()` / `get_expression_pointer_level()` already
do this kind of walk for the primitive type and the pointer depth.

## Suggested direction

- Func-call: `get_function(...)->return_struct_name` → `get_struct_def`
  → `total_size`. Do not call the function.
- Nested deref: strip `OP_DEREFERENCE` until the identifier (or other
  base) and use that variable's `struct_name`, same as the single-deref
  arm. `get_expression_pointer_level` already computed that the result
  is a value (`plevel == 0`).
- Prefer a shared helper over a third and fourth special case.

## Tests

Extend `test_cases/sizeof_struct.brainrot` (and
`tests/expected_results.json`) with:

- `maxxing(make_point(1, 2))` — struct-returning call
- `maxxing(**pp)` — double deref of a `gang Point **`

While touching that fixture, also pin the cases that already work after
#198 but aren't covered yet, so they can't regress silently:

- `maxxing(q)` / `maxxing(&p)` — pointer size
- a union-typed field (e.g. `w.data` on a struct wrapping a `chungus`)
- `maxxing(p)` on a by-value `gang Point p` parameter

## Also relevant

- #195 / #198 — the sizeof-on-struct fix this is leftover from
- #193 — struct-valued sub-expressions as args/returns (related
  "only identifiers are wired" restriction, different call path)

leo-aa88:
## Description

#198 taught `handle_sizeof()` two struct/union operand shapes — field access
(`l.start`) and single-identifier deref (`*q`) — but left others falling through
to `Error: Invalid type in sizeof` and printing `0`. The shared resolver
`get_struct_def_for_expression()` already walks identifiers, array accesses,
field accesses, **nested** derefs, and pointer arithmetic — so `maxxing(**pp)`
already works on current `main` (via #336) — but it had **no `NODE_FUNC_CALL`
case**, so a struct-returning call still errored:

```
gang Point make_point(rizz a, rizz b) { ... bussin p; }
...
yapping("%lu", maxxing(make_point(1, 2)));   // was: Error + 0, now: 8
```

### Fix

Add the `NODE_FUNC_CALL` case to that one shared resolver — the issue's
preferred *"shared helper over a third and fourth special case."* It resolves
the callee's declared return tag (`func->return_desc.struct_name`) to its
`StructDef`, mirroring the existing `NODE_FUNC_CALL` guard used for parent
resolution elsewhere in `ast.c`.

Key correctness points:
- **The call is not executed** — `sizeof` never evaluates its operand. Verified:
  a `make_point` with a side-effecting `yapping` prints nothing under `maxxing`.
- **Builtins excluded** — no native returns a struct across the ABI.
- **Pointer-returning calls** are still sized as a pointer by `handle_sizeof()`'s
  `plevel > 0` branch, untouched.

### Tests

`test_cases/sizeof_struct_call_and_deref.brainrot` pins the two shapes the issue
names plus the neighbours it asks to guard:

| Operand | Expected |
|---|---|
| `maxxing(make_point(1, 2))` (struct-returning call) | 8 |
| `maxxing(**pp)` (double deref) | 8 |
| `maxxing(q)` / `maxxing(&p)` (pointer) | 8 native / 4 wasm |
| `maxxing(w.data)` (union-typed field) | 8 |
| `maxxing(bp)` (by-value struct parameter) | 8 |

The two pointer-size lines get the wasm32 ILP32 override (`4` vs `8`) in
`run_wasm_tests.mjs`, exactly like `native_sizeof_ptr_result`. The fixture also
asserts (via a side-effecting `make_point`) that the operand is never evaluated.

## Related Issue

Fixes #201

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement
- [ ] Refactor

## Checklist

> `make test` → **507 passed**; full `make valgrind` sweep → exit 0; the new
> fixture is leak- and error-clean under both ASan and valgrind. `make
> format-check` clean. `make cppcheck` couldn't run locally (this box has
> cppcheck 2.7 < the required 2.13); the change is a small switch case mirroring
> existing code, left for CI's `static-analysis` job.

### DIFF
--- a/ast.c
+++ b/ast.c
@@ -3027,6 +3027,26 @@ static StructDef *get_struct_def_for_expression(ASTNode *expr)
             return get_struct_def_for_expression(expr->data.op.right);
         return NULL;
     }
+    case NODE_FUNC_CALL:
+    {
+        /* sizeof of a struct-returning call (`maxxing(make_point(1, 2))`,
+           #201): take the layout from the callee's declared return tag.
+           The call is NOT executed -- sizeof never evaluates its operand --
+           the tag alone gives the size. Only a by-value struct return
+           (pointer_level == 0) has a struct layout here; a pointer return is
+           sized as a pointer by handle_sizeof()'s plevel > 0 branch. No native
+           returns a struct across the ABI, so builtins never reach a def
+           (mirrors the NODE_FUNC_CALL guard in get_struct_field_alignment's
+           parent resolution above). */
+        if (is_builtin_function(expr->data.func_call.function_name))
+            return NULL;
+        Function *func = get_function(expr->data.func_call.function_name);
+        if (func && func->return_desc.type == VAR_STRUCT &&
+            func->return_desc.pointer_level == 0 &&
+            func->return_desc.struct_name.data)
+            return get_struct_def(func->return_desc.struct_name);
+        return NULL;
+    }
     default:
         return NULL;
     }
--- a/test_cases/sizeof_struct_call_and_deref.brainrot
+++ b/test_cases/sizeof_struct_call_and_deref.brainrot
@@ -1,0 +1,36 @@
+🚽 Regression for #201: maxxing (sizeof) on struct/union-typed expressions that
+🚽 are neither a plain identifier, a field access, nor a single-identifier deref
+🚽 -- the shapes #198 left erroring with "Invalid type in sizeof". sizeof never
+🚽 evaluates its operand, so the struct-returning call below must size from the
+🚽 callee's return tag WITHOUT running make_point (no "ran" line prints).
+🚽 Also pins the neighbouring shapes that already work so they can't regress:
+🚽 a pointer operand, a union-typed field, and a by-value struct parameter.
+gang Point { rizz x; rizz y; };            🚽 2x rizz -> 8
+chungus Data { rizz i; gigachad d; };      🚽 union: max(4, 8) -> 8
+gang Wrap { chungus Data data; };          🚽 struct with a union-typed field
+
+gang Point make_point(rizz a, rizz b) {
+    yapping("make_point ran");             🚽 must NOT print under maxxing
+    gang Point p;
+    p.x = a;
+    p.y = b;
+    bussin p;
+}
+
+skibidi by_value_param(gang Point bp) {
+    yapping("%lu", maxxing(bp));           🚽 by-value struct param -> 8
+}
+
+skibidi main {
+    gang Point p;
+    gang Point *q = &p;
+    gang Point **pp = &q;
+    gang Wrap w;
+
+    yapping("%lu", maxxing(make_point(1, 2)));  🚽 struct-returning call -> 8
+    yapping("%lu", maxxing(**pp));              🚽 double deref -> 8
+    yapping("%lu", maxxing(q));                 🚽 pointer size (8 LP64 / 4 wasm)
+    yapping("%lu", maxxing(&p));                🚽 pointer size (8 LP64 / 4 wasm)
+    yapping("%lu", maxxing(w.data));            🚽 union-typed field -> 8
+    by_value_param(p);                          🚽 prints 8
+}
--- a/tests/expected_results.json
+++ b/tests/expected_results.json
@@ -429,5 +429,6 @@
     "file_io_stale_handle_fail": "close a 0\nsame handle 0\nb works, pos 5\nnow using the retired handle, while b is still open:\nStderr:\nError: yapto: not an open SAUCE -- it was already closed with peaceout, or was never a handle at all, at line 62\n",
     "freed_keywords_as_identifiers": "functions 42 42\nfields    2 3\nlocals    5 7\n",
     "sieve_of_eras": "2\n3\n5\n7\n11\n13\n17\n19\n23\n29\n31\n37\n41\n43\n47\n53\n59\n61\n67\n71\n73\n79\n83\n89\n97",
-    "parse_error_unsized_array_arity_fail": "Error: Initializer count does not match array dimensions at line 1\nParsing failed"
+    "parse_error_unsized_array_arity_fail": "Error: Initializer count does not match array dimensions at line 1\nParsing failed",
+    "sizeof_struct_call_and_deref": "8\n8\n8\n8\n8\n8"
 }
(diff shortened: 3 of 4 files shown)


leo-aa88 (commented):
# APPROVE

**LGTM.** Adversarial pass on this `sizeof`-of-call fix. Checked out the head in an isolated worktree, traced `handle_sizeof` end-to-end, and built + ran the fixture plus pointer/scalar-returning call variants under sanitizers and valgrind. The change is minimal, correctly guarded, and the "operand is never evaluated" invariant holds across every call shape.

## Findings

### Contract traced end-to-end
`handle_sizeof` resolves a `VAR_STRUCT` operand by `plevel`: `plevel > 0` → pointer size; `plevel == 0` → `get_struct_def_for_expression()` → `total_size` (else "Invalid type in sizeof"). The new `NODE_FUNC_CALL` case slots into exactly the `plevel == 0` struct branch and is properly guarded:
- **CONFIRMED — builtins excluded.** `is_builtin_function()` short-circuits to NULL (verified the symbol exists and is used identically at `ast.c:457,1228`); no native returns a by-value struct.
- **CONFIRMED — only by-value struct returns resolve.** Requires `return_desc.type == VAR_STRUCT && pointer_level == 0 && struct_name.data`; a pointer-returning call has `pointer_level > 0` and is sized by the pre-existing pointer branch, untouched. `return_desc` fields verified in `ast.h:262-274`.

### Behavior verified live (worktree @ 3be20d1, isolated from the user tree)
- Fixture prints exactly `8\n8\n8\n8\n8\n8` and **no `make_point ran`** — sizeof does not evaluate its operand. `pytest -k sizeof_struct_call_and_deref` → passed.
- **Adversarial: pointer-returning call** `maxxing(mk(q))` → `8` (pointer size), no `mk ran`, no error — the new case correctly returns NULL and the `plevel > 0` branch takes over.
- **Adversarial: scalar-returning call** `maxxing(add(1,2))` → `4`, no `add ran`, no error — scalar operands never reach the struct branch, so no regression from adding the case.
- valgrind (`--leak-check=full --track-fds=yes`): **0 def / 0 indirect lost, 0 errors**.
- `expected_results.json` valid; `run_wasm_tests.mjs` passes `node --check`; the wasm override `8\n8\n4\n4\n8\n8` correctly ILP32-adjusts only the two pointer lines, matching the `native_sizeof_ptr_result` convention.

## VERDICT

Right fix in the right place: one guarded case added to the single shared resolver (the issue's stated preference over more special-casing), sizing a struct-returning call from its declared return tag without evaluating it. Correct across by-value / pointer / scalar returns, leak-clean, and comprehensively tested including the non-evaluation assertion. Deserves to merge. (Posting as COMMENT rather than a formal approval only because the reviewer account is also the PR author — GitHub rejects self-approval.)

leo-aa88 (commented):
# APPROVE

**LGTM.** Re-review of new head `4acf1e3` (was `3be20d1`).

The only changes are **merge commits** integrating current `origin/main` (needed for the shared `expected_results.json` append-conflict, and because #350 also landed in `ast.c`). The content commit `3be20d1` I reviewed and stress-tested is unchanged.

Verified the merge is sound, not just conflict-free:
- No conflict markers anywhere; merged `ast.c` correctly contains **both** #350's `register_struct_def` guard and this PR's `NODE_FUNC_CALL` sizeof case (they edit different, far-apart functions — no interaction).
- Merged `expected_results.json` valid; this PR's net contribution is its own `sizeof_struct_call_and_deref` key, nothing clobbered.
- Rebuilt from the merge head: `sizeof_struct_call_and_deref` → `8\n8\n8\n8\n8\n8`, exit 0 — behavior identical to my original review.

## VERDICT

Same APPROVE — clean, functionally-verified merge-from-main, content and behavior unchanged. Deserves to merge. (COMMENT, not formal approval, only because the reviewer account is also the PR author.)

PR #353, opened 2026-09-06
fix(ci): fail the valgrind sweep on child crashes, not just exit 100 (#203)

### LINKED ISSUE #203: run_valgrind_tests.sh treats only exit 100 as failure, so crashes look green
## Summary

`run_valgrind_tests.sh` only fails when valgrind exits `100` (`--error-exitcode=100`):

```bash
if [[ $valgrind_exit_code -eq 100 ]]; then
    echo "Valgrind detected memory issues in $f"
    exit 1
fi
```

Any other non-zero — missing binary, ASan startup SIGSEGV (typically 139), interpreter crash — is treated as success and the loop continues. That is how #186 stayed hidden: valgrind never ran the program (`0 allocs, 0 frees`) and the script still exited 0.

## Why this matters

Even after #202 (`make valgrind` builds a non-ASan `brainrot-valgrind`), a mis-invoked script (`./run_valgrind_tests.sh` with no args / ASan `./brainrot`) or a hard crash still reports green.

Expected Brainrot parse/semantic errors must keep passing: those fixtures exit non-zero on purpose, and valgrind then forwards that exit code when there are no leaks. Do **not** fail on every non-zero.

## Suggested direction

Fail (or at least abort the suite) on unexpected valgrind/child deaths without treating application-level error exits as leaks. For example:

- treat signal exits (128+N, e.g. 139 SIGSEGV / 134 SIGABRT) as failure
- treat valgrind usage errors / missing binary as failure
- keep allowing the interpreter's own non-zero exits when valgrind reports no errors

Spotted while reviewing #202.

leo-aa88:
## Description

`run_valgrind_tests.sh` only failed when Valgrind exited `100`
(`--error-exitcode=100`). Any *other* non-zero was treated as success and the
loop continued — so a child that **crashed** slipped through:

- an ASan-build binary mis-run under Valgrind that SIGSEGVs at startup — the
  #186 case, where Valgrind ran nothing (`0 allocs, 0 frees`) and the script
  still exited 0,
- an assertion `abort()`, a stack overflow, any hard interpreter crash.

These die by signal, so Valgrind re-raises and the script sees `128+N` (e.g.
`139` SIGSEGV, `134` SIGABRT) — never `100`.

### Fix

After the existing exit-100 and 126/127 launch-failure checks, add a
**signal-death gate**: exit `> 128` is a crash → fail the sweep.

The issue's constraint — *"Expected Brainrot parse/semantic errors must keep
passing … Do not fail on every non-zero"* — is preserved: parse/semantic errors
exit `1`, and the largest `ragequit(N)` fixture exits `69`, all well below 128,
and Valgrind forwards them unchanged when it finds no leaks. Legitimate `126/127`
exits (e.g. `ragequit(127)`) are still handled by the launch-failure check just
above.

Verified exit codes under `valgrind --error-exitcode=100`: clean `→0`,
`return 1` `→1`, `abort()` `→134`.

### Tests

`tests/test_valgrind_runner.py` drives the **actual script** against tiny
crafted targets in a temp directory and asserts all three gate outcomes:

| Target | Old runner | New runner |
|---|---|---|
| clean (`exit 0`) | pass | **pass** |
| app error (`exit 1`, like a parse error) | pass | **pass** |
| signal death (`abort()` → 134) | **pass (bug)** | **fail** |

To point the script at a temp fixtures dir, I added a backward-compatible
`TESTCASES_DIR` override (defaults to `test_cases`, also handy for running a
subset). The test is a pytest module, so both `make test` and CI's existing
Pytest step run it; it skips when valgrind or a C compiler is unavailable (the
same tools `make valgrind` already requires).

## Related Issue

Fixes #203

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement
- [ ] Refactor

## Checklist

> This changes the CI harness, not interpreter/stdlib C source, so
> `cppcheck`/`clang-format` are unaffected (`make format-check` clean). `make
> test` → **509 passed** (incl. the 3 new self-tests); the real `make valgrind`
> sweep still exits 0 (the `TESTCASES_DIR` default and the new gate don't
> misfire on the actual fixtures).

### DIFF
--- a/run_valgrind_tests.sh
+++ b/run_valgrind_tests.sh
@@ -8,6 +8,12 @@
 # for anyone who built that target first.
 TARGET="${1:-./brainrot-valgrind}"
 
+# Directory the fixtures are swept from. Defaults to the repo's test_cases/;
+# overridable so the suite can be pointed at a subset and so
+# tests/test_valgrind_runner.py can drive this script against crafted targets
+# in a temp directory without touching the real fixtures.
+TESTCASES_DIR="${TESTCASES_DIR:-test_cases}"
+
 if ! command -v valgrind >/dev/null 2>&1; then
     echo "Error: valgrind is not installed or not in PATH" >&2
     exit 1
@@ -18,7 +24,7 @@ if [[ ! -x "$TARGET" ]]; then
     exit 1
 fi
 
-for f in test_cases/*.brainrot; do
+for f in "$TESTCASES_DIR"/*.brainrot; do
     echo "Running Valgrind on $f..."
     base=$(basename "$f" .brainrot)
 
@@ -85,6 +91,25 @@ for f in test_cases/*.brainrot; do
     if [[ $valgrind_exit_code -eq 126 || $valgrind_exit_code -eq 127 ]] \
         && grep -q '^valgrind: ' "$fd_log"; then
         echo "Valgrind could not execute '$TARGET' on $f (exit $valgrind_exit_code)"
+        rm -f "$fd_log"
+        exit 1
+    fi
+
+    # Fail on a child that died from a SIGNAL rather than exiting normally. When
+    # the child is killed by signal N, Valgrind re-raises it and this script
+    # sees exit code 128+N -- e.g. 139 (SIGSEGV), 134 (SIGABRT), 136 (SIGFPE).
+    # The exit-100 gate above never catches these: a crash that isn't a
+    # Valgrind-detected memory error (an ASan-build binary mis-run under
+    # Valgrind and SIGSEGV'ing at startup -- the #186 case -- an assertion
+    # abort, a stack overflow) leaves --error-exitcode unfired, so the sweep
+    # used to sail past a program that executed nothing (#203). The
+    # interpreter's own deliberate non-zero exits stay well below this: parse/
+    # semantic errors exit 1, and the largest ragequit(N) fixture exits 69, so
+    # >128 cleanly separates "crashed" from "exited with an error". (126/127
+    # launch failures are handled just above; a child that legitimately exits
+    # 126/127, e.g. ragequit(127), is passed through there.)
+    if [[ $valgrind_exit_code -gt 128 ]]; then
+        echo "Valgrind child crashed (killed by signal $((valgrind_exit_code - 128))) on $f (exit $valgrind_exit_code)"
         rm -f "$fd_log"
         exit 1
     fi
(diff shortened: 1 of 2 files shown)


leo-aa88 (commented):
# APPROVE

**LGTM.** Adversarial pass on this CI-gate fix. Checked out the head in an isolated worktree, read the full runner gate, audited every fixture's exit code, and drove the actual script (self-test + real fixtures) to confirm it neither under- nor over-fires.

## Findings

None blocking. One non-blocking QUESTION below.

### Gate correctness traced end-to-end
The new `valgrind_exit_code -gt 128` check sits after the exit-100 and 126/127 gates:
- **CONFIRMED — crashes now fail.** A signal-killed child (Valgrind re-raises → 128+N) previously slipped through the exit-100-only gate; it now fails. Verified live: an `abort()` target → 134 → runner exits non-zero and prints "crashed".
- **CONFIRMED — legit exits still pass.** I audited every fixture: literal `ragequit` args are `{0,1,4,69}`, `ragequit(3.9)`→3, `file_io_ragequit_no_leak`→4; all other `ragequit(...)` are semantic-error fixtures that exit 1. Max is 69 — nothing reaches 128. Ran the real runner over the highest-exit fixtures (`ragequit(69)`, exit-4, exit-3): exits 0, no false gate message.
- **CONFIRMED — `ragequit(127)` pass-through preserved.** The 126/127 gate fires only when the log also has a `^valgrind: ` launch-failure line, which a genuine child exit never prints — so a legit 126/127 exit is not failed, and a real exec failure still is.

### Test quality
- **CONFIRMED** `tests/test_valgrind_runner.py` drives the *real* script (not a reimplementation) via the new backward-compatible `TESTCASES_DIR` override, asserting all three outcomes: clean→pass, app-error(1)→pass, SIGABRT(134)→fail. It skips cleanly without valgrind/cc. Ran it: 3 passed.

## QUESTION (MINOR, non-blocking)

The gate cannot distinguish a signal death (128+N) from a *deliberate* `ragequit(N)` with N > 128 — at the shell, `$?` is identical for `exit(200)` and some signal deaths. No such fixture exists today (and exit codes >125 are conventionally reserved), so this is latent, not a live bug. If a future fixture ever needs `ragequit(>128)`, this gate would flag it as a crash. Worth a one-line note in the script's comment, or clamping ragequit's arg to 8 bits at the low end — but nothing to change now.

## VERDICT

Closes exactly the #203/#186 hole — a crashed child that executed nothing no longer reports green — while provably preserving the "parse/semantic errors and ragequit exits keep passing" constraint. The self-test exercises the real script across all gate branches. Correct, well-scoped, and well-tested. Deserves to merge. (Posting as COMMENT rather than a formal approval only because the reviewer account is also the PR author — GitHub rejects self-approval.)

PR #354, opened 2026-09-06
test: pin struct field RMW and unparenthesized field arithmetic (#245)

leo-aa88:
## Description

#245's root cause — member access `.` binding **looser** than binary/assignment
operators, so `p.x = p.x + 1` and `a.x + a.y` misparsed (`(s = p).x`, `(a.x + a).y`)
and crashed with an ASan SEGV — was fixed by `2dfc5e3`, which moved `DOT` to the
highest precedence tier in `lang.y` and added `struct_field_scalar_assign.brainrot`.

I verified on current `main` that every failure-surface case from the issue now
works with no parens and no sanitizer abort (`p.x = p.x + 1` → 6, `a.x + a.y` → 7,
`1 + a.x`, `a.x - a.y`, `s = p.x`, `a.x && a.y`, nested `a.b.c`).

The existing fixture pins `v = m.x`, `v = m.x + m.y` (arithmetic on an assignment
RHS), and nested `o.inner.v` — but **not** the issue's headline **read-modify-write**
case (`p.x = p.x + 1`, writing back into the same field) nor field arithmetic used
**directly as a call argument** (no assignment wrapper, no parens). This adds
`struct_field_rmw_and_field_arith.brainrot` to close that acceptance-criteria gap:

- `p.x = p.x + 1` — RMW back into the same field (5 → 6)
- `p.y = p.y + a.x` — cross-struct RMW (0 + 3 → 3)
- `a.x + a.y`, `1 + a.x`, `a.x - a.y` — field/literal arithmetic as direct `yapping` args
- `edgy (a.x && a.y)` — `&&` of two fields in a condition

Also checked the issue's other asks: no docs note claims field arithmetic needs
parens (the one parens mention is a *positive* statement about parenthesized
struct-pointer access), so nothing to remove.

## Related Issue

Fixes #245

## Type of Change

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Refactor

## Checklist

- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [x] I have documented my changes in the code or documentation
[...]

Issue #245, opened 2026-08-25
gang member access binds too loosely: `p.x = p.x + 1` and `a.x + a.y` misparse and crash


leo-aa88:
## Summary

Two common `gang` (struct) patterns fail on current `main`:

1. **Read-modify-write:** `p.x = p.x + 1`
2. **Field + field:** `a.x + a.y`

They look like interpreter bugs, but the **root cause is grammar precedence**: `DOT` (`.`) is declared with the **lowest** operator precedence in `lang.y`, so member access loses to `+`, `=`, `&&`, etc. Expressions are associated incorrectly, then the evaluator crashes or prints a misleading error when it sees a non-struct object.

Related prior report (same family, incomplete diagnosis): #192.

## Repro

```c
gang Point {
    rizz x;
    rizz y;
};

skibidi main {
    gang Point p;
    gang Point a;

    p.x = 5;
    a.x = 3;
    a.y = 4;

    p.x = p.x + 1;           /* ASan SEGV */
    yapping("%d", a.x + a.y); /* Error: Unsupported struct member access expression */

    bussin 0;
}
```

### Actual

**`p.x = p.x + 1`:**
```
Error: Unsupported variable type at line …
ast.c:…: runtime error: load of null pointer of type 'int'
AddressSanitizer: SEGV … in evaluate_expression_int
```

**`a.x + a.y`:**
```
Error: Unsupported struct member access expression at line …
0
```

### Expected

```
6
7
```
(and no sanitizer abort)

## What works today (controls)

| Expression | Result |
| --- | --- |
| `p.x = 5;` | ✅ |
| `yapping("%d", p.x);` | ✅ |
| `bussin p.x;` | ✅ |
| `a.x + 1` | ✅ |
| `(a.x) + (a.y)` | ✅ |
| `p.x = (p.x) + 1;` | ✅ |
| `p.x = (p.x + 1);` | ✅ |
| `s = (p.x);` | ✅ |

Parentheses “fix” every case — strong signal this is **parse association**, not missing `NODE_STRUCT_ACCESS` evaluation (that path already works when the AST is right).

## Failure surface (mapped)

| Code | Symptom | Likely wrong parse |
| --- | --- | --- |
| `a.x + a.y` | `Unsupported struct member access expression` | `(a.x + a).y` |
| `1 + a.x` | same | `(1 + a).x` |
| `a.x - a.y` | same | `(a.x - a).y` |
[...]

PR #355, opened 2026-09-06
fix: bussin in skibidi main stops execution and sets the process exit code (#246)

### LINKED ISSUE #246: `bussin` in `skibidi main` neither stops execution nor sets the process exit code
## Summary

`bussin` is documented as C’s `return`, but inside `skibidi main` it is effectively a no-op for control flow and process status:

1. **Execution continues** after `bussin …;` (later statements still run).
2. **The process always exits 0**, even for `bussin 42;` — `lang.y`’s `main()` hardcodes `return 0`.
3. **The return expression’s value is discarded** for main (`declared_type == NONE`), so nothing is ever stored as an exit code.

User-defined functions already get early return via `setjmp`/`longjmp`. Main was explicitly carved out and never finished.

## Repro

```c
skibidi main {
    yapping("before");
    bussin 42;
    yapping("after");
}
```

```bash
$ ./brainrot repro.brainrot; echo EXIT:$?
before
after
EXIT:0
```

Also broken:

```c
skibidi main {
    edgy (W) {
        yapping("in");
        bussin 3;
    }
    yapping("after if");   /* still runs */
}
```

```c
skibidi main {
    bussin 7;   /* EXIT should be 7, is 0 */
}
```

### Expected

```
before
EXIT:42
```

No `after`. Nested `bussin` inside `edgy` / loops must unwind out of `main` the same way.

Falling off the end of `main` with no `bussin` may keep exiting `0` (reasonable default).

## What works today (contrast)

| Situation | Behavior |
| --- | --- |
| `bussin N;` in a **user-defined** function | Stops the function body (via `LONGJMP`); caller gets `N` |
| `ragequit(42);` in main | Stops the process with exit code 42 (workaround, not `bussin`) |
| `bussin 0;` at end of main in fixtures | “Works” only because nothing follows and exit was going to be 0 anyway |

```c
rizz early() {
    yapping("f-before");
    bussin 9;
    yapping("f-after");   /* does not run */
}
skibidi main {
    rizz v = early();
    yapping("got %d", v); /* got 9 */
    bussin 0;
}
```

So the return *machinery* exists — main simply doesn’t use it.

## Root cause

Three cooperating gaps:

### 1. No jump buffer around `main`

[`handle_return_statement`](https://github.com/Brainrotlang/brainrot/blob/main/ast.c) ends with:

```c
// skibidi main function do not have jump buffer
if (jump_buffer)
{
    exit_scope();
    LONGJMP();
}
```

User calls go through `execute_function_call()` → `PUSH_JUMP_BUFFER()` + `setjmp`.  
`skibidi main` is **not** a registered `Function` and is **not** run that way.

Grammar ([`lang.y`](https://github.com/Brainrotlang/brainrot/blob/main/lang.y)):

```yacc
skibidi_function:
    SKIBIDI MAIN LBRACE statements RBRACE
        { $$ = $4; }   /* bare statement list */
```

`interpret()` just `ast_accept`s that list. `interpreter_visit_statement_list` walks every statement with no early-exit check (`Interpreter.should_return` exists in `interpreter.h` but is never consulted).

### 2. Main’s `bussin` expression is ignored

In `handle_return_statement`, with no function scope, `declared_type` stays `NONE`. The `NONE` / `VAR_VOID` case **does not evaluate** the return expression into `current_return_value.value` — it only sets `has_value = true` (comment acknowledges main has “no declared return type”).

So even if exit-code wiring existed, `bussin 42` never stores `42`.

(Side note: `visitor.c` still `ast_accept`s the return child *before* `visit_return_statement`, so a call like `bussin bump();` can run for side effects via the function-call visitor, then get `free_pending_return_value()`’d — value still never becomes main’s exit code. Related double-eval noise: #225.)

### 3. Host `main()` always returns 0

```c
interpret(root, global_interpreter);
interpreter_free(global_interpreter);
…
return 0;   /* lang.y — ignores current_return_value entirely */
```

Parse/semantic failures correctly `return 1`. Successful runs never consult the program’s `bussin` value.

## Proposed fix

Mirror the user-function return path for `skibidi main`:

1. **Run main under a jump buffer**  
   In `interpret()` (or a dedicated `execute_main(statements)`): `PUSH_JUMP_BUFFER()` + `setjmp`, accept the statement list, `POP_JUMP_BUFFER()` on normal completion. Then `bussin`’s existing `LONGJMP()` unwinds main the same as any other function.

2. **Treat main’s return as `rizz` (int)**  
   When `declared_type == NONE` because we’re in main (no function scope), evaluate the expression with `evaluate_expression_int` (or `evaluate_expression` → int) into `current_return_value` (`type = VAR_INT`, `ivalue = …`, `has_value = true`).  
   Bare `bussin;` (if allowed) → `0`.  
   Keep true `VAR_VOID` user functions ignoring the expression if that’s still desired — don’t conflate them with main.

3. **Propagate exit code** in `lang.y` after `interpret()`:
   - if `current_return_value.has_value` and type is int-like → `return current_return_value.value.ivalue & 0xFF` (or full int; match `ragequit` / POSIX conventions the project already uses);
   - else → `return 0` (fell off end / no `bussin`).

4. **Stop statement lists on return**  
   Jump buffer handles nested `edgy` / loops. Optionally also honor `should_return` in `interpreter_visit_statement_list` as defense in depth — not required if `longjmp` is wired.

5. **Avoid double-executing the return expression**  
   Prefer evaluating only inside `handle_return_statement` for the interpreter (same idea as the native-call pre-visit fix in `interpreter_visit_function_call`). Don’t leave “visit child then ignore value” as the source of truth for main.

6. **Tests** (AGENTS.md): fixtures + `tests/expected_results.json`. The harness already supports exit-only asserts:

   ```json
   "bussin_main_exit_code": "ExitCode:42"
   ```

   and a stdout case:

   ```json
   "bussin_main_stops": "before\n"
   ```

   Cover: trailing code after `bussin`, `bussin` inside `edgy`, nonzero code, fall-off-end → 0. Run `make test` and `make valgrind`.

## Acceptance criteria

- [ ] `bussin N;` in `skibidi main` → process exit status `N` (at least for small positive codes used in tests)
- [ ] Statements after `bussin` in main do not run (including after `bussin` inside `edgy` / loop bodies)
- [ ] User-defined function `bussin` behavior unchanged
- [ ] No `bussin` / fall off end → exit `0`
- [ ] New fixtures + `ExitCode:` / stdout expectations; `make test` && `make valgrind` pass
- [ ] Docs (`docs/the-brainrot-programming-language.md` §7.5) remain accurate: `bussin 0` ≡ `return 0` for main’s exit status

## Out of scope / related

- `#225` — user-defined calls in expression position executing twice (can interact with `bussin f();` in main; fix return path carefully)
- Changing `ragequit` (already correct)
- Making `skibidi main` a first-class typed `Function` in the hashmap (optional refactor; not required if interpret() gets a dedicated main frame)

## Workaround (today)

Use `ragequit(code);` when you need a real nonzero exit and to stop the process. That is not a substitute for fixing `bussin` — every example and fixture that ends with `bussin 0;` is teaching the wrong mental model until this works.

## Environment

Reproduced on current `main` with the sanitizer build (`make && ./brainrot …`).

leo-aa88:
## Description

`bussin` is documented as C's `return`, but inside `skibidi main` it was
effectively a no-op: execution continued after it, its value was discarded, and
the process always exited 0 (`lang.y`'s `main()` hardcoded `return 0`).

```c
skibidi main {
    yapping("before");
    bussin 42;
    yapping("after");   // still ran
}                       // exit was 0, not 42
```

User-defined functions already return early via `setjmp`/`longjmp`; `main` was
explicitly carved out and never finished. Three cooperating fixes, mirroring the
user-function return path:

1. **`interpret()` runs `main` in its own frame** — a function-marked scope for
   `handle_return_statement()`'s scope walk to stop at, and a function jump
   buffer for its existing `LONGJMP` to land in. `main` isn't a registered
   `Function`, so the frame's `function_name` stays empty → `get_function()`
   returns `NULL` and `declared_type` stays `NONE`, which is exactly what routes
   `main`'s return through the new exit-code path.

2. **`handle_return_statement()`'s `NONE` arm** (reached only for `main`) now
   evaluates the operand as an int and records it in a new `g_program_exit_code`
   (bare `bussin;` → 0). It first calls `free_pending_return_value()` so
   overwriting the shared `current_return_value` slot can't leak an owned
   struct/string blob a prior unconsumed call left there — caught via ASan on
   `struct_array_field_of_structs` during development.

3. **`lang.y`'s `main()` returns `g_program_exit_code & 0xFF`** after
   `interpret()` (POSIX-style truncation, matching `ragequit`). A dedicated
   global — not `current_return_value`, which a nested call overwrites — so only
   `main`'s own `bussin` sets the exit code.

### Result (all acceptance criteria)

- `bussin N;` in `main` → process exit status `N`.
- Statements after `bussin` don't run — including a `bussin` inside `edgy` / a
  loop, which unwinds out of `main`.
- No `bussin` / fall off the end → exit `0`.
- User-defined function `bussin` unchanged.

## Related Issue

Fixes #246

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement
- [ ] Refactor

## Checklist

### Tests

- `bussin_main_exit_code` → `ExitCode:42` (nonzero exit).
- `bussin_main_stops` → `before` (statements after `bussin` don't run; uses `bussin 0` so it can assert on stdout).
- `bussin_main_edgy_exit` → `ExitCode:3` (`bussin` nested in `edgy` sets the code and stops).
- `return_from_inside_loop_in_main` — **renamed** from `return_from_inside_loop_in_main_fail`, which pinned the OLD "No scope to exit" behaviour and whose own comment said *"when [#246] lands, this fixture is expected to change."* It now asserts clean termination from inside a loop.
- Docs §7.5 updated to describe `main`'s exit status.

`make test` → **509 passed**; full `make valgrind` sweep → exit 0 (a struct-blob leak surfaced during development is fixed by the `free_pending_return_value()` call above). `make format-check` clean. `make cppcheck` couldn't run locally (this box has cppcheck 2.7 < the required 2.13); the changes are left for CI's `static-analysis` job.


interpreter.c
```diff
@@ -89,8 +89,25 @@ void interpret(ASTNode *root, Interpreter *interp)
...
+    /* Run skibidi main inside its own function frame so `bussin N;` unwinds out
+       of main the same way it does out of any user-defined function (#246): a
+       function-marked scope for handle_return_statement()'s scope walk to stop
+       at, and a function jump buffer for its LONGJMP to land in. Without this,
+       main's `bussin` fell through and statements after it still ran. main is
+       not a registered Function, so the frame's function_name stays empty --
+       get_function() returns NULL for it and declared_type stays NONE, which is
+       exactly what routes main's return through the exit-code path above. */
+    enter_scope();
+    current_scope->is_function_scope = true;
```
leo-aa88, line 101: MINOR (stale comment, non-blocking): adding this function frame to `main` invalidates the long `has_function_frame` explanatory comment in `ast.c` (~7920–7945). It still asserts the opposite of what now happens: "in `main` it does not [longjmp]", "main's body is not run through [execute_function_call]... so nothing below a `bussin` in `main` is ever is_function", "main keeps the one diagnostic it had before", and "even a plain `bussin 3;` there exits 0 rather than 3, so the value is not plumbed. That is a separate, larger change... deliberately not folded in here." After this PR, main *does* have an is_function frame, the LONGJMP *does* fire, and the value *is* plumbed. Worth updating that block so it doesn't mislead future readers. Code is correct — verified live.

leo-aa88 (commented):
# APPROVE

**LGTM** with one MINOR non-blocking note (posted inline). This is the meatiest PR in the batch — control-flow surgery on `interpret()` + `handle_return_statement()`. I checked out the head in an isolated worktree, traced the unwinding end-to-end, and stress-tested exit codes, non-evaluation, function-return non-regression, and leaks under ASan + valgrind. The behavior is correct across everything I threw at it.

## Findings

### MINOR — stale explanatory comment (`ast.c` ~7920–7945), non-blocking
The `has_function_frame` comment block now contradicts the code this PR ships: it states `main` has no is_function frame, never longjmps, "keeps the one diagnostic it had before", and that plumbing `bussin`'s value is "a separate, larger change... deliberately not folded in here." This PR *is* that change — `interpret()` now gives `main` a function frame, so the drain/LONGJMP path is taken and the value is plumbed. Code is correct; only the comment misleads. Details inline on `interpreter.c`.

## Verified end-to-end (worktree @ e45fe01)
- **CONFIRMED — routing is airtight.** `declared_type` is derived from the function scope's registered `Function`; `main`'s frame has an empty `function_name` → `get_function()` → NULL → `NONE` (exit-code arm). A real void function resolves to `VAR_VOID`, which falls through to the shared bare-call logic but is gated out of the exit-code block by `if (declared_type == NONE)`. No misrouting.
- **CONFIRMED — exit status.** `bussin 42;`→42, `bussin 3;` (nested in `edgy`)→3, `bussin 0;`/fall-off-end→0. Truncation `& 0xFF` is correct across the range: 200→200, 255→255, 256→0, 300→44, 512→0, `-1`→255 (`yapping` shows the value still evaluates to 300 while the process exits 44).
- **CONFIRMED — execution stops.** Statements after `bussin` don't run, including from inside `edgy` and a `flex` loop (the renamed `return_from_inside_loop_in_main` fixture, correctly replacing the old `_fail` one whose own comment predicted this change). Bare `bussin;` stops and exits 0.
- **CONFIRMED — no regression to user functions.** `bussin a+b;` in a helper returns to the caller (r=42), doesn't exit the process, and code after it is unreachable — unchanged.
- **CONFIRMED — leak-clean.** All 4 fixtures and a crafted struct-returning-call-unconsumed-then-`bussin` case (the scenario `free_pending_return_value()` targets) are valgrind-clean: 0 def / 0 indirect lost, 0 errors. `pytest -k "bussin_main or return_from_inside_loop"` → 5 passed. Scope/jump-buffer push/pop balances on both the fall-through and longjmp paths.
- Docs §7.5 correctly updated; `expected_results.json` valid.

## VERDICT

Correct, well-tested, and it closes #246's acceptance criteria exactly — process exit status set, execution halted (loops/`edgy` included), user-function returns untouched, and the struct-blob leak fixed. The three cooperating changes (main frame, NONE arm, `lang.y` return) are coherent and the routing invariant holds. The only thing I'd change before merge is refreshing the now-contradictory `ast.c` comment — a documentation nit, not a correctness issue. Deserves to merge. (Posting as COMMENT rather than a formal approval only because the reviewer account is also the PR author — GitHub rejects self-approval.)

leo-aa88 (commented):
# APPROVE

**LGTM.** Re-review of the new head `55be47c` (was `e45fe01`).

The only change since my previous review is the follow-up commit `docs(ast): refresh has_function_frame comment now that main has a frame` — which addresses exactly the MINOR finding I raised. I verified it is **comment-only**: stripping all comments from `ast.c` at both SHAs leaves byte-identical code (the sole diffs are a shifted `#line` marker and two blank lines). No functional delta from the version I already traced and stress-tested (exit codes + truncation, execution-stop, user-function non-regression, and valgrind-clean leak paths all still hold).

## Findings

None. The previously-stale `has_function_frame` block (`ast.c` ~7923–7960) now correctly states that `main` runs under its own is_function frame since #246, that `has_function_frame` is true in `main` too, and that the guard is retained as defense-in-depth — and the old `// skibidi main function do not have jump buffer` line is likewise corrected.

## VERDICT

Same strong APPROVE as before, with the documentation nit now fixed. Deserves to merge. (Posting as COMMENT rather than a formal approval only because the reviewer account is also the PR author — GitHub rejects self-approval.)

leo-aa88 (commented):
# APPROVE

**LGTM.** Re-review of new head `0cfdf00` (was `55be47c`).

The only change is a **merge commit** integrating current `origin/main` (shared `expected_results.json` conflict; #350 also touched `ast.c`). The content commits `e45fe01` + `55be47c` I reviewed (including the comment refresh that resolved my earlier MINOR note) are unchanged.

Verified the merge is sound:
- No conflict markers; `ast.c` changes here (`handle_return_statement` NONE arm) don't overlap #350's `register_struct_def` guard.
- Merged `expected_results.json` valid; net contribution is this PR's own `bussin_*` / `return_from_inside_loop_in_main` keys, nothing clobbered.
- Rebuilt from the merge head: `bussin_main_exit_code`→exit 42, `bussin_main_stops`→`before`/exit 0, `bussin_main_edgy_exit`→exit 3, `return_from_inside_loop_in_main`→`before`/exit 0 — identical to my original review.

## VERDICT

Same APPROVE — clean, functionally-verified merge-from-main. Deserves to merge. (COMMENT, not formal approval, only because the reviewer account is also the PR author.)

leo-aa88 (commented):
# APPROVE

**LGTM.** Re-review of new head `5ef956d` (was `0cfdf00`).

Only change is another **merge commit** integrating `origin/main`. The content commits `e45fe01` + `55be47c` I reviewed are unchanged.

Verified: no conflict markers; merged `expected_results.json` valid with this PR's `bussin_*`/`return_from_inside_loop_in_main` keys intact. Rebuilt from the merge head — `bussin_main_exit_code`→42, `bussin_main_edgy_exit`→3, `return_from_inside_loop_in_main`→`before`/0 — identical to my original review. (This branch hasn't pulled #352 yet, which is fine; its own `ast.c` change is `handle_return_statement`, non-overlapping with anything in main.)

## VERDICT

Same APPROVE — clean, functionally-verified merge-from-main. Deserves to merge. (COMMENT, not formal approval, only because the reviewer account is also the PR author.)

leo-aa88 (commented):
# APPROVE

**LGTM.** Re-review of head `727562e` — the conflict-resolution merge I just pushed to bring the branch up to current `origin/main` (which now includes #352). The content commits `e45fe01` + `55be47c` I reviewed are unchanged.

This merge needed care because #245's/#246's rename removes the `return_from_inside_loop_in_main_fail` JSON key while adding `return_from_inside_loop_in_main`. I did a proper 3-way merge:
- **removed** `return_from_inside_loop_in_main_fail` (its fixture file is deleted by this PR — a naive union would have left a dangling test entry),
- **added** `return_from_inside_loop_in_main` + the three `bussin_main_*` keys,
- **kept** main's `sizeof_struct_call_and_deref`.

Verified: `ast.c` auto-merged cleanly (this PR's `handle_return_statement` change and #352's `get_struct_def_for_expression` case are in different functions); rebuilt the merged tree and confirmed **all four #355 fixtures pass** (exit 42 / `before` / exit 3 / `before`) **and #352's `sizeof_struct_call_and_deref` still returns `8\n8\n8\n8\n8\n8`** — no interaction. Merged JSON valid; deleted-fixture entry confirmed gone.

## VERDICT

Same APPROVE — mergeable now, content and behavior unchanged, merge (including the fixture-rename key deletion) verified by rebuild. (COMMENT, not formal approval, only because the reviewer account is also the PR author.)

leo-aa88 (commented):
# APPROVE

**LGTM.** Re-review of head `b777c7b` (was `727562e`) — a merge commit bringing the branch up to current `origin/main` (now includes #352 and #357). The content commits `e45fe01` + `55be47c` I reviewed are unchanged.

Verified by rebuild from this merge head: no conflict markers; `ast.c` (this PR's `handle_return_statement` + #352's `get_struct_def_for_expression`) and `lib/input.c` (#357) all coexist; **all four #355 fixtures pass** (exit 42 / `before` / exit 3 / `before`), **#352's `sizeof_struct_call_and_deref` still returns `8×6`**, and the renamed-away `return_from_inside_loop_in_main_fail` fixture/JSON key is correctly gone. Merged `expected_results.json` valid (440 keys); `MERGEABLE`.

## VERDICT

Same APPROVE — merge-only, content and behavior unchanged, verified by rebuild. (COMMENT, not formal approval, only because the reviewer account is also the PR author.)

leo-aa88 (commented):
# APPROVE

**LGTM.** Re-review of head `16d1873` (was `b777c7b`) — a merge commit bringing the branch up to current `origin/main` (after #354 landed). The content commits `e45fe01` + `55be47c` I reviewed and rebuilt are unchanged.

Verified: no conflict markers; merged `tests/expected_results.json` valid (441 keys) with this PR's `bussin_main_*` / `return_from_inside_loop_in_main` keys present and the renamed-away `return_from_inside_loop_in_main_fail` key correctly absent. The just-merged #354 is test-only (a new fixture), with no overlap on this PR's `ast.c`/`interpreter.c`/`lang.y` — so the code is byte-identical to the tree I already built and ran (all four fixtures passing). `MERGEABLE`.

## VERDICT

Same APPROVE — merge-only, content and behavior unchanged. This is now the last open PR in the batch and is mergeable. (COMMENT, not formal approval, only because the reviewer account is also the PR author.)

PR #356, opened 2026-09-06
fix(stdrot): clamp print buffer offset to stop snprintf-return overflow (#269)

### LINKED ISSUE #269: [Bug]: yapping/yappin/baka stack-buffer-overflow on long formatted output (snprintf return misused)
### Description

The print-formatting helpers `process_yapping_format()` (`stdrot/yapping.c`, used by `yapping`/`yappin`) and `process_baka_format()` (`stdrot/baka.c`, used by `baka`) accumulate `snprintf()` **return values** into `buffer_offset`:

```c
buffer_offset += snprintf(buffer + buffer_offset,
                          sizeof(buffer) - buffer_offset,
                          specifier, ...);
```

`snprintf` returns the number of bytes it *would* have written, not the number actually written. When a formatted argument (e.g. a long `%s` string, or a wide `%999d`) exceeds the space remaining in the fixed 1024-byte stack `buffer`, `buffer_offset` is advanced **past** `sizeof(buffer)`. The `while` guard then stops the loop, but the unconditional terminator write at the end of the function

```c
buffer[buffer_offset] = '\0';   // yapping.c:170, baka.c:155
```

writes out of bounds — a **stack-buffer-overflow**. Additionally, if more args follow, the next iteration computes `sizeof(buffer) - buffer_offset`, which underflows to a huge `size_t` passed as `snprintf`'s size argument.

This is not caught by CI because `libstdrot.so` is built **without** `-fsanitize=address` (only the main `brainrot` binary is instrumented), so ASan never sees the overflowing write.

### Minimal reproduction code

```c
skibidi main {
    🚽 a string literal (or any value) whose formatted form exceeds 1024 bytes
    yapping("%s", "AAAAA…(≥1024 A's)…AAAAA");
    bussin 0;
}
```

### Expected behavior

Output is written (truncated if necessary) with no out-of-bounds memory access; `buffer_offset` is clamped to `sizeof(buffer) - 1` before the terminator write.

### Actual behavior

```shell
# Extracting the exact yapping.c:170 write path and compiling it with ASan:
==ERROR: AddressSanitizer: stack-buffer-overflow on address 0x... 
WRITE of size 1 at 0x... thread T0
    #0 ... in main ov.c:10   # buffer[buffer_offset] = '\0';
  This frame has 2 object(s):
    [48, 1072) 'buffer'
    ...
```

`buffer_offset` reaches 1100 for a 1099-char `%s` argument, and `buffer[1100] = '\0'` writes 76 bytes past the 1024-byte buffer. Running through `./brainrot` "succeeds" (exit 0) only because `libstdrot.so` is not ASan-instrumented; the corruption is silent.

### Brainrot version / commit

### Operating system

WSL2 (Ubuntu), x86-64

### Additional context

Fix: after each `snprintf`, clamp, e.g.

```c
int n = snprintf(buffer + buffer_offset, sizeof(buffer) - buffer_offset, ...);
if (n < 0) break;
buffer_offset += n;
if (buffer_offset >= (int)sizeof(buffer))
{
    buffer_offset = sizeof(buffer) - 1;
    break;
}
```

Both `stdrot/yapping.c` and `stdrot/baka.c` need the same fix. Separately, consider building `libstdrot.so` with the same sanitizer flags as the main binary so this class of bug is caught by `make test`/`make valgrind`.

leo-aa88:
## Description

`process_baka_format()` (`stdrot/baka.c`) and `stdrot_format_to_stream()`
(`stdrot/yapping.c`, the shared `yapping`/`yappin`/`yapto` formatter) accumulate
`snprintf()`'s **return value** into `buffer_offset`:

```c
buffer_offset += snprintf(buffer + buffer_offset,
                          sizeof(buffer) - buffer_offset, ...);
```

`snprintf` returns the length it *would* have written, not what it wrote, so a
formatted argument longer than the space left in the fixed 1024-byte stack
`buffer` advances `buffer_offset` **past** `sizeof(buffer)`. `snprintf` itself
never overflows (it honors its size arg) and the loop guard stops re-entry, but
the unconditional terminator write `buffer[buffer_offset] = '\0'` then writes out
of bounds — a **stack-buffer overflow** (e.g. `buffer_offset` = 1099 →
`buffer[1099]` on a 1024-byte buffer).

### Fix

Clamp `buffer_offset` to `sizeof(buffer) - 1` before the terminator write in both
functions. Output is truncated to what fit; nothing else changes.

### Testing — why a host test

This class is invisible to the existing harness:
- `libstdrot.so` is built **without** sanitizers, so `make test`'s ASan build
  never sees the write;
- **Valgrind does not detect stack-buffer overflows** (only ASan's redzones do);
- a `test_cases/*.brainrot` fixture can't tell fixed from broken — `fprintf`
  stops at `snprintf`'s own NUL at index 1023, so stdout is identical either way.

So I added `tests/print/print_overflow_check.c`, which `#include`s
`stdrot/yapping.c` and `stdrot/baka.c` compiled **with** `-fsanitize=address` and
drives both formatters with an over-long argument (baka's formatter is `static`,
so it's reached through `stdrot_baka` — hence the unity include). It's wired into
`make test` via a new `print-overflow-check` target, kept out of `tests/stdrot/`
(which is globbed into `libstdrot.so`).

**Verified by mutation:** reverting either clamp makes ASan abort the test with
`stack-buffer-overflow ... in stdrot_format_to_stream`; with the fix it exits
clean. The target's flags omit `-Werror`/`-Wpedantic` because the
`STDROT_EXPORT_*` registration macros expand to file-scope compound-literal
initializers `-pedantic` rejects (a GNU extension the production, non-pedantic
libstdrot build relies on) — ASan detection is the point, not warning coverage.

> The issue also suggests, separately, building `libstdrot.so` with sanitizers so
> this whole class is caught by `make test`. That's a larger, riskier build
> change (the tree deliberately keeps libstdrot uninstrumented); this PR closes
> the reported overflow and adds a targeted ASan guard for it instead.

## Related Issue

Fixes #269

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement
- [ ] Refactor

## Checklist

> `make test` → **506 passed** (plus the new `print-overflow-check` target it now
> depends on); full `make valgrind` sweep → exit 0; `make format-check` clean.
> `make cppcheck` couldn't run locally (this box has 2.7 < the required 2.13); the
> stdrot changes are a two-line clamp, left for CI's `static-analysis` job.

### DIFF
--- a/stdrot/yapping.c
+++ b/stdrot/yapping.c
@@ -215,6 +215,17 @@ void stdrot_format_to_stream(FILE *out, const char *format,
         }
     }
 
+    /* snprintf() returns the length it WOULD have written, not what it
+       actually wrote, so a formatted argument longer than the space left in
+       `buffer` advances buffer_offset past sizeof(buffer) (#269). snprintf
+       itself never overflows -- it honors the size arg -- but the terminator
+       write below would, and the loop guard above only stops re-entry, it
+       doesn't clamp. Clamp into range so `buffer[buffer_offset]` stays in
+       bounds; output is simply truncated to what fit. */
+    if (buffer_offset >= (int)sizeof(buffer))
+    {
+        buffer_offset = (int)sizeof(buffer) - 1;
+    }
     buffer[buffer_offset] = '\0';
     if (add_newline)
     {
--- a/stdrot/yapping.c
+++ b/stdrot/yapping.c
@@ -221,8 +221,15 @@ void stdrot_format_to_stream(FILE *out, const char *format,
        itself never overflows -- it honors the size arg -- but the terminator
        write below would, and the loop guard above only stops re-entry, it
        doesn't clamp. Clamp into range so `buffer[buffer_offset]` stays in
-       bounds; output is simply truncated to what fit. */
-    if (buffer_offset >= (int)sizeof(buffer))
+       bounds; output is simply truncated to what fit. The `< 0` branch guards
+       the rarer case where snprintf returned negative on an output/encoding
+       error, which would otherwise leave buffer_offset negative and underflow
+       the terminator write. */
+    if (buffer_offset < 0)
+    {
+        buffer_offset = 0;
+    }
+    else if (buffer_offset >= (int)sizeof(buffer))
     {
         buffer_offset = (int)sizeof(buffer) - 1;
     }
--- a/.gitignore
+++ b/.gitignore
@@ -24,3 +24,4 @@ tests/brainrot-test.wasm
 tests/brainrot-test.mjs
 tests/abi/struct_layout_abi_check
 tests/arena/arena_check
+tests/print/print_overflow_check
--- a/Makefile
+++ b/Makefile
@@ -510,6 +510,25 @@ $(ARENA_CHECK_BIN): tests/arena/arena_check.c $(SRC_DIR)/arena.c $(SRC_DIR)/mem.
 arena-check: $(ARENA_CHECK_BIN) ## Run the arena allocator unit tests (host reset/reuse under sanitizers). Clean memory, no cap.
 	./$(ARENA_CHECK_BIN)
 
+# Host ASan test for #269: the print formatters' fixed 1024-byte buffer must not
+# overflow on a formatted argument longer than it. It #includes stdrot/yapping.c
+# and stdrot/baka.c so their formatters (baka's is static) are compiled WITH
+# sanitizers -- unlike libstdrot.so, whose lack of instrumentation is exactly why
+# `make test`'s ASan build never caught this stack overflow (and Valgrind can't).
+# Flags deliberately omit -Werror/-Wpedantic: the STDROT_EXPORT_* registration
+# macros expand to file-scope compound-literal initializers that -pedantic
+# rejects (a GNU extension the production, non-pedantic libstdrot build relies
+# on). ASan is the point here, not warning coverage.
+PRINT_OVERFLOW_BIN := tests/print/print_overflow_check
+PRINT_OVERFLOW_CFLAGS := -Wall -Wextra $(SANITIZER_FLAGS) -fno-omit-frame-pointer -g -O1
+
+$(PRINT_OVERFLOW_BIN): tests/print/print_overflow_check.c $(STDROT_DIR)/yapping.c $(STDROT_DIR)/baka.c
+	$(CC) $(PRINT_OVERFLOW_CFLAGS) -I$(STDROT_DIR) -o $@ $< -lm
+
+.PHONY: print-overflow-check
+print-overflow-check: $(PRINT_OVERFLOW_BIN) ## Run the stdrot print-formatter overflow test (#269) under ASan.
+	./$(PRINT_OVERFLOW_BIN)
+
 # Main executable build
 $(TARGET): $(ALL_SRCS) $(STDROT_LIB) $(STDROT_ABI_HDR)
 	$(CC) $(CFLAGS) -o $@ $(ALL_SRCS) $(LDFLAGS)
@@ -560,7 +579,7 @@ $(FLEX_OUTPUT): lang.l
 
 # Run tests
 .PHONY: test
-test: $(TARGET) $(TEST_STDROT_LIB) badnatives nativemodules old-abi-sim abi-check arena-check ## Build, then run the pytest suite. Huggy Wuggy approves.
+test: $(TARGET) $(TEST_STDROT_LIB) badnatives nativemodules old-abi-sim abi-check arena-check print-overflow-check ## Build, then run the pytest suite. Huggy Wuggy approves.
 	STDROT_LIB_PATH=$(CURDIR)/$(TEST_STDROT_LIB) $(PYTHON) -m pytest -v
 	@echo "Tests ran bussin', no cap."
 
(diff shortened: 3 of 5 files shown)


stdrot/yapping.c
```diff
@@ -215,6 +215,17 @@ void stdrot_format_to_stream(FILE *out, const char *format,
...
     }
 
+    /* snprintf() returns the length it WOULD have written, not what it
+       actually wrote, so a formatted argument longer than the space left in
+       `buffer` advances buffer_offset past sizeof(buffer) (#269). snprintf
+       itself never overflows -- it honors the size arg -- but the terminator
+       write below would, and the loop guard above only stops re-entry, it
+       doesn't clamp. Clamp into range so `buffer[buffer_offset]` stays in
+       bounds; output is simply truncated to what fit. */
+    if (buffer_offset >= (int)sizeof(buffer))
```
leo-aa88: MINOR (defensive, non-blocking): this clamps only the HIGH side. `buffer_offset` is a signed int accumulating `snprintf`'s return, which is negative on a (rare) output/encoding error — that would leave `buffer_offset` < 0 and `buffer[buffer_offset] = '\0'` would underflow. Not reachable for the %s/%d/%f/%c specifiers here, and pre-existing (not introduced by this PR), so out of scope for #269 — but a `buffer_offset < 0` guard (or checking each `snprintf` return for <0) would make the clamp total. Same applies to the identical clamp in baka.c.

leo-aa88 (commented):
# APPROVE

**LGTM** with one MINOR non-blocking note (posted inline). Adversarial pass on this stdrot overflow fix: checked out the head in an isolated worktree, read the accumulation logic, and proved the fix + its ASan test by mutation. The write-out-of-bounds is real and the clamp closes it.

## Findings

### MINOR — clamp guards only the high side (`stdrot/yapping.c:225`, same in `baka.c`), non-blocking
`buffer_offset` is a signed `int` accumulating `snprintf`'s return, which is negative on a rare output/encoding error → a negative index underflow the `>= sizeof` clamp doesn't catch. Unreachable for the `%s/%d/%f/%c` specifiers used, and pre-existing, so out of scope for #269 — a `< 0` guard would just make it total. Details inline.

## Verified end-to-end (worktree @ 25d9d09)
- **CONFIRMED — the overflow is real and the clamp fixes it.** `buffer_offset += snprintf(...)` accumulates the would-be length; the loop guard (`< 1023`) keeps every `snprintf` size arg valid, so the sole OOB was the terminator write `buffer[buffer_offset]` once the last `+=` pushed offset past 1024. Clamping to `sizeof-1` puts it back in bounds; fitting output is untouched (clamp never triggers below 1024).
- **CONFIRMED — the test has teeth (mutation).** With the fix, `make print-overflow-check` prints "buffer stayed in bounds", exit 0. Reverting the `yapping.c` clamp makes it abort: `AddressSanitizer: stack-buffer-overflow ... stdrot/yapping.c:226 in stdrot_format_to_stream`. So it genuinely distinguishes fixed from broken.
- **CONFIRMED — the host-test rationale holds.** libstdrot.so is built without sanitizers (so `make test`'s ASan build never sees this), Valgrind doesn't catch stack overflows, and a `.brainrot` fixture can't tell fixed from broken (fprintf stops at snprintf's own NUL). The unity-include of `baka.c`/`yapping.c` under `-fsanitize=address` is the right way to reach the static baka formatter; omitting `-Werror/-Wpedantic` is justified (the `STDROT_EXPORT_*` compound-literal macros are a GNU extension `-pedantic` rejects).
- **CONFIRMED — integration.** Wired into `make test` (`Makefile:582`), `.gitignore` + `clean` updated. Normal print unaffected (`hello 42`); a 1200-char print through the real interpreter exits 0 and truncates cleanly to 1024 bytes, no crash.

## VERDICT

Correct two-line memory-safety fix for a genuine stack-buffer-overflow, with a well-designed, mutation-verified ASan guard reaching code the normal harness structurally cannot. Scoping out the "sanitize all of libstdrot" idea is the right call for this PR. The only thing I'd add is a low-side clamp for completeness — a defensive nicety, not a blocker. Deserves to merge. (Posting as COMMENT rather than a formal approval only because the reviewer account is also the PR author — GitHub rejects self-approval.)

leo-aa88 (commented):
# APPROVE

**LGTM.** Re-review of the new head `64598ec` (was `25d9d09`).

The only change since my last review is the follow-up commit `harden(stdrot): also clamp a negative snprintf return in the print buffer` — which addresses exactly the MINOR low-side note I raised. Both `stdrot/yapping.c` and `stdrot/baka.c` now guard the full range: `buffer_offset < 0 → 0`, `>= sizeof → sizeof-1`, else unchanged. The clamp is now total, so a negative `snprintf` return can no longer underflow the terminator write.

## Findings

None. Verified at `64598ec`: builds clean, `make print-overflow-check` still passes ("buffer stayed in bounds"), the high-side overflow guard is untouched (still mutation-proven from the prior review), and normal formatted output is unaffected (`hello 7`).

## VERDICT

Same strong APPROVE, with the defensive nicety now folded in — the clamp handles both the high-side snprintf-return overflow (the #269 report) and the rare negative-return underflow. Deserves to merge. (Posting as COMMENT rather than a formal approval only because the reviewer account is also the PR author — GitHub rejects self-approval.)

PR #358, opened 2026-09-07
feat: implement grind (continue) — tokenized but never wired into the grammar (#274)

### LINKED ISSUE #274: [Bug]: grind (continue) is documented as implemented but is a parse error in every loop

leo-aa88:
## Description

`grind` mapped to a `CONTINUE` token that **no grammar production ever used**, so
every `grind` — in `flex` (for), `goon` (while), or `mewing` (do-while) — was a
hard parse error (`unexpected CONTINUE`), even though the README lists it as ✅
implemented.

```c
flex (rizz i = 0; i < 5; i = i + 1) {
    edgy (i == 2) { grind; }   // was: Parsing failed
    yapping("%d", i);          // now prints 0 1 3 4
}
```

### Approach

Unlike `bruh` (break), which `longjmp`s straight out of the loop, `continue` must
**resume** the loop. A longjmp would abandon the body's scopes and leave them
accumulating across iterations, so instead `grind` raises a `continue_requested`
flag that `interpreter_visit_statement_list()` honors: it stops running the rest
of the current list, so control unwinds through **normal returns** (scopes stay
balanced) up to the enclosing loop, which clears the flag and runs its next
iteration.

- In a `flex` loop the **increment still runs** first, matching C's `continue`.
- An inner loop clears the flag before the outer body resumes, so a **nested**
  `grind` leaves the outer loop alone.
- Coexists with `bruh` in the same body.

Touch points mirror `NODE_BREAK_STATEMENT` exactly: the token already existed;
this adds the grammar production (`continue_statement: CONTINUE`),
`NODE_CONTINUE_STATEMENT` + `create_continue_node()`, the visitor dispatch/hook,
the interpreter handler, and the flag checks in `statement_list` and the three
loop visitors. No new bison conflicts.

## Related Issue

Fixes #274

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement
- [ ] Refactor

## Checklist

### Test

`test_cases/continue_grind.brainrot` exercises `grind` in `flex`/`goon`/`mewing`,
statements after `grind` in the same block (skipped), nesting (inner
`grind`/`bruh` leave the outer loop alone), and the increment-still-runs
behavior. `make test` → **517 passed**; the full `make valgrind` sweep → exit 0
(no scope leak from the flag-based unwind); `make format-check` clean. `make
cppcheck` left to CI (local box has 2.7 < the required 2.13).


interpreter.c
```diff
@@ -905,6 +930,11 @@ void interpreter_visit_statement_list(Visitor *self, ASTNode *node)
     {
         if (stmt->statement)
             interpreter_accept_or_execute_call(stmt->statement, self);
+        /* `grind` (continue) stops the rest of this list and every enclosing
+           one, so control unwinds up to the loop that clears the flag (#274).
+         */
+        if (continue_requested)
+            break;
```
leo-aa88, line 954: MAJOR (#274): `continue_requested` is a global flag that ONLY a loop clears, and nothing guards `grind` used outside a loop. Two concrete failures result:

1. **Silent program truncation.** `grind` with no enclosing loop (top level, or inside `edgy` with no loop) sets the flag, this `break` unwinds every statement list up to the top, and nothing ever clears it — so the rest of the program is silently skipped. `skibidi main { yapping("A"); grind; yapping("B"); }` prints only `A`, exit 0, no diagnostic.
2. **Cross-function control-flow leak.** A `grind` that runs inside a called function which has no loop of its own leaves the flag set on return, so the CALLER's loop treats it as a continue. `helper(){ yapping("start"); grind; yapping("end"); }` called in `main`'s loop prints `start` each iteration, never `end`, and skips everything after the call in the caller — the callee hijacks the caller's loop.

Both are the classic 'continue not within a loop' case, which C rejects at compile time. Note `bruh` (break) outside a loop at least fails LOUDLY here (`Error: No jump buffer available`, exit 1); `grind` fails silently, which is worse.

Suggested fix: a semantic-analysis check that `grind` (and ideally `bruh`) appears only within a loop, emitting a clear error otherwise. (Saving/restoring `continue_requested` across function-call boundaries would additionally contain the cross-function leak, but the semantic guard is the real fix and matches C.) The in-loop happy path is correct and valgrind-clean; this is only about the unguarded out-of-loop case.

leo-aa88 (commented):
# REQUEST CHANGES

Adversarial pass on `grind` (continue). I built the head in an isolated worktree and traced the flag mechanism end-to-end. The intended in-loop behavior is correct and leak-clean, but the flag-based design is missing the "continue outside a loop" guard, and that produces silent wrong execution — including cross-function corruption. One MAJOR issue to fix before merge.

## Findings

### MAJOR — `grind` outside a loop is unguarded: silent truncation + cross-function leak
`continue_requested` is a process-global flag cleared only by a loop visitor. Nothing rejects or contains a `grind` that executes without an enclosing loop, so:

- **CONFIRMED — silent program truncation.** `skibidi main { yapping("A"); grind; yapping("B"); bussin 0; }` prints only `A` and exits 0 with no diagnostic — the flag is set, every statement list unwinds, and nothing clears it. Same for `grind` inside an `edgy` that isn't in a loop.
- **CONFIRMED — cross-function control-flow leak.** A helper with a bare `grind` (no loop of its own), called from `main`'s loop, hijacks the caller's loop: `helper(){ yapping("start"); grind; yapping("end"); }` in a 3-iteration loop prints `start` ×3, never `end`, and skips the code after `helper()` in the caller. A function's internal `grind` silently alters the caller's control flow.

C rejects `continue` outside a loop at compile time precisely to prevent this. Note the asymmetry with the existing `bruh` (break): `bruh` outside a loop at least fails loudly (`Error: No jump buffer available`, exit 1); `grind` fails silently, which is strictly worse.

**Scope of the bug (verified):** the intended case is correct. A `grind` inside a loop in the *same* function works, nesting works (inner `grind`/`bruh` leave the outer loop alone), and a helper with its *own* loop+`grind` called from a loop behaves correctly (its loop clears the flag before returning). The defect is strictly `grind` with no enclosing loop in its frame.

**Suggested fix:** add a semantic-analysis check that `grind` (and ideally `bruh`) appears only within a loop, emitting a clear diagnostic otherwise — matching C and the project's other parse/semantic rejections. (Optionally also save/restore `continue_requested` across function-call boundaries to contain the leak defensively, but the semantic guard is the primary fix.) A `test_cases/*_fail.brainrot` fixture pinning the out-of-loop error should accompany it.

## What's correct (CONFIRMED)
- The shipped `continue_grind.brainrot` fixture produces exactly its expected output across `flex`/`goon`/`mewing`, statements-after-`grind` skipped, `for`-increment-still-runs, and nesting with `bruh`.
- Scopes stay balanced — full `make valgrind` on the fixture is clean (0 lost, 0 errors), validating the "unwind via normal returns, not longjmp" approach.
- The wiring mirrors `NODE_BREAK_STATEMENT` cleanly (token already existed; grammar production, AST node, visitor hook, interpreter handler) with no new bison conflicts.

## VERDICT

The feature is 90% there and the design (flag + normal unwind) is sound for the in-loop case. But shipping it without a guard means `grind` in the wrong place silently truncates the program and can corrupt an innocent caller's control flow with no error — a soundness footgun C explicitly prevents. Please add the out-of-loop diagnostic (with a `_fail` fixture) before merge; with that, this is a clean APPROVE. (Posting as COMMENT-event with a REQUEST CHANGES verdict because the reviewer account is also the PR author — GitHub won't let me submit a formal Request-Changes on my own PR.)

leo-aa88 (commented):
# APPROVE

**LGTM.** Re-review of head `1d5ba93` (was `f837653`). The follow-up commit *"reject grind (continue) outside a loop at semantic-analysis time"* resolves the MAJOR finding from my previous review. Verified by building the head in a worktree.

## Previous finding — RESOLVED
The fix adds a `loop_depth` counter to the semantic analyzer, incremented around each loop body (`flex`/`goon`/`mewing`) and **reset to 0 on entering a function body** (saved/restored), then rejects `NODE_CONTINUE_STATEMENT` when `loop_depth == 0`. This closes both halves of the bug:

- **CONFIRMED — silent truncation gone.** `skibidi main { yapping("A"); grind; yapping("B"); }` now errors `grind (continue) outside a loop at line 1`, exit 1 — no more silent skip.
- **CONFIRMED — cross-function leak gone.** The function-entry `loop_depth = 0` reset means a bare `grind` in a helper called from a loop is now rejected at compile time rather than hijacking the caller's loop.
- **CONFIRMED — happy path intact.** `continue_grind.brainrot` still prints `0 1 3 4 / w1 w3 / d1 d3 / 0-1 1-1 2-1`; a helper with its *own* loop+`grind` called from a loop still works (`j0 j2 i0 j0 j2 i1`) — the reset doesn't over-reject. `grind` in a `switch` nested in a loop remains allowed (switch doesn't reset depth), matching C.
- **CONFIRMED — coverage + hygiene.** New `test_cases/semantic_error_grind_outside_loop.brainrot` pins the diagnostic (`...at line 8`); `pytest -k "continue_grind or grind_outside_loop"` → 2 passed; valgrind clean on both fixtures (0 lost, 0 errors), error path included.

The guard is placed correctly (semantic analysis, mirroring the project's other compile-time rejections) and matches C's "continue not within a loop". This is the right fix, not a band-aid.

## Note (not a finding)
The PR currently shows `CONFLICTING` — `tests/expected_results.json` needs a merge-from-main now that #355 has landed. Code-wise it's ready; it just needs the branch brought up to date (the same trivial append-union as the earlier PRs) before it can merge.

## VERDICT

The MAJOR issue is fixed properly and the feature is now sound: `grind` works in every loop form, nests correctly, coexists with `bruh`, and is a clean compile-time error outside a loop. Deserves to merge once the branch is updated against main. (Posting as COMMENT-event with an APPROVE verdict because the reviewer account is also the PR author — GitHub won't accept a formal approval from me.)

leo-aa88 (commented):
# APPROVE

**LGTM.** Re-review of head `48f3e1b` (was `1d5ba93`). Builds clean, all features verified.

## Note — a BLOCKING intermediate state was caught and is now resolved
An intermediate merge commit on this branch (`7a9dc78`) had **committed conflict markers left in `ast.c`** — where this PR's `continue_requested` global collided with #246's `g_program_exit_code` (which reached `main` via #355). That head did **not compile** (`ast.c: error: version control conflict marker in file`). GitHub still reported it MERGEABLE because the markers were literal file content, so `git` saw no *merge* conflict — a reminder that "mergeable" ≠ "builds". The follow-up commit `48f3e1b` resolves it correctly by **keeping both globals**, which is the right union. Flagging it so it's on record; no action needed since it's already fixed.

## Verified at `48f3e1b`
- **CONFIRMED — compiles clean**, no markers anywhere; both `continue_requested` and `g_program_exit_code` are declared.
- **CONFIRMED — grind feature intact:** `continue_grind.brainrot` → `0 1 3 4 / w1 w3 / d1 d3 / 0-1 1-1 2-1`; the semantic guard still rejects out-of-loop `grind` (`semantic_error_grind_outside_loop` → `Error: grind (continue) outside a loop at line 8`, exit 1).
- **CONFIRMED — coexists with #246:** `bussin_main_exit_code` still exits 42, so both merged globals work together.

The #274 code I approved at `1d5ba93` (grind implementation + out-of-loop semantic rejection) is unchanged; only the main-integration merge was added, and after the marker fix it's sound.

## VERDICT

Deserves to merge — feature correct, out-of-loop guard in place, cleanly integrated with current `main`, `MERGEABLE`. (Posting as COMMENT-event with an APPROVE verdict because the reviewer account is also the PR author.)

PR #359, opened 2026-09-07
fix(stdrot): reject '*' dynamic width/precision instead of passing it to snprintf (#275)

### LINKED ISSUE #275: [Bug]: yapping/yappin/baka pass '*' (dynamic width/precision) to snprintf without the extra argument — undefined behavior / info leak
### Description

The print-formatting helpers `process_yapping_format()` (`stdrot/yapping.c`,
used by `yapping`/`yappin`) and `process_baka_format()` (`stdrot/baka.c`, used
by `baka`) build up a per-conversion specifier string and then hand it to
`snprintf()` with **exactly one** value argument:

```c
buffer_offset += snprintf(buffer + buffer_offset,
                          sizeof(buffer) - buffer_offset,
                          specifier, arg->val.i);   /* one arg only */
```

While parsing the conversion, the flags/width/precision skip loop explicitly
includes `*`:

```c
/* Skip flags, width, precision */
while (strchr("-+ #0123456789.*", *format) != NULL)
{
    format++;
}
```

`*` is copied verbatim into `specifier`, but a `*` field width or precision
tells `snprintf` to consume an **extra** `int` argument *before* the value.
The formatter never supplies that extra argument, so `snprintf` is invoked
with **fewer arguments than the format string consumes** — undefined behavior
(C11 §7.21.6.1p2). In practice the value conversion then reads an indeterminate
adjacent vararg slot: garbage output, an information leak of stack/register
contents, or a crash (`%*s` interprets the leaked slot as a `char *` and
dereferences it).

This is a **separate root cause** from #269 (that one is about misusing
`snprintf`'s return value); this one is about passing `*` through to `snprintf`
without a matching argument. Note also — as #269 documents — `libstdrot.so` is
built **without** sanitizers, so neither ASan nor the test suite catches this.

### Minimal reproduction code

```c
skibidi main {
    🚽 one brainrot arg, but "%*d" makes snprintf consume TWO C varargs:
    🚽 the '*' eats 5 as the width, then the 'd' reads an unprovided arg.
    yapping("[%*d]", 5);
    bussin 0;
}
```

### Expected behavior

`*` in a format is not something this marshalling layer can support (it can
only pair one brainrot argument with one conversion), so it should be rejected
with a clear diagnostic, or the `*` should be stripped/handled explicitly.
Under no input should `snprintf` be called with fewer arguments than its format
consumes.

### Actual behavior

```shell
$ ./brainrot star.brainrot
[    1]          # width 5 from the arg; the "1" is an indeterminate stack/register slot

# The value is undefined and can vary / leak. With a string arg it is worse:
#   yapping("%*s", "hi");   -> snprintf(buf,n,"%*s", str.data)
#   '*' consumes str.data (a pointer) as the width int, then '%s' reads an
#   unprovided vararg as a char* and dereferences it -> potential SIGSEGV /
#   arbitrary read. '%.*d' with a large precision also feeds the #269 overflow.
```

### Brainrot version / commit

### Operating system

WSL2 (Ubuntu), x86-64

### Additional context

Same fix location in both `stdrot/yapping.c` and `stdrot/baka.c`. Simplest
fix: drop `*` from the flags/width/precision skip set and treat a `*` in the
conversion as an error (or transform it into a fixed width drawn from the next
marshalled argument). Related: #269 (same two functions, different bug), and
the fact that `libstdrot.so` is not ASan-instrumented, which is why the fuzz
corpus / CI never flags either issue.

leo-aa88:
## Description

The print formatters — `stdrot_format_to_stream` (`stdrot/yapping.c`, behind
`yapping`/`yappin`/`yapto`) and `process_baka_format` (`stdrot/baka.c`) — copy a
conversion's flags/width/precision (including `*`) verbatim into the specifier
handed to `snprintf`, but only ever supply **one** value argument per
conversion. A `*` field width or precision makes `snprintf` read an **extra
`int`** argument *before* the value, so `snprintf` was called with fewer
arguments than the format consumes — **undefined behavior** reading an adjacent
vararg slot: garbage output, a stack/register **leak**, or a **crash** for
`%*s` (the leaked slot is dereferenced as a `char *`). Separate root cause from
#269 (return-value misuse), same two functions.

### Fix

After building the per-conversion specifier, reject it if it contains `*`: emit a
diagnostic to stderr, print nothing for that conversion, and still consume the
paired argument so later conversions stay aligned. Fixed width/precision (`%5d`,
`%.2f`) is unaffected.

```
yapping("[%*d][%*s]", 5, "hi");
// before: "[    <garbage>][<crash/leak>]"
// after : stdout "[][]", stderr two "Error: '*' ... not supported" lines
```

Note: `baka`'s grammar is `BAKA LPAREN expression RPAREN` — a single argument —
so its formatter can't receive a value vararg from Brainrot source; baka's guard
is a **defensive parity** fix, and the reachable/tested path is variadic
`yapping`/`yappin`.

## Related Issue

Fixes #275

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement
- [ ] Refactor

## Checklist

`make test` → **520 passed**; full `make valgrind` sweep → exit 0; `make format-check` clean. `make cppcheck` left to CI (local 2.7 < required 2.13).

### DIFF
--- a/stdrot/baka.c
+++ b/stdrot/baka.c
@@ -69,6 +69,25 @@ static void process_baka_format(const char *format, const StdrotValue *args,
             memcpy(specifier, start, length);
             specifier[length] = '\0';
 
+            /* A '*' field width or precision makes snprintf() read an EXTRA int
+               argument (before the value) from its varargs. This marshalling
+               layer only ever pairs one Brainrot argument with one conversion,
+               so it never supplies that extra int -- passing the '*' through
+               would call snprintf with fewer arguments than the format
+               consumes: undefined behavior that reads an adjacent vararg slot
+               (garbage, a stack/register leak, or a crash for `%*s`). Reject it
+               with a diagnostic instead, emit nothing for this conversion, and
+               still consume the paired argument so the rest stay aligned
+               (#275). */
+            if (strchr(specifier, '*') != NULL)
+            {
+                fprintf(stderr, "Error: '*' dynamic width/precision is not "
+                                "supported in format strings\n");
+                arg_idx++;
+                format++;
+                continue;
+            }
+
             const StdrotValue *arg = &args[arg_idx];
 
             if (spec == 'b')
--- a/stdrot/yapping.c
+++ b/stdrot/yapping.c
@@ -93,6 +93,25 @@ void stdrot_format_to_stream(FILE *out, const char *format,
             memcpy(specifier, start, length);
             specifier[length] = '\0';
 
+            /* A '*' field width or precision makes snprintf() read an EXTRA int
+               argument (before the value) from its varargs. This marshalling
+               layer only ever pairs one Brainrot argument with one conversion,
+               so it never supplies that extra int -- passing the '*' through
+               would call snprintf with fewer arguments than the format
+               consumes: undefined behavior that reads an adjacent vararg slot
+               (garbage, a stack/register leak, or a crash for `%*s`). Reject it
+               with a diagnostic instead, emit nothing for this conversion, and
+               still consume the paired argument so the rest stay aligned
+               (#275). */
+            if (strchr(specifier, '*') != NULL)
+            {
+                fprintf(stderr, "Error: '*' dynamic width/precision is not "
+                                "supported in format strings\n");
+                arg_idx++;
+                format++;
+                continue;
+            }
+
             const StdrotValue *arg = &args[arg_idx];
 
             if (spec == 'b')
--- a/test_cases/format_star_width_rejected.brainrot
+++ b/test_cases/format_star_width_rejected.brainrot
@@ -1,0 +1,11 @@
+🚽 Regression for #275: a '*' dynamic field width or precision in a format
+🚽 string used to be copied straight into the snprintf conversion, which then
+🚽 read an EXTRA vararg this marshalling layer never supplies -- undefined
+🚽 behavior: garbage output, a stack/register leak, or a crash for %*s (the
+🚽 leaked slot is dereferenced as a char*). It is now rejected with a
+🚽 diagnostic and emits nothing for that conversion; literals around it and the
+🚽 arguments after it stay intact.
+skibidi main {
+    yapping("[%*d][%*s]", 5, "hi");
+    bussin 0;
+}
--- a/tests/expected_results.json
+++ b/tests/expected_results.json
@@ -439,5 +439,6 @@
     "return_from_inside_loop_in_main": "before",
     "bussin_main_exit_code": "ExitCode:42",
     "bussin_main_stops": "before",
-    "bussin_main_edgy_exit": "ExitCode:3"
+    "bussin_main_edgy_exit": "ExitCode:3",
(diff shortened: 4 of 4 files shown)


leo-aa88 (commented):
# APPROVE

**LGTM.** Adversarial pass on this `*`-width/precision rejection. Built the head in a worktree and exercised the star forms, fixed width/precision, literal `*`, and the dangerous `%*s` path under ASan + valgrind.

## Findings

### The bug is real and the fix is correct
- **CONFIRMED — the UB.** `%*d`/`%.*f`/`%*.*d` make `snprintf` read an extra `int` (or two) *before* the value from its varargs, but this marshalling layer supplies only one value per conversion — so `snprintf` was called with fewer args than the format consumes: an adjacent-vararg read (garbage/leak, or a `char*` deref crash for `%*s`).
- **CONFIRMED — rejection covers every form.** `strchr(specifier, '*')` on the *built conversion* catches width-star, precision-star, and both. Live: `[%*d][%*s]` → stdout `[][]` + two stderr diagnostics; `%.*f` and `%*.*d` likewise rejected (`a[]b[]c`). The `%*s` crash path is now safe — no crash, valgrind clean.
- **CONFIRMED — alignment preserved.** The reject path does `arg_idx++; format++; continue;`, so the paired argument is consumed and later conversions stay aligned; surrounding literals print intact.
- **CONFIRMED — no regression / no false positives.** Fixed width/precision `%5d`/`%.2f`/`%-3d` → `[   42][3.14][7  ]` unchanged; a literal `*` in ordinary text (`3 * 4`) is not rejected (the check is on the conversion specifier only, not literal bytes).
- **CONFIRMED — harness + memory.** New `format_star_width_rejected.brainrot`; `expected_results.json` valid; `pytest -k format_star_width_rejected` passes; valgrind on the fixture (incl. `%*s`) → 0 def/indirect lost, 0 errors.

The `baka.c` guard is correctly described as defensive parity (its grammar takes a single argument, so it can't receive a value vararg from Brainrot source); the reachable path is variadic `yapping`/`yappin`, which is what the test exercises. Separate root cause from #269, same two functions — good to fix both here.

## VERDICT

Correct, minimal fix for a genuine varargs-UB/crash, covering all `*` forms without regressing fixed width/precision or misparsing literal `*`, with the crash-prone `%*s` case now safe and leak-clean. Deserves to merge. (Posting as COMMENT-event with an APPROVE verdict because the reviewer account is also the PR author.)

leo-aa88 (commented):
# APPROVE

**LGTM.** Re-review of head `2d20e16` (was `55e0f28`) — merge commit bringing the branch up to current `origin/main` (after #358 landed). Content commit `55e0f28` I reviewed is unchanged.

Verified: **no committed conflict markers** (checked explicitly), builds clean, `format_star_width_rejected` fixture → two `*`-rejection diagnostics + stdout `[][]`. Merged `expected_results.json` valid; `MERGEABLE`. This PR touches only `stdrot/*.c`, with no overlap on anything #358 changed, so the merge was conflict-free in code.

## VERDICT

Same APPROVE — merge-only, content and behavior unchanged. (COMMENT-event; reviewer is also the PR author.)

PR #362, opened 2026-09-07
fix: compute int +, -, *, unary - with defined wrapping, not signed-overflow UB (#280)

leo-aa88:
## Description

`rizz` (int) arithmetic used native **signed** `int` operations that are
**undefined behavior on overflow** — UBSan flags `2147483647 + 1`, `INT_MIN - 1`,
`INT_MIN * 2`, and `-INT_MIN`:

```
ast.c:...: runtime error: signed integer overflow: 2147483647 + 1 ...
ast.c:...: runtime error: negation of -2147483648 cannot be represented ...
```

The language's own semantics are two's-complement wrapping
(`integer_overflow.brainrot` expects `-2147483648` from `INT_MAX + 1`), but that
only "worked" because x86-64 happens to wrap — the operations were still UB and
not guaranteed under `-O2` or a different compiler.

### Fix

Do the four `VAR_INT` operations in `unsigned int` (defined modular arithmetic)
and cast back, in `handle_binary_operation()`'s `OP_PLUS`/`OP_MINUS`/`OP_TIMES`
and `handle_unary_expression()`'s `OP_NEG`. The cast back is the universal
two's-complement conversion, so the documented wrapped values are now produced
by **defined** operations. `VAR_SHORT` is untouched: `short` promotes to `int`,
so its results can't overflow `int`.

### Testing — why the JSON suite isn't enough

`make test` only diffs stdout, and UBSan defaults to print-and-continue, so the
UB stayed invisible (stdout still wrapped). Added `int_overflow_wraps.brainrot`
plus a dedicated pytest, `test_int_arithmetic_no_signed_overflow_ub`, that runs
it under `UBSAN_OPTIONS=halt_on_error=1` — so any signed-overflow UB **aborts**
(nonzero exit). It asserts exit 0, no `runtime error` on stderr, and the wrapped
output `-2147483648 / 2147483647 / -2 / -2147483648`. Verified the fixture is
UBSan-clean now.

## Related Issue

Fixes #280

## Type of Change

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Refactor

## Checklist

- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [x] I have documented my changes in the code or documentation
[...]

PR #361, opened 2026-09-07
fix: free a rant's old buffer on reassignment instead of leaking it (#277)

leo-aa88:
## Description

A `VAR_STRING` variable owns its heap `.data` buffer (declaration `safe_strdup`s
it; `hm_free` frees it at scope teardown), but the assignment write path
overwrote the `strvalue` slot with a bare `= evaluate_expression_string(expr)`,
**orphaning the previous buffer** — a leak on every reassignment, once per
iteration in a loop:

```c
rant s = "hello";
s = "world";        // "hello" buffer leaked
```

### Fix

Free the old buffer before installing the new one, in **both** `VAR_STRING`
write sites (`write_value_to_address` and `initialize_variable_from_expr`).

Order matters and is safe: **evaluate the RHS first, then free, then assign.**
`evaluate_expression_string()` always returns an independently `safe_strdup`'d
copy (even for a bare identifier), so:
- `s = s` copies s's content *before* the free — no use-after-free;
- no two variables ever share a `.data` pointer — no double-free at teardown.

On a first write the slot is zero-initialized (`variable_new` memsets), so the
`SAFE_FREE` is a no-op.

## Related Issue

Fixes #277

## Type of Change

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Refactor

## Checklist

- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [x] I have documented my changes in the code or documentation
- [x] I have added tests that prove my change works, at the lowest appropriate layer (`test_cases/*.brainrot` for language-visible behavior; a host-side C test wired into `make test` for internal APIs Brainrot source cannot reach). Required for every bug fix, feature, refactor, performance change, and breaking change — not optional, not "if applicable". See CONTRIBUTING.md ("TESTS ARE MANDATORY").
- [x] Those tests cover the happy path, the original bug (for fixes), error cases, edge cases, **and** adversarial cases. If the existing harness could not reach the behavior, I extended it or added a lower-level test.
- [ ] If this PR cannot affect program behavior (docs/comments/license only), I explained that in the Description instead of skipping the items above in silence
- [x] I have run `make format-check` locally (or `make format` to fix)
- [x] I have run the unit tests locally
- [x] I have run the valgrind memory tests locally
- [x] All new and existing tests pass

### Test

[...]

PR #363, opened 2026-09-07
fix: stop `i++ + i++` crashing by binding ++/-- tighter than arithmetic (#281)

### LINKED ISSUE #281: [Bug]: Two increment/decrement operators in one expression (i++ + i++) crash with a heap-buffer-overflow

leo-aa88:
## Description

Two increment/decrement operators in one binary expression (e.g. `i++ + i++`) crashed with a heap-buffer-overflow. `INC`/`DEC` sat at the **relational** precedence tier — below `+` — so on the lookahead `++` after `a + b` the parser reduced `a + b` first, and `i++ + i++` parsed as `(i++ + i)++`: a post-increment applied to a **non-lvalue** operation node. `handle_unary_expression()` then read `operand->data.name` off that node, misreading the AST union as a variable-name `String` whose length no longer matched its buffer, so `fnv1a_hash()` read past the arena (`lib/hm.c`).

The fix has two parts:

- **Grammar (`lang.y`).** Move `INC`/`DEC` to a precedence tier **above** every arithmetic and relational operator, matching C's postfix/prefix `++`/`--`, so the `++` shifts onto its own operand and `i++ + i++` groups as `(i++) + (i++)`. Prefix `++`/`--` bind at the unary (`UMINUS`) tier. `bison -Wcounterexamples` reports **no new conflicts**.
- **Semantics (`semantic_analyzer.c`, `ast.c`).** The interpreter's `++`/`--` only ever supported a plain variable target (it calls `set_int_variable(name, ...)`). A non-identifier operand such as `(a + b)++`, `a[i]++`, or `p.x++` either **crashed** (`a[0]++` on the old build) or **silently did nothing** (`p.x++`). Reject a non-`NODE_IDENTIFIER` operand in the semantic analyzer with a clear diagnostic **before** interpretation, with a matching defensive guard in `handle_unary_expression()`.

The exact totals in the new fixture still reflect the pre-existing #225 double-evaluation of a post/pre-inc operand in expression context (a separate, open issue). This PR's scope, per the issue, is that these expressions run **without memory corruption** and the interpreter **does not crash** — which they now do in every context the issue lists (declaration initializer, assignment RHS, `yapping` argument, bare statement, distinct variables).

## Related Issue

Fixes #281

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement
- [ ] Refactor

## Checklist

### Tests added
- `test_cases/inc_dec_in_expression.brainrot` — `i++ + i++`, `++j + ++j`, `k-- + k--` all run without crashing (the original bug).
- `test_cases/semantic_error_incdec_non_lvalue.brainrot` — `(a + b)++` is rejected cleanly with `increment/decrement requires a variable operand` instead of corrupting memory (adversarial / non-lvalue case).

Both wired into `tests/expected_results.json`. `make test` (525 passed), `make valgrind` (clean, exit 0), and `make format-check` all pass locally.

### DIFF
--- a/ast.c
+++ b/ast.c
@@ -4165,6 +4165,27 @@ static bool expression_is_truthy(ASTNode *expr)
 void *handle_unary_expression(ASTNode *node, void *operand_value,
                               int operand_type)
 {
+    /* ++/-- write back through the operand, so they require an lvalue: a plain
+       variable name. The four cases below read operand->data.name directly, so
+       a non-identifier operand (e.g. `5++`, `(a + b)++`) would misread the AST
+       union as a String and hand garbage to the hashmap (#281). Reject it here
+       with a clear diagnostic instead of corrupting memory. */
+    switch (node->data.unary.op)
+    {
+    case OP_PRE_INC:
+    case OP_PRE_DEC:
+    case OP_POST_INC:
+    case OP_POST_DEC:
+        if (node->data.unary.operand->type != NODE_IDENTIFIER)
+        {
+            yyerror("increment/decrement requires a variable operand");
+            return NULL;
+        }
+        break;
+    default:
+        break;
+    }
+
     switch (node->data.unary.op)
     {
     case OP_NEG:
--- a/lang.y
+++ b/lang.y
@@ -469,10 +469,18 @@ static void register_anonymous_aggregate_typedef(String alias_name,
 %left OR                /* Logical OR */
 %left AND               /* Logical AND */
 %nonassoc EQ NE         /* Equality operators */
-%nonassoc LT GT LE GE DEC INC   /* Relational operators */
+%nonassoc LT GT LE GE   /* Relational operators */
 %left PLUS MINUS        /* Addition and subtraction */
 %left TIMES DIVIDE MOD  /* Multiplication, division, modulo */
 %right UMINUS           /* Unary minus */
+/* Increment/decrement bind TIGHTER than every arithmetic/relational operator,
+   matching C's postfix/prefix `++`/`--`. Kept at the relational tier before
+   (#281), the postfix form mis-grouped: on lookahead `++` after `a + b` the
+   parser reduced `a + b` first, so `i++ + i++` parsed as `(i++ + i)++` -- a
+   `++` applied to a NON-lvalue operation node, whose data union was then misread
+   as a variable name (heap-buffer-overflow). Above the binary operators, the
+   `++` shifts onto its own operand instead, so `i++ + i++` is `(i++) + (i++)`. */
+%nonassoc INC DEC
 /* Member access `.` (struct_access: expression DOT IDENTIFIER) is a
    postfix operator and must bind TIGHTER than every binary/assignment
    operator above, matching C -- otherwise `v = m.x` parses as `(v = m).x`
@@ -2136,13 +2144,13 @@ unary_operation:
         { $$ = create_unary_operation_node(OP_DEREFERENCE, $2); }
     | AMPERSAND expression %prec UMINUS
         { $$ = create_unary_operation_node(OP_ADDRESS_OF, $2); }
-    | INC expression %prec LOWER_THAN_ELSE
+    | INC expression %prec UMINUS
         { $$ = create_unary_operation_node(OP_PRE_INC, $2); }
-    | DEC expression %prec LOWER_THAN_ELSE
+    | DEC expression %prec UMINUS
         { $$ = create_unary_operation_node(OP_PRE_DEC, $2); }
-    | expression INC %prec LOWER_THAN_ELSE
+    | expression INC %prec INC
         { $$ = create_unary_operation_node(OP_POST_INC, $1); }
-    | expression DEC %prec LOWER_THAN_ELSE
+    | expression DEC %prec DEC
         { $$ = create_unary_operation_node(OP_POST_DEC, $1); }
     ;
 
--- a/semantic_analyzer.c
+++ b/semantic_analyzer.c
@@ -3794,15 +3794,37 @@ void semantic_analyze_with_scope_tracking(SemanticAnalyzer *analyzer,
                     node->line_number > 0 ? node->line_number : 1);
             }
         }
+        else if (node->data.unary.op == OP_PRE_INC ||
+                 node->data.unary.op == OP_PRE_DEC ||
+                 node->data.unary.op == OP_POST_INC ||
+                 node->data.unary.op == OP_POST_DEC)
+        {
+            /* ++/-- write their result back through the operand, so they
+               need an lvalue, not just a value. handle_unary_expression()
+               (ast.c) reads operand->data.name to name the target, so a
+               non-identifier operand (`5++`, `(a + b)++`) would misread the
+               AST union as a variable name and hand garbage to the hashmap
+               (#281). This codebase's ++/-- only support a plain variable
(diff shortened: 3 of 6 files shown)


leo-aa88 (commented):
# APPROVE

**LGTM.** Adversarial pass on this `++`/`--` precedence + lvalue-guard fix. Built the head in a worktree; verified the crash is gone (ASan + valgrind), attacked the precedence change for regressions (full `make test`), and checked the new rejection doesn't over-reach.

## Findings

None blocking. One non-blocking observation below.

- **CONFIRMED — the crash is fixed.** `i++ + i++` (and `++j + ++j`, `k-- + k--`) now run without the heap-buffer-overflow. Fixture → `15 9 / 17 9 / 15 6`, exit 0, ASan-clean; valgrind → 0 lost, 0 errors, no invalid reads. The precedence move makes `i++ + i++` group as `(i++) + (i++)` instead of the old `(i++ + i)++` that misread a non-lvalue op node as a variable name.
- **CONFIRMED — no precedence regression.** `make test` → **525 passed**. Existing `++`/`--` usages all still work: `flex (...; i++)` (sum=6), `arr[j++]` (idx=10, j=1), `c++` in a `goon` loop (c=5), and `i++ < n` in a relational context (`yes 1`). No new bison conflicts reported.
- **CONFIRMED — the lvalue guard is correct and doesn't over-reach.** Non-identifier operands are rejected before interpretation with a clear diagnostic: `(a + b)++` → "increment/decrement requires a variable operand at line 10" (exit 1). Plain `i++`/`++i` still accepted. The double guard (semantic analyzer + defensive `handle_unary_expression`) is belt-and-suspenders; the semantic pass catches it first.
- **CONFIRMED — `arr[i]++` / `p.x++` now fail cleanly** where they previously crashed (`arr[0]++`) or silently no-op'd (`p.x++`) — a strict improvement.

## Observation (non-blocking)
Rejecting `arr[i]++` and `p.x++` is a behavior change: these C-lvalue forms are now compile errors rather than (broken) runtime behavior. Given they never worked correctly here and #281's scope is "don't corrupt memory / don't crash", turning them into a clean diagnostic is the right call — but it does leave element/field increment unimplemented (a reasonable follow-up). The fixture's totals reflect the separate, still-open #225 double-evaluation of an inc/dec operand in expression context; the PR correctly scopes that out and pins only the deterministic no-crash behavior.

## VERDICT

Correct two-part fix: the precedence move eliminates the mis-grouping that caused the overflow, and the semantic guard turns every remaining non-lvalue `++`/`--` into a clean pre-interpretation error instead of memory corruption. Crash gone, memory clean, no suite regression, existing usage intact. Deserves to merge. (Posting as COMMENT-event with an APPROVE verdict because the reviewer account is also the PR author.)

leo-aa88 (commented):
# APPROVE

**LGTM.** Re-review of head `a01f844` (was `3df828b`) — merge commit bringing the branch up to current `origin/main` (after #361 landed). Content commit `3df828b` I reviewed is unchanged.

Verified: no committed conflict markers, merged `expected_results.json` valid (448 keys) with `inc_dec_in_expression` and `semantic_error_incdec_non_lvalue` intact, `MERGEABLE`. Rebuilt from the merge head — `inc_dec_in_expression` → `15 9 / 17 9 / 15 6` (no crash), and #361's `rant_reassign_no_leak` still works (`iter`). This PR's `handle_unary_expression`/semantic-analyzer changes don't overlap #361's VAR_STRING write sites, so the ast.c merge is clean.

## VERDICT

Same APPROVE — merge-only, content and behavior unchanged. (COMMENT-event; reviewer is also the PR author.)

leo-aa88 (commented):
# APPROVE

**LGTM.** Re-review of head `fe6bfcf` (was `a01f844`) — merge commit bringing the branch up to current `origin/main` (after #362 landed). Content commit `3df828b` I reviewed is unchanged.

Verified by rebuild: no committed conflict markers, `MERGEABLE`. #362's now-merged `handle_unary_expression` OP_NEG change and this PR's `++`/`--` lvalue guard (top of the same function) coexist cleanly — `inc_dec_in_expression` → `15 9 / 17 9 / 15 6` (no crash) and #362's `int_overflow_wraps` → correct wraps. Merged `expected_results.json` valid.

## VERDICT

Same APPROVE — merge-only, content and behavior unchanged, coexistence with #362 confirmed. (COMMENT-event; reviewer is also the PR author.)

PR #364, opened 2026-09-07
fix: `++`/`--` on chad/gigachad adds 1.0 instead of corrupting the value (#284)

### LINKED ISSUE #284: [Bug]: ++/-- on chad/gigachad (float/double) corrupts the value instead of adding 1
### Description

Applying `++`, `--` (prefix or postfix) to a `chad` (float) or `gigachad` (double) variable **corrupts the variable** instead of adding/subtracting 1.0. Integer increment works correctly, so this is specific to floating-point operands.

Root cause: `interpreter_visit_unary_operation` (`interpreter.c:206-212`) routes every inc/dec through `evaluate_expression_int(node)` regardless of operand type. That integer evaluator's unary case (`ast.c:4130-4131`) forces the operand type to `VAR_INT`:

```c
int operand = evaluate_expression_int(node->data.unary.operand);   // reads the float's bits as int
int *result = (int *)handle_unary_expression(node, &operand, VAR_INT);
```

So `handle_unary_operation` takes its `VAR_INT` branch and calls `set_int_variable()` on a `VAR_FLOAT`/`VAR_DOUBLE` slot — even though that function *already has* correct `VAR_FLOAT`/`VAR_DOUBLE` branches (`ast.c:3063-3080`) that are never reached from this path. The float variable ends up holding an int written into its union, so a later read/print sees garbage.

### Minimal reproduction code

```c
skibidi main {
    chad f = 1.5;
    yapping("before %.2f", f);
    ++f;
    yapping("after %.2f", f);   🚽 expect 2.50
}
```

### Expected behavior

```
before 1.50
after 2.50
```

### Actual behavior

```shell
before 1.50
after 
```

The `%.2f` after the increment prints nothing — `f` no longer marshals as a valid float. `f++`, `--f`, `f--` on `chad`, and all four on `gigachad`, behave identically. The `rizz` (int) equivalent correctly prints `6` for `rizz i = 5; ++i;`.

### Brainrot version / commit

### Operating system

WSL2 (Ubuntu), x86-64

### Additional context

`handle_unary_operation` already implements the float and double cases correctly; the fix is to stop hard-coding `VAR_INT` at the statement entry point. `interpreter_visit_unary_operation` should dispatch on the operand's actual type (`evaluate_expression_float`/`evaluate_expression_double` for float/double operands, as it does elsewhere) so the existing correct branches are used. Distinct from the other recently-filed unary issues #278 (short negation), #280 (int-overflow UB), and #281 (double-increment heap overflow).

leo-aa88:
## Description

Applying `++`/`--` (prefix or postfix) to a `chad` (float) or `gigachad` (double) variable **corrupted** it instead of adding/subtracting 1.0.

`interpreter_visit_unary_operation()` routed **every** increment/decrement through `evaluate_expression_int()`, which reads the float/double slot's bits as an `int` and then writes the int result back through `set_int_variable()` on a `VAR_FLOAT`/`VAR_DOUBLE` slot. So `++f` on `chad f = 1.5` produced `2.0` (fractional part lost), and a loop counter like `flex (chad c = 0.5; c < 3.0; ++c)` stepped by whole ints. On the commit named in the issue it corrupted the slot so badly the value no longer marshalled as a float at all (`%.2f` printed nothing).

`handle_unary_expression()` already has correct `VAR_FLOAT`/`VAR_DOUBLE` branches, and `evaluate_expression_float()`/`evaluate_expression_double()` already reach them — the statement-level dispatch simply never used them.

**Fix:** dispatch on the operand's real type (`get_expression_type(node)`) so a float operand goes through `evaluate_expression_float()` and a double through `evaluate_expression_double()`; `int`/`short`/`char` stay on the integer path, which already handles them correctly. `get_expression_type()` is now declared in `ast.h` so `interpreter.c` can call it (it was defined in `ast.c` with no prototype).

Verified fixed in every context: statement `++f;`/`f++;`/`--f;`/`f--;` on both `chad` and `gigachad`, a `flex` loop counter driven by `++c`, and expression contexts (`chad b = ++a;`, `yapping("%.2f", c++)`). Integer/short/char inc-dec unchanged.

## Related Issue

Fixes #284

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement
- [ ] Refactor

## Checklist

### Tests added
- `test_cases/float_inc_dec.brainrot` — prefix and postfix `++`/`--` on `chad` and `gigachad` in statement context, a float loop counter driven by `++c`, and an integer inc/dec regression guard. Wired into `tests/expected_results.json`.

`make test` (525 passed), `make valgrind` (clean, exit 0), and `make format-check` all pass locally.

### DIFF
--- a/ast.h
+++ b/ast.h
@@ -757,6 +757,7 @@ void execute_if_statement(ASTNode *node);
 void reset_modifiers(void);
 bool check_and_mark_identifier(ASTNode *node, const String contextErrorMessage);
 bool is_expression(ASTNode *node, VarType type);
+VarType get_expression_type(ASTNode *node);
 int get_expression_pointer_level(ASTNode *node);
 uintptr_t evaluate_expression_pointer(ASTNode *node);
 void *evaluate_lvalue_address(ASTNode *node);
--- a/interpreter.c
+++ b/interpreter.c
@@ -227,7 +227,26 @@ void *interpreter_visit_unary_operation(Visitor *self, ASTNode *node)
         node->data.unary.op == OP_POST_INC ||
         node->data.unary.op == OP_POST_DEC)
     {
-        evaluate_expression_int(node);
+        /* Dispatch on the operand's real type. Routing every inc/dec through
+           evaluate_expression_int() reads a chad/gigachad slot's bits as an
+           int and then writes the int result back through set_int_variable()
+           on a VAR_FLOAT/VAR_DOUBLE slot, corrupting the value (#284) --
+           `++f` on `chad f = 1.5` gave 2.0, not 2.5. The float/double
+           evaluators already take handle_unary_expression()'s correct
+           VAR_FLOAT/VAR_DOUBLE branches; int/short/char stay on the integer
+           path, which handles them correctly. */
+        switch (get_expression_type(node))
+        {
+        case VAR_FLOAT:
+            evaluate_expression_float(node);
+            break;
+        case VAR_DOUBLE:
+            evaluate_expression_double(node);
+            break;
+        default:
+            evaluate_expression_int(node);
+            break;
+        }
     }
     return NULL;
 }
--- a/test_cases/float_inc_dec.brainrot
+++ b/test_cases/float_inc_dec.brainrot
@@ -1,0 +1,42 @@
+🚽 Regression test for issue #284: ++/-- (prefix and postfix) on a chad
+🚽 (float) or gigachad (double) variable used to corrupt the value instead
+🚽 of adding/subtracting 1.0. interpreter_visit_unary_operation() routed
+🚽 every inc/dec through evaluate_expression_int(), reading the float slot's
+🚽 bits as an int and writing an int back through set_int_variable() on a
+🚽 VAR_FLOAT/VAR_DOUBLE slot -- so `++f` on `chad f = 1.5` produced 2.0, not
+🚽 2.5. It now dispatches on the operand's real type, using the float/double
+🚽 branches handle_unary_expression() already had. Integer inc/dec is
+🚽 unaffected (still on the integer path).
+skibidi main {
+    🚽 chad (float) statement-context ++/--
+    chad f = 1.5;
+    ++f;
+    yapping("%.2f", f);   🚽 2.50
+    f++;
+    yapping("%.2f", f);   🚽 3.50
+    --f;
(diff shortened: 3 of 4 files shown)


leo-aa88 (commented):
# APPROVE

**LGTM.** Adversarial pass on this float/double `++`/`--` fix. Built the head in a worktree and verified statement, loop, and expression contexts under ASan + valgrind.

## Findings

- **CONFIRMED — the corruption is fixed.** The statement-level dispatch `interpreter_visit_unary_operation()` routed every inc/dec through `evaluate_expression_int()`, reading a `chad`/`gigachad` slot as `int` and writing an int back via `set_int_variable()`. Dispatching on `get_expression_type(node)` sends float→`evaluate_expression_float()`, double→`evaluate_expression_double()` (which already take `handle_unary_expression()`'s correct VAR_FLOAT/VAR_DOUBLE branches). Fixture → `2.50 / 3.50 / 2.50 / 1.50` (chad ++/--), `3.50 / 2.50` (gigachad), `4.50` (float loop counter stepping by 1.0), exit 0 — the fractional part is preserved (`++f` on `1.5` → `2.5`, not `2.0`).
- **CONFIRMED — `get_expression_type(node)` resolves the operand's real type** for the inc/dec node (the float output proves the VAR_FLOAT branch is taken); the new `ast.h` prototype lets `interpreter.c` call it.
- **CONFIRMED — integer path unchanged.** `++n` on `rizz n = 5` → `6` (the `default:` arm still uses `evaluate_expression_int()`); short/char likewise unaffected.
- **CONFIRMED — expression contexts correct.** `chad b = ++a` (a=1.5) → `2.50 2.50` (pre-inc returns new value, single increment); `yapping("%.2f", c++)` (c=2.0) → `2.00` then c=`3.00` (post-inc returns old, single increment). No double-evaluation in these contexts.
- **CONFIRMED — memory clean.** `pytest -k float_inc_dec` passes; valgrind on the fixture → 0 lost, 0 errors (the old bug corrupted the slot so `%.2f` printed nothing — now well-formed).

## VERDICT

Minimal, correct dispatch fix reusing the float/double branches that already existed but were never reached from statement context. Fractional inc/dec now works in every context, integer inc/dec is untouched, leak-clean. Deserves to merge. (Posting as COMMENT-event with an APPROVE verdict because the reviewer account is also the PR author.)

PR #365, opened 2026-09-07
fix: error on unrecognized characters instead of silently discarding them (#285, #297)

### LINKED ISSUE #285: [Bug]: Lexer silently discards unrecognized characters (e.g. ~x becomes x) instead of erroring
### Description

The lexer's catch-all rule silently discards any character it doesn't recognize:

```
.                { /* Ignore unrecognized characters */ }
```
(`lang.l:676`)

Because the character is dropped rather than reported, typos and unsupported operators are silently miscompiled instead of raising an error. The most dangerous case is a character that *looks* like a valid operator: `~` (bitwise NOT) is not a token, so `~x` lexes as just `x` — the program compiles and runs, producing a value that is the exact opposite of what the source says, with no diagnostic.

### Minimal reproduction code

```c
skibidi main {
    rizz x = 5;
    yapping("%d", ~x);   🚽 the ~ is silently dropped; prints 5, not -6
}
```

### Expected behavior

Either `~` is a supported operator (then `~5` prints `-6`), or — since it is not — the stray `~` is a **lex/parse error**. An unrecognized character should never be silently swallowed.

### Actual behavior

```shell
5
```

`~0` prints `0`, `~5` prints `5`, `~x` prints `x` — `~` is a no-op. Likewise `rizz x $ = 7;` prints `7`, silently ignoring the `$`. (Stray characters that happen to sit between two tokens, e.g. `5 @ 3`, instead surface as a confusing downstream "unexpected INT_LITERAL" rather than a clear "invalid character" message.)

### Brainrot version / commit

### Operating system

WSL2 (Ubuntu), x86-64

### Additional context

The 🚽 single-line comment has its own rule (`lang.l:557`), so the catch-all is not needed for comments. Recommended fix: replace the silent `.` rule with one that reports an "unexpected character '%c'" error (and fails the parse), the way most compilers treat stray bytes. If bitwise operators (`~`, `&`, `|`, `^`) are intended to exist, that is a separate feature gap — but until then they should error rather than silently vanish. Related in spirit to the "silently wrong instead of erroring" family (#189, #196).

### LINKED ISSUE #297: The lexer silently discards unrecognized characters, so unsupported operators vanish instead of erroring (~ still affected)
## Summary

`lang.l`'s final rule discards anything it doesn't recognise:

```
.                { /* Ignore unrecognized characters */ }
```

An unsupported operator therefore doesn't fail to compile — it *vanishes*, and the surrounding expression quietly means something else. This is the root cause of #289, and at least one more instance is still live.

## Still live: `~`

```c
skibidi main {
    rizz a = ~5;
    yapping("%d", a);   🚽 prints 5
    bussin 0;
}
```

`~` is dropped, so `~5` is `5`. Exactly the shape `!L` evaluating to `L` had before #296 — a bitwise complement that silently isn't one.

## Why it hides so well

Whether the mistake surfaces depends entirely on what sits either side of the discarded character:

| input | result |
| --- | --- |
| `~5` | **silently becomes `5`** |
| `1 @ 2` | syntax error — but only because two literals end up adjacent |
| ``a ` 1`` | syntax error, same accident |

So the failure mode is decided by neighbouring tokens rather than by the unknown character itself. `!` survived unnoticed for as long as it did because `!x` always leaves exactly one valid expression behind.

## Suggested fix

Replace the silent catch-all with a diagnostic — `yyerror` naming the character and the line, the way an unexpected token is reported. Two things worth deciding first:

- **Blast radius.** Any source that currently relies on a stray character being ignored would start failing. That is almost certainly a bug being surfaced rather than a feature being removed, but it should be a deliberate call.
- **Whitespace and comments** are handled by their own rules above, so they are unaffected. The 🚽 comment rule likewise.

An intermediate step, if a hard error feels too aggressive: warn on stderr and keep discarding. That converts silence into evidence without breaking anything.

## Related

- #289 / #296 — `!` was this bug; the operator now has a real token.
- `~` has no token either. Whoever adds one closes that instance; the catch-all still waits for the next.

leo-aa88:
## Description

`lang.l`'s catch-all rule silently dropped any character no token rule recognized:

```
.                { /* Ignore unrecognized characters */ }
```

So an unsupported operator didn't fail to compile — it **vanished**, and the surrounding expression quietly meant something else. `~` is not a token, so `~x` lexed as just `x`: `~5` printed `5`, a bitwise complement that silently wasn't one (the same shape `!L` evaluating to `L` had before #296). A stray `$`, `@`, etc. was dropped the same way, and whether the mistake surfaced at all depended entirely on what happened to sit either side of the discarded byte.

**Fix:** replace the silent catch-all with a diagnostic that names the character and line and sets a new `lex_error_occurred` flag. `lang.y`'s post-`yyparse` gate already fails compilation for `struct_def_had_error`/`typedef_had_error`, so it now also fails for a lexer error. Lexing continues past the bad character so a single stray byte yields one clear error rather than aborting mid-token. Non-printable bytes are reported by hex value. The flag is one-shot per process, matching the two existing flags (the wasm harness runs each fixture in a fresh module instance, so no reset is needed).

Characters inside string literals and 🚽 comments are consumed by their own rules and never reach the catch-all, so they are unaffected (`yapping("a~b")` still prints `a~b`).

Example:

```
$ cat t.brainrot
skibidi main {
    rizz x = 5;
    yapping("%d", ~x);
    bussin 0;
}
$ ./brainrot t.brainrot
Error: unexpected character '~' at line 3
Parsing failed
```

This closes both #285 and #297 (same root cause; #297 notes `~` was still live).

## Related Issue

Fixes #285
Fixes #297

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement
- [ ] Refactor

## Checklist

### Tests added
- `test_cases/lex_error_unexpected_tilde.brainrot` — the `~5` → `5` case from the issue, now `Error: unexpected character '~' at line 8` + `Parsing failed`.
- `test_cases/lex_error_unexpected_dollar.brainrot` — a stray non-operator byte (`$`) that was previously dropped.

Both wired into `tests/expected_results.json`. `make test` (531 passed), `make valgrind` (clean, exit 0 — the error path is leak-free), and `make format-check` all pass locally.

### DIFF
--- a/ast.c
+++ b/ast.c
@@ -28,6 +28,8 @@ bool struct_def_had_error = false;
    semantic analysis/execution when true, and cleanup resets it through
    free_type_alias_registry(). */
 bool typedef_had_error = false;
+/* Set by lang.l's catch-all rule on an unrecognized character; see ast.h. */
+bool lex_error_occurred = false;
 ReturnValue current_return_value;
 /* See the declaration in ast.h: `grind` (continue) sets this, and the
    statement-list visitor and loop visitors consume it (#274). */
--- a/ast.h
+++ b/ast.h
@@ -906,6 +906,12 @@ bool resolve_by_value_struct_source(ASTNode *expr, void **blob_out,
    for why we don't YYABORT for these instead. */
 extern bool struct_def_had_error;
 extern bool typedef_had_error;
+/* Set by lang.l's catch-all rule on a character no token rule recognizes
+   (a stray `~`, `$`, `@`). Like the two flags above, lang.y's post-yyparse
+   gate treats it as a parse failure, so an unrecognized character is reported
+   and stops compilation instead of being silently discarded (#285, #297).
+   One-shot per process, like struct_def_had_error/typedef_had_error. */
+extern bool lex_error_occurred;
 
 /* Enum types (see the comment on EnumDef above for the registry split). */
 void register_enum_def(EnumDef *def);
--- a/lang.l
+++ b/lang.l
@@ -2,6 +2,7 @@
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
+#include <ctype.h>
 #include <errno.h>
 #include <sys/stat.h>
 #if !defined(_WIN32)
@@ -751,7 +752,23 @@ extern int yylineno;
 }
 
 [ \t\n]+         ; /* Ignore whitespace */
-.                { /* Ignore unrecognized characters */ }
+.                {
+    /* Any character no rule above recognizes. Previously silently discarded,
+       which let unsupported operators vanish instead of erroring: `~x` lexed
+       as just `x`, so `~5` printed 5 -- a bitwise complement that silently
+       isn't one -- and `rizz x $ = 7;` dropped the `$` (#285, #297). Report it
+       naming the character and line, and set lex_error_occurred so main()'s
+       post-yyparse gate fails compilation instead of running miscompiled code.
+       Lexing continues so a single stray character yields one clear error
+       rather than aborting mid-token. */
+    if (isprint((unsigned char)yytext[0]))
+        fprintf(stderr, "Error: unexpected character '%c' at line %d\n",
+                yytext[0], yylineno);
+    else
+        fprintf(stderr, "Error: unexpected byte 0x%02x at line %d\n",
+                (unsigned char)yytext[0], yylineno);
+    lex_error_occurred = true;
+}
 
 %%
 
--- a/lang.y
+++ b/lang.y
@@ -2236,7 +2236,8 @@ int main(int argc, char *argv[]) {
     /* Phase 1: Parse the source code to build AST */
     /* typedef_had_error is set by lit alias registration/rejection helpers;
        those parse-time failures must stop before semantic analysis. */
-    if (yyparse() != 0 || struct_def_had_error || typedef_had_error) {
+    if (yyparse() != 0 || struct_def_had_error || typedef_had_error ||
+        lex_error_occurred) {
         if (!parse_error_already_reported) {
             fprintf(stderr, "Parsing failed\n");
         }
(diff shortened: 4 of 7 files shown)


leo-aa88 (commented):
# APPROVE

**LGTM.** Adversarial pass on this lexer catch-all fix. The change affects *all* lexing, so I focused on regression risk — especially the 🚽 comment marker — and ran the full `make test`.

## Findings

- **CONFIRMED — the silent-discard bug is fixed.** An unrecognized character now emits `Error: unexpected character '<c>' at line N` and sets `lex_error_occurred`, which `main()`'s post-`yyparse` gate treats as a parse failure. `~5` → `Error: unexpected character '~' at line 8` + `Parsing failed`, exit 1 (was: silently printed `5`); stray `$` likewise. Matches the fixtures.
- **CONFIRMED — no regression, and 🚽 is safe.** This was my top concern: the 🚽 comment marker is a 4-byte UTF-8 sequence, and if it reached the new catch-all every fixture would error. It doesn't — `lang.l:628` `"🚽"[^\n]*` consumes the marker and the rest of the line before the catch-all. Full `make test` → **531 passed**. 🚽 inside a string literal also still prints (`yapping("a~b 🚽 c")` → `a~b 🚽 c`, exit 0) — the string rule consumes it.
- **CONFIRMED — non-printable/high bytes handled.** A stray UTF-8 smart quote reports `unexpected byte 0xNN at line N` (one per byte) via the `isprint` fallback and fails compilation (exit 1). Slightly noisy for a multibyte char (one line per continuation byte), but each byte is genuinely unrecognized and the result is still one clear compile failure — reasonable, not a defect.
- **CONFIRMED — continues past the bad char** (one clear error per stray byte rather than aborting mid-token), and the **error path is leak-free** (valgrind on `lex_error_unexpected_tilde` → 0 lost, 0 errors). The `lex_error_occurred` flag correctly mirrors the existing one-shot `struct_def_had_error`/`typedef_had_error` pattern and gate.

## VERDICT

Correct, well-scoped fix that turns a silent-corruption footgun (dropped operators/bytes changing program meaning) into a clear compile error, without touching comment or string lexing. Closes both #285 and #297. Full suite green, error path leak-clean. Deserves to merge. (Posting as COMMENT-event with an APPROVE verdict because the reviewer account is also the PR author.)

PR #366, opened 2026-09-07
fix: warning-gate libstdrot.so so stdrot/ warnings fail an ordinary make (#330)

### LINKED ISSUE #330: [Bug]: libstdrot.so is compiled with no warning flags and no sanitizers — stdrot/ is the one part of the tree neither -Werror nor ASan checks
## Problem

`libstdrot.so` is built with **no diagnostics whatsoever**:

```make
Makefile:11  CFLAGS    := -Wall -Wextra -Wpedantic -Werror -O2 -Wuninitialized -fsanitize=address,undefined -fno-omit-frame-pointer -g
Makefile:13  SO_CFLAGS := -fPIC -shared
```

```make
Makefile:217  $(STDROT_LIB): $(STDROT_SRCS) $(STDROT_ABI_HDR)
Makefile:218      $(CC) $(SO_CFLAGS) $(CRYPTO_CFLAGS) -I. -o $@ ...
```

So all **12** files under `stdrot/` (plus `lib/input.c`) compile at gcc's *default* warning level, with no `-Werror` and no ASan/UBSan. Every other `.c` in the tree gets the full set.

Demonstrated — adding an unused variable to `stdrot/bet.c` and running `make libstdrot.so`:

```
warnings/errors: 0
```

That is the standard library. `yapping`, `slorp`, `gamba`, the string builtins and the new file I/O all live there.

## Why this matters more than it looks

**The only job that compiles these sources with warnings is `Release dry-run wasm`.** The wasm target pulls `$(STDROT_SRCS)` into the main binary via `-DSTDROT_STATIC` and uses the full set:

```make
Makefile:163  WASM_CFLAGS := -Wall -Wextra -Wpedantic -Werror -O2 -DSTDROT_STATIC ...
Makefile:467  $(EMCC) $(WASM_CFLAGS) -I. -o $(WASM_JS) $(SRCS) $(STDROT_SRCS) ...
```

A warning introduced in `stdrot/` therefore passes `build`, `test`, `lint` **and** `static-analysis`, and fails only on a job whose name gives no hint that it is the warning gate for the standard library.

That is not hypothetical. In #327 a `/*` accidentally nested inside a block comment (writing `stdrot/*.c` in prose) compiled fine locally and failed only the wasm job:

```
stdrot/yapcat.c:30:69: error: '/*' within block comment [-Werror,-Wcomment]
```

Both gcc and clang reject that at `-Wall`. Nothing local was looking.

The missing **sanitizers** are the other half, and have hidden real bugs — #269 (`yapping`/`yappin`/`baka` stack-buffer-overflow on long formatted output) and #270 (`input_string()` off-by-one) are both open, both in exactly this code. `make valgrind` is currently the only tool that can see memory bugs in the standard library at all.

## Suggested fix

Adding the warning flags is a **clean one-line change today** — measured, not assumed:

| `SO_CFLAGS` addition | gcc warnings |
| --- | --- |
| `-Wall -Wextra -Wuninitialized` | **0** |
| `+ -Wpedantic` | 24 |

So:

```make
SO_CFLAGS := -fPIC -shared -Wall -Wextra -Wuninitialized -Werror
```

builds clean as of `c763400`.

### The `-Wpedantic` wrinkle, for whoever picks this up

The 24 are all the same thing, and none are in the builtins:

```
stdrot/stdrot_api.h:519:24: warning: initializer element is not constant [-Wpedantic]
```

That is `STDROT_EXPORT_SIG`'s static `StdrotEntry` initializer. **gcc and clang disagree about it**, which is why the wasm job stays green while using `-Wpedantic -Werror`:

| compiler | `stdrot/bet.c` at `-Wall -Wextra -Wpedantic` |
| --- | --- |
| gcc | 1 warning |
| clang-15 | 0 warnings |

(Same result with `-DSTDROT_STATIC`, so it is the compiler, not the macro form.) emcc is clang-based, hence the asymmetry.

Two reasonable options:
- ship `-Wall -Wextra -Wuninitialized -Werror` now and leave `-Wpedantic` off the `.so` with a comment saying why; or
- make the export macro's initializer a constant expression first, then add `-Wpedantic` and match `CFLAGS` exactly.

The first is the smaller change and closes the actual gap.

### Sanitizers

Adding `-fsanitize=address,undefined` to `SO_CFLAGS` is the more valuable half but needs more care — the `.so` is `dlopen`'d by a host that is already sanitized, and `make valgrind` runs against these same artifacts (ASan's shadow mapping and valgrind do not coexist on every platform). Worth doing as a separate step, possibly as a dedicated CI job rather than the default build.

## Definition of done

- [ ] `stdrot/*.c` compiles with `-Wall -Wextra -Werror` in an ordinary `make`
- [ ] A deliberately introduced warning in `stdrot/` fails `make`, not just the wasm job
- [ ] `-Wpedantic` either enabled or explicitly excluded with a comment explaining the gcc/clang split
- [ ] Sanitizer coverage for `stdrot/` decided on — enabled, or documented as deliberately absent with `make valgrind` named as the substitute
- [ ] `make test` / `make valgrind` / `make format-check` still green

## Context

Found while implementing #213. Baseline `c763400`.

leo-aa88:
## Description

`libstdrot.so` (all of `stdrot/*.c` plus `lib/input.c`) compiled with `SO_CFLAGS = -fPIC -shared` — gcc's **default** warning level, no `-Wall`/`-Wextra`, no `-Werror` — while every other `.c` in the tree gets the full set. The only job compiling these sources with warnings was `Release dry-run wasm` (which pulls `$(STDROT_SRCS)` into the main binary via `-DSTDROT_STATIC`). So a warning introduced in the standard library passed `build`, `test`, `lint` **and** `static-analysis`, and failed only a job whose name gives no hint that it is the warning gate for `stdrot/` — exactly what happened in #327 (a `/*` nested in a block comment).

**Fix:** add `-Wall -Wextra -Wuninitialized -Werror` to `SO_CFLAGS`. Measured on the current tree: **0 warnings** with these flags, so an ordinary `make` stays clean, and a deliberately introduced warning in `stdrot/` now fails it (verified — an unused variable in `stdrot/bet.c` → `-Werror=unused-variable`, `make` exits non-zero).

### `-Wpedantic` (deliberately excluded)
Unlike the main `CFLAGS`, `-Wpedantic` is left off the `.so`: gcc flags `STDROT_EXPORT_SIG`'s static `StdrotEntry` initializer as *"initializer element is not constant"* (24 warnings across `stdrot/`) while clang-15 and emcc accept it. The wasm job (clang/emcc) already covers `-Wpedantic` on this code with the compiler that agrees with the standard here. Documented in a comment on `SO_CFLAGS`.

### Sanitizers (deliberately absent)
`-fsanitize=address,undefined` stays off: the `.so` is `dlopen`'d by an already-sanitized host, and `make valgrind` runs against these same artifacts (ASan's shadow map and valgrind don't coexist on every platform). So `make valgrind` remains the memory-safety gate for `stdrot/`. Also documented in the `SO_CFLAGS` comment, and `docs/building.md` is corrected (it claimed the default `make` builds `libstdrot.so` "(sanitizers on)" — that applies to the interpreter only).

### Definition of done (from the issue)
- [x] `stdrot/*.c` compiles with `-Wall -Wextra -Werror` in an ordinary `make`
- [x] A deliberately introduced warning in `stdrot/` fails `make`, not just the wasm job (verified)
- [x] `-Wpedantic` explicitly excluded with a comment explaining the gcc/clang split
- [x] Sanitizer coverage decided — documented as deliberately absent, `make valgrind` named as the substitute
- [x] `make test` / `make valgrind` / `make format-check` still green

## Related Issue

Fixes #330

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement
- [ ] Refactor

## Checklist

### Test added
This is a build-configuration fix (compiler diagnostics), so no `.brainrot` fixture can exercise it. The lowest appropriate layer is a Makefile-recipe guard: `tests/test_docs_consistency.py::test_libstdrot_recipe_is_warning_gated` reads the **actual** recipe via `make -Bn libstdrot.so` (so it tracks the Makefile rather than duplicating flag strings) and asserts `-Wall`/`-Wextra`/`-Werror` are present — so the gate can't be silently dropped again. `make test` (530 passed), `make valgrind` (clean, exit 0), and `make format-check` all pass locally.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -12,7 +12,19 @@ SANITIZER_FLAGS := -fsanitize=address,undefined
 CFLAGS := -Wall -Wextra -Wpedantic -Werror -O2 -Wuninitialized $(SANITIZER_FLAGS) -fno-omit-frame-pointer -g
 VALGRIND_CFLAGS := $(filter-out $(SANITIZER_FLAGS),$(CFLAGS))
 LDFLAGS := -lfl -lm -ldl -rdynamic
-SO_CFLAGS := -fPIC -shared
+# libstdrot.so (and the test/badnatives .so's built with the same recipe) get
+# the same warning gate as the rest of the tree, so a warning in stdrot/ fails
+# an ordinary `make` -- not only the `Release dry-run wasm` job, which was the
+# sole thing compiling these sources with warnings before (#330). -Wpedantic is
+# deliberately EXCLUDED here (unlike CFLAGS): gcc flags STDROT_EXPORT_SIG's
+# static StdrotEntry initializer as "initializer element is not constant"
+# (24 warnings across stdrot/) while clang-15 and emcc accept it, so the wasm
+# job already covers -Wpedantic on this code with the compiler that agrees with
+# the standard here. Sanitizers are also intentionally absent: this .so is
+# dlopen'd by an already-sanitized host and `make valgrind` runs against these
+# same artifacts (ASan's shadow map and valgrind don't coexist everywhere), so
+# valgrind is the memory-safety gate for stdrot/ rather than ASan/UBSan.
+SO_CFLAGS := -fPIC -shared -Wall -Wextra -Wuninitialized -Werror
 SO_LDFLAGS :=
 
 # `make release` ships binaries (GitHub Actions release matrix). Drop
--- a/docs/building.md
+++ b/docs/building.md
@@ -126,7 +126,7 @@ Run `make help` for the annotated list. The ones that matter day to day:
 
 | Target | What it does |
 | --- | --- |
-| `make` / `make all` | interpreter + `libstdrot.so` (sanitizers on) — the default |
+| `make` / `make all` | interpreter (sanitizers on) + `libstdrot.so` (warning-gated, no sanitizers — `make valgrind` is its memory-safety check) — the default |
 | `make lib` | only `libstdrot.so` |
 | `make debug` | rebuild with `-g` (sanitizers on) for GDB |
 | `make release` | sanitizer-free, rpath'd build for shipping |
--- a/tests/test_docs_consistency.py
+++ b/tests/test_docs_consistency.py
@@ -129,6 +129,46 @@ def test_no_bad_ubuntu_raylib_command_in_code_blocks():
     )
 
 
+# Warning flags the libstdrot.so recipe must carry so stdrot/ is warning-gated
+# in an ordinary `make`, not only the `Release dry-run wasm` job (#330).
+# -Wpedantic is deliberately NOT required here: gcc and clang disagree about
+# STDROT_EXPORT_SIG's static initializer, so the wasm (clang/emcc) job covers
+# -Wpedantic on this code -- see SO_CFLAGS's comment in the Makefile.
+STDROT_REQUIRED_WARNING_FLAGS = ["-Wall", "-Wextra", "-Werror"]
+
+
+def test_libstdrot_recipe_is_warning_gated():
+    """#330: libstdrot.so (stdrot/*.c + lib/input.c) used to compile at gcc's
+    default warning level -- no -Werror, no -Wall -- so a warning in the standard
+    library passed `build`, `test`, `lint` and `static-analysis` and failed only
+    the wasm job. Assert the actual recipe (read via `make -Bn`, so this tracks
+    the Makefile rather than duplicating flags) carries the warning gate, so an
+    introduced warning in stdrot/ fails an ordinary `make`."""
+    result = subprocess.run(
+        ["make", "-Bn", "libstdrot.so"],
+        cwd=REPO_ROOT,
+        capture_output=True,
+        text=True,
+    )
+    assert result.returncode == 0, result.stderr
+    compile_lines = [
+        line
+        for line in result.stdout.splitlines()
+        if "-o libstdrot.so" in line
+    ]
+    assert compile_lines, (
+        "no libstdrot.so compile command found in `make -Bn` output:\n"
+        + result.stdout
+    )
+    recipe = compile_lines[0]
+    missing = [f for f in STDROT_REQUIRED_WARNING_FLAGS if f not in recipe]
+    assert not missing, (
+        f"libstdrot.so recipe is missing {missing} -- stdrot/ would compile "
+        f"without a warning gate again (#330), so a warning there would fail "
+        f"only the wasm job. Recipe:\n{recipe}"
+    )
+
+
 # Every function rayrot exports is spelled `STDROT_EXPORT_SIG("rl_...", ...)`.
 RAYROT_EXPORT_PATTERN = re.compile(r'STDROT_EXPORT_SIG\(\s*"(rl_[a-z0-9_]+)"')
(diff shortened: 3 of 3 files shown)


leo-aa88 (commented):
# APPROVE

**LGTM.** Adversarial pass on this build-gate fix. The change adds `-Werror` to a recipe that many `.so`s share, so the real risk is a *pre-existing* warning breaking `make`; I verified the whole tree stays clean and that the gate actually fires. Built and ran the full `make test` in a worktree.

## Findings

- **CONFIRMED — no build breakage.** The critical risk with adding `-Wall -Wextra -Werror` to `SO_CFLAGS` is that it applies to `libstdrot.so` *and* the badnatives/nativemodules/old-abi-sim/test `.so`s built by the same recipe. All compile clean: `make lib` → "compiled with max rizz", and full `make test` → **530 passed** (which builds every `.so` recipe with the new flags and runs the new docs guard). The recipe carries exactly `-Wall -Wextra -Wuninitialized -Werror`.
- **CONFIRMED — the gate has teeth (mutation).** Injecting an unused variable into `stdrot/baka.c` now makes an ordinary `make lib` fail: `error: unused variable ... [-Werror=unused-variable]`, exit 2 — where before it would have passed everything but the `Release dry-run wasm` job (the #330 hole). Restored → clean rebuild.
- **CONFIRMED — `-Wpedantic` exclusion is correct.** gcc rejects `STDROT_EXPORT_SIG`'s static `StdrotEntry` initializer as "initializer element is not constant" (a GNU extension clang-15/emcc accept) — the same reason #356's print-overflow-check omits `-pedantic`. Leaving it to the wasm (clang/emcc) job is the right call and is documented on `SO_CFLAGS`.
- **CONFIRMED — sanitizers-absent decision is sound and documented.** The `.so` is `dlopen`'d by an already-sanitized host and `make valgrind` runs against these artifacts (ASan shadow vs valgrind), so valgrind stays the memory-safety gate for `stdrot/`. `docs/building.md` is correctly amended (it previously implied the default `make` builds `libstdrot.so` with sanitizers).
- **CONFIRMED — the regression guard is robust.** `test_libstdrot_recipe_is_warning_gated` reads the *actual* recipe via `make -Bn libstdrot.so` and asserts `-Wall`/`-Wextra`/`-Werror`, so the gate can't be silently dropped; it tracks the Makefile rather than duplicating flag strings.

(Local verification used the system gcc; if CI's gcc surfaces an additional warning, that is exactly what this gate now exists to catch.)

## VERDICT

Closes the #330 hole cleanly: `stdrot/` warnings now fail an ordinary `make` instead of only an oddly-named wasm job. Tree is warning-clean under the new flags (no breakage), the gate is mutation-proven, the two intentional exclusions (`-Wpedantic`, sanitizers) are justified and documented, and a recipe-reading test prevents silent regression. As a build-config change there's correctly no `.brainrot` fixture; the Makefile-recipe guard is the right lowest layer. Deserves to merge. (Posting as COMMENT-event with an APPROVE verdict because the reviewer account is also the PR author.)

Issue #280, opened 2026-08-27
[Bug]: int arithmetic (+, -, *, unary -) is signed-overflow UB — interpreter relies on undefined behavior for wrapping


leo-aa88:
### Description

The interpreter implements `rizz` (int) arithmetic with native signed `int`
operations that invoke **undefined behavior on overflow**, which UBSan flags at
runtime. The language's own semantics call for two's-complement wrapping — the
committed fixture `test_cases/integer_overflow.brainrot` literally comments
`🚽 Should output: -2147483648 (overflow)` and its `expected_results.json`
entry is `"-2147483648\n2147483647\n"`. The interpreter produces that output
only because signed overflow *happens* to wrap on x86-64; the operations are
still UB and are not guaranteed to keep working under `-O2` / a different
compiler.

Affected sites (all in `ast.c`, `handle_binary_operation()` /
`handle_unary_expression()`), for `VAR_INT` operands:

- `OP_PLUS`  — `*(int*)result = a + b;`   (ast.c:2433)
- `OP_MINUS` — `*(int*)result = a - b;`   (ast.c:2444)
- `OP_TIMES` — `*(int*)result = a * b;`   (ast.c:2456)
- `OP_NEG`   — `*result = -(*(int*)operand_value);`  (ast.c:3011, on INT_MIN)

This is distinct from #272/#273 (those are SIGFPE *crashes* from `INT_MIN / -1`
and modulo). These four are silent UB, not crashes.

Because UBSan defaults to *recover* (print-and-continue) and the test harness
only diffs stdout, `make test` stays green while the sanitizer prints errors to
stderr — so the corpus never fails on it despite AGENTS.md treating UB as a
blocker.

### Minimal reproduction code

```c
skibidi main {
    rizz max = 2147483647;
    yappin("%d\n", max + 1);   🚽 UBSan: signed integer overflow: 2147483647 + 1
    bussin 0;
}
```

Or just run the already-committed fixture:
`./brainrot test_cases/integer_overflow.brainrot`

### Expected behavior

Wrapping (two's-complement) arithmetic with **no** undefined behavior — e.g.
compute in `unsigned int` (or `int32_t` via `__builtin_add_overflow` etc.) and
store the wrapped result, so the documented `-2147483648` / `2147483647`
outputs are produced by defined operations.

### Actual behavior

```shell
$ ./brainrot test_cases/integer_overflow.brainrot
[...]

Issue #177, opened 2026-08-16
wasm build: sizeof(giga) is 4, not 8 (wasm32 ILP32 vs native LP64)


leo-aa88:
## What happens

`giga` (C `long`) reports a different `sizeof` result on the WASM build (#175) than on native. `thicc` (C `long long`) is unaffected — `long long` is 8 bytes under both data models, and `test_cases/thicc*.brainrot` pass unchanged through the wasm harness.

```
test_cases/giga.brainrot:
    giga rizz num1 = 3;
    rizz num2 = 7;
    yapping("%d", maxxing(num1));   // sizeof(long)
    yapping("%d", maxxing(num2));   // sizeof(int)

native:  8, 4
wasm:    4, 4
```

`test_cases/giga_array.brainrot` shows the same thing through an array: `3 giga elements` sizes to `24` bytes natively, `12` under wasm.

## Root cause (confirmed, not a guess)

`ast.c` implements `maxxing`/`sizeof` for `giga` with the actual C `sizeof(long)` (`ast.c:71`), and for `thicc` with `sizeof(long long)` (`ast.c:69`). Native Linux x86-64 uses the LP64 data model (`long` = 8 bytes, `long long` = 8 bytes). Emscripten's default target is **wasm32**, which uses ILP32 (`long` = 4 bytes, `long long` still 8 bytes) — the same split as 32-bit x86. This is an inherent property of the wasm32 architecture, not a bug in Brainrot's sizeof logic, and it only touches `giga`.

wasm64 exists in the WASM spec and Emscripten has experimental support for it, but it isn't broadly supported in browsers yet, so it's not a realistic target for the playground (#7) right now.

## Impact

Any Brainrot program that assumes `sizeof(giga)` is `8` (matching native) will silently compute `4` in the browser playground — e.g. manual pointer arithmetic or buffer sizing based on that size would be wrong. This is a real, user-visible platform difference, not just a test-fixture quirk.

## What I'm not proposing

Not suggesting a code fix — there isn't a clean one without wasm64 (not viable yet) or redefining `giga`'s width per-platform (would make `giga` mean different things on different builds of the *same* language, which is a much bigger design conversation than this issue should carry).

## Suggested next step

[...]

PR #367, opened 2026-09-07
feat: real 64-bit storage for giga/thicc (long/long long) (#282)

### LINKED ISSUE #282: [Bug]: giga/thicc (long/long long) silently truncate to 32 bits — no 64-bit storage, literal, or marshalling

leo-aa88:
## Description

`giga` (`long`) and `thicc` (`long long`) were documented as ✅ but had **no 64-bit storage anywhere in the pipeline**: the value union held only `int ivalue`, integer literals were parsed with `atoi()` into an `int` (UB on overflow), and native-call marshalling had no 64-bit `StdrotValue` variant. A `giga`/`thicc` value above 2³¹ silently truncated — the issue's repro `giga rizz x = 5000000000; yapping("%lld", x)` printed `705032704`.

This adds a full 64-bit integer path alongside the existing 32-bit one, addressing all three sites the issue enumerates plus arithmetic:

- **Storage** — `long long llvalue` added to the `ASTNode` data, `Variable`, and `ReturnValue` unions (and `Value`). A giga/thicc slot is written and read only through `llvalue`, a plain `rizz` only through `ivalue`, so they never alias for the same slot.
- **Literals** — the lexer parses with `strtoll` and emits a new `LONG_LITERAL` token (`create_long_node`) for any value outside 32-bit `int` range, so it is no longer truncated or UB.
- **Evaluation** — a new `evaluate_expression_long()` (the 64-bit sibling of `evaluate_expression_int`), plus `expression_is_long()` to decide width. Arithmetic uses **unsigned wrapping** for defined overflow (matching #280) — UBSan-clean; `LLONG_MAX + 1` wraps to `LLONG_MIN`.
- **All read/write sites** route giga/thicc through the 64-bit path: declaration init, assignment (scalar and packed), comparisons/conditions, `++`/`--`, function **parameters**, giga/thicc **arrays** and **struct fields**.
- **Native ABI** — a new `STDROT_LONG` `StdrotType` and `long long ll` union member, **appended** so every existing enum value and the `StdrotValue` size are unchanged → **no `STDROT_ABI_VERSION` bump needed** (an old v3 `.so`'s memory can never be misread; see the comment on `STDROT_LONG`). `yapping`/`baka`/`yapto` print it, forcing an `ll` length modifier so `%d`, `%ld` and `%lld` all render a 64-bit value correctly.

### Platform note (native vs wasm)
`giga` is `long`, so packed giga array/struct **elements** follow the platform width via `packed_long_store`/`packed_long_load` — 8 bytes native, 4 on wasm32 (ILP32) — matching their stride so an 8-byte write can't overflow a 4-byte wasm slot. `thicc` (`long long`) is 8 bytes everywhere. This is the same native/wasm width divergence as #177 (which is only about `sizeof` reporting): **values round-trip on native, and `thicc` round-trips on every platform.**

## Related Issue

Fixes #282

## Type of Change

- [ ] Documentation update
- [ ] Performance improvement
- [ ] Refactor

## Checklist

### Tests added
- `test_cases/thicc_64bit.brainrot` — portable (thicc = 8 bytes everywhere): a literal above INT_MAX, print via `%lld` and `%d`, `+`/`-`/`*`, `++`, a thicc parameter, defined two's-complement wrapping at the 64-bit boundary, a comparison in an `edgy` condition, and 8-byte array elements.
- `test_cases/giga_64bit.brainrot` — the issue's exact reproduction (`giga rizz x = 5000000000`), plus 64-bit multiplication.

Both wired into `tests/expected_results.json` (portable outputs — no wasm overrides needed). `make test` (534 passed), `make valgrind` (clean, exit 0), and `make format-check` all pass locally.

### Scope note
Function **return types** of `giga`/`thicc` are not yet accepted by the grammar (a `giga rizz f()` return type is a parse error, not silent truncation); giga/thicc parameters, arrays, struct fields, and all scalar operations are supported. Direct `giga`/`thicc` struct fields use the existing `lit`-alias field grammar (as `struct_field_long_modifier` already does). These grammar extensions are natural follow-ups.


leo-aa88 (commented):
# APPROVE

**LGTM.** Adversarial pass on this 64-bit `giga`/`thicc` feature (751 additions, 13 files). Built the head in a worktree and exercised storage, literals, arithmetic, overflow, mixed-width, printing, the ABI, and non-regression under ASan/UBSan + valgrind + full `make test`.

## Findings

None blocking. One non-blocking QUESTION (wasm, un-testable locally) below.

- **CONFIRMED — the truncation bug is fixed.** The issue's `giga rizz x = 5000000000; yapping("%lld", x)` now prints `5000000000` (was `705032704`); `%d` on the same value also prints `5000000000` because the print path force-rewrites the conversion to `ll` (`stdrot_ll_specifier`), avoiding the varargs UB of a 64-bit value under `%d`. Both shipped fixtures produce their exact expected output.
- **CONFIRMED — full 64-bit path works.** Storage via a dedicated `llvalue`/`ll` union member (scalar giga/thicc never aliases a `rizz`'s `ivalue`); `strtoll` + `LONG_LITERAL` for out-of-int-range literals; `+`/`-`/`*`, `++`, parameters, comparisons, and 8-byte array elements all round-trip 64-bit values. Mixed `thicc rizz * rizz` promotes to 64-bit (`9000000000`). Small values in a giga/thicc slot (`giga rizz x = 5`) work — the width decision is sound.
- **CONFIRMED — overflow is defined.** 64-bit arithmetic uses unsigned wrapping (matching #280): `LLONG_MAX + 1` → `-9223372036854775808`, and the fixture runs UBSan `halt_on_error=1` clean (no `runtime error`).
- **CONFIRMED — the "no ABI bump" claim.** `STDROT_LONG` is appended after `STDROT_NONE` (existing enum values unchanged) and `long long ll` joins a union that already holds `double d` (8 bytes), so `sizeof(StdrotValue)` is unchanged — an old v3 `.so`'s memory can't be misread. `abi-check` (in `make test`) passes.
- **CONFIRMED — no regression.** `rizz` 32-bit path unaffected (`2000000001`); full `make test` → **534 passed**; valgrind on both fixtures → 0 lost, 0 errors.

## QUESTION (non-blocking, couldn't verify — no local emscripten)
`giga` is `long` = 4 bytes on wasm32, and the PR says *packed* giga array/struct elements follow platform width there. A **scalar** `giga rizz`, though, stores through the 64-bit `llvalue`, so it round-trips even on wasm — which is why `giga_64bit` (scalar-only) needs no wasm override. That does leave a platform-dependent inconsistency on wasm: a scalar `giga` holds 64 bits while a *packed* giga array element holds 32 — worth a doc note, and worth confirming the wasm CI job is green on `giga_64bit`. Native behavior (the feature's target) is fully correct.

## Scope (as documented, not a finding)
`giga`/`thicc` function **return types** are a clean parse error (not silent truncation); parameters, arrays, struct fields, and all scalar ops are supported. Reasonable follow-up.

## VERDICT

A large but carefully-built feature that adds a genuine 64-bit integer path end-to-end without disturbing the 32-bit path or the native ABI, with defined overflow and correct printing. Native round-trips verified across every context I attacked; suite green, memory clean. Deserves to merge. (Posting as COMMENT-event with an APPROVE verdict because the reviewer account is also the PR author.)

leo-aa88 (commented):
# APPROVE

**LGTM.** Re-review of head `ed49195` (was `7ac29fd`). The only change since my previous review is the follow-up commit *"merge STDROT_LONG into STDROT_INT's case to satisfy clang-tidy"* — a 3-line change in `stdrot_type_to_vartype()`.

Verified it's semantically identical and safe: `STDROT_INT` now falls through (comment only, no statement between the labels) into `STDROT_LONG`'s `return VAR_INT`, so both types still map to `VAR_INT` exactly as before — this just collapses a clang-tidy branch-clone. The 64-bit feature code is otherwise byte-identical to what I approved. Built the head: compiles clean, and both `giga_64bit` (`5000000000 / 10000000000`) and `thicc_64bit` (full 64-bit round-trip incl. `-9223372036854775808` wrap) produce unchanged output. No conflict markers, `MERGEABLE`.

## VERDICT

Same APPROVE — lint-only follow-up, behavior unchanged. My prior non-blocking QUESTION (confirm the wasm CI job is green on the platform-width `giga` element behavior) still stands as the one thing I couldn't verify locally. Deserves to merge. (COMMENT-event; reviewer is also the PR author.)

